PR #401, opened 2015-10-15
fix btn-group mixin bug

kevyu:
in less, we need to put **&** before **>** if there is a **&** after **>**, otherwise the generated css will be wrong, for example:

less style like this

``` css
span {
  color: #fff;
    > & {
     color: #fff;
     }
}
```

will produce

``` css
span {
  color: #fff;
}
 > span {
  color: #fff;
}
```

which is wrong, what expected is 

``` css
span {
  color: #fff;
}
 span > span {
  color: #fff;
}
```

I wonder if this is a bug in less.

### DIFF
--- a/style/mixins/button.less
+++ b/style/mixins/button.less
@@ -253,15 +253,15 @@
     padding-left: 8px;
   }
 
-  > & {
+ & > & {
     float: left;
   }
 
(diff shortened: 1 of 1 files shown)


simaQ:
Yeah, you were! THX! :+1:

kevyu:
@simaQ  :smile:

PR #1239, opened 2016-03-21
More ES2015 classes.

waywardmonkeys:
This covers Carousel, InputNumber, Select, Slider, Switch, TreeSelect
and Tree.

This is a continuation of the work from #1223 and #1237, working on the task laid out in #1179.


benjycui:
Gelivable :+1:

PR #1272, opened 2016-03-25
Use babel as the eslint parser.

waywardmonkeys:
This allows eslint to validate all of the code that babel is
configured to accept. Without this, eslint won't validate code
that is part of ES7.

We use the beta for 6.0 to avoid https://github.com/babel/babel-eslint/issues/249

### DIFF
--- a/.eslintrc.json
+++ b/.eslintrc.json
@@ -1,4 +1,5 @@
 {
+  "parser": "babel-eslint",
   "extends": ["eslint-config-airbnb"],
   "env": {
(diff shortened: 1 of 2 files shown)


waywardmonkeys:
@yiminghe, @afc163, @benjycui:

With this commit, doing this sort of code is accepted by `eslint`:

```javascript
handleClick = (...args) => { ... }
```

yiminghe:
:+1:

PR #1289, opened 2016-03-28
Convert several React.createClass usages to classes.

waywardmonkeys:


benjycui:
After you have convert all the `React.createClass` to ES6 class, you can turn on [react/prefer-es6-class](https://github.com/ant-design/ant-design/blob/develop-1.0.0/.eslintrc.json#L31)

yiminghe:
let's use static

```js
class x {
 static propTypes={};
}
```

waywardmonkeys:
@yiminghe I was going to suggest that as well now that we have ES7 features available ... but figured that I'd do it in a separate pass. I will add it to this patch though today and submit a patch for existing usages separately.

waywardmonkeys:
@yiminghe Done. (Getting closer and closer to just being TypeScript...)

yiminghe:
yes, I am planing to use typescript

PR #1292, opened 2016-03-29
Use static class members for propTypes and friends.

waywardmonkeys:


waywardmonkeys:
@yiminghe Here you go.

yiminghe:
thanks

benjycui:
Code convention updated https://github.com/ant-design/ant-design/wiki/Code-convention-for-antd#define-static-property-in-class-definition

PR #1295, opened 2016-03-29
Convert more createClass over to extending Component.

waywardmonkeys:


waywardmonkeys:
This leaves ``component/date-picker`` (which uses mixins) and ``component/table`` (which there's a pending PR that changes it) to be converted.

benjycui:
Relative: https://github.com/ant-design/ant-design/pull/1298

PR #1301, opened 2016-03-30
style: enable react/jsx-no-bind

benjycui:


waywardmonkeys:
Thanks for doing this! I'll wait for this to land before I update the ``components/date-picker`` to use es6 classes.

PR #1411, opened 2016-04-13
ru locale

plandem:


plandem:
I haven't translated Transfer, because it's not so simple. In russian we have: http://formatjs.io/guides/message-syntax/#plural-format

not only item/items, but: for 1 - one word, for 2-4 - another, for 5-20 - another, 21 - another, 22-24 - another, 25-30 - another and after that same pattern...

PR #1420, opened 2016-04-14
Replace property `spining` with `spinning`

wizawu:
It might be easy for users to just misuse `spinning`, which is the correct spelling, without reading the [docs](http://ant.design/components/spin/#api) very carefully.

The commit(s) below will also support `spining` while providing an option to use `spinning`. But I suggest only `spinning` should be mentioned in the usage docs.

### DIFF
--- a/components/spin/demo/nested.md
+++ b/components/spin/demo/nested.md
@@ -27,7 +27,7 @@ const Card = React.createClass({
     );
     return (
       <div>
-        <Spin spining={this.state.loading}>{container}</Spin>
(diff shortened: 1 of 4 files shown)


benjycui:
:+1:

afc163:
Thank u, and it is a breaking change.

We should add warning console for it.

yt46767:
spining
For details, see "node_modules/antd/lib/spin/index.js".
spining is a wrong spelling in the usage docs. we must write  'spinning'.

PR #1641, opened 2016-05-11
fix: should pass locale to TimePicker

benjycui:
Relative: https://github.com/ant-design/ant-design/issues/1635


benjycui:
GitHub 更新后，这个扫描很慢。。。

PR #2043, opened 2016-06-14
Missing size props in interface TabProps from index.d.ts

andycall:


index.d.ts
```diff
@@ -1549,6 +1549,8 @@ interface TabsProps extends React.Props<Tabs> {
   tabPosition?:TabsPosition,
   /** 新增和删除页签的回调，在 `type="editable-card"` 时有效*/
   onEdit?:(targetKey:string, action:any) => void,
+  /** 大小，提供 default 和 small 两种大小    */
+  size? string,
```
bang88: size?: string,

PR #2221, opened 2016-06-30
 Rewrite Checkbox in typescript

RaoHai:
- add return type of function splitObject
- rewrite `import React from 'react'` to `import * as React from 'react'`
- rewrite Checkbox in typescript

### DIFF
--- a/components/checkbox/Group.tsx
+++ b/components/checkbox/Group.tsx
@@ -1,8 +1,34 @@
-import React from 'react';
+import * as React from 'react';
 import Checkbox from './index';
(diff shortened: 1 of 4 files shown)


components/checkbox/Group.tsx
```diff
@@ -1,8 +1,34 @@
...
 import assign from 'object-assign';
-export default class CheckboxGroup extends React.Component {
+
+export interface CheckboxOptionType {
+  label:string,
+  value:string,
+  disabled?:boolean
+}
+
+interface CheckboxGroupProps {
```
GrayChoi: Ask a question, when should we export a props interface?

PR #2574, opened 2016-08-01
Review of form / demo / advanced-search-form.md

dxps:
Small review.

### DIFF
--- a/components/form/demo/advanced-search-form.md
+++ b/components/form/demo/advanced-search-form.md
@@ -13,9 +13,9 @@ title:
 
 ## en-US
 
-Three columns layout is often used to advanced searching of data table.
+Three columns layout is often used for advanced searching of data table.
 
-Because the width of label is not definite, you need to adjust the custom stlye case by case.
+Because the width of label is not fixed, you may need to adjust it by customizing its style.
 
 ````jsx
 import { Form, Input, Row, Col, Button, DatePicker } from 'antd';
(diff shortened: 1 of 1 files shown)


components/form/demo/advanced-search-form.md
```diff
@@ -114,7 +114,7 @@ ReactDOM.render(
   border-radius: 6px;
 }
 
-/* becasue the label width is not definite, you need to adjust it to center by yourself */
+/* because the label width is not fixed, you may need to adjust it */
```
GrayChoi: In this case, I use this style to center the form item, 
but be because the label width is not fixed,
anyone want to center the element, need to modify the left position,
so I think you need to mention that?

visvadw: Hi Gray,
I just discovered this UI library and felt in love with it. So, I was in a hurry to start contributing.
You're right, I wasn't sure about the meaning of that statement.
Let me play a little bit with this example and get back to this case.

benjycui: updated https://github.com/ant-design/ant-design/commit/4cf04c579992d4f65a2e8c2a4b02c69a476da36d

visvadw: @GrayChoi Hi Gray, I played a little bit with the demo, so now I understand what you mean, however, adjusting the left edge of the form row using `left: -6` does not really center the form no matter the width of the viewport.

PR #2740, opened 2016-08-17
docs: tranlsation of Checkbox

benjycui:
ref: #1471

### DIFF
--- a/components/checkbox/index.en-US.md
+++ b/components/checkbox/index.en-US.md
@@ -1,0 +1,31 @@
+---
+category: Components
+type: Form Controls
+title: Checkbox
+---
+
+Checkbox.
+
+## When To Use
+
+- Used for selecting multiple values from several options.
+- If you use only one checkbox, it is the same as using Switch to toggle between two states. The difference is that Switch will trigger the state change directly, but Checkbox just marks the state as changed and this needs to be submitted.
+
+## API
+
+### Checkbox
+
+| Property      | Description             | Type      | Default  |
+|----------|------------------|----------|--------|
+| checked | Specifies whether the checkbox is selected. | Boolean | false |
+| defaultChecked | Specifies the initial state: whether or not the checkbox is selected. | Boolean | false |
+| onChange | The callback function that is triggered when the state changes. | Function(e:Event) | - |
+
+### Checkbox Group
+
+| Property      | Description             | Type      | Default  |
+|----------|------------------|----------|--------|
+| defaultValue | Default selected value | Array | [] |
+| value | Used for setting the currently selected value. | Array | [] |
+| options  | Specifies options | Array | [] |
+| onChange | The callback function that is triggered when the state changes. | Function(checkedValue) | - |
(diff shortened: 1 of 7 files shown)


benjycui:
Ping~ @visvadw

components/checkbox/index.en-US.md
```diff
@@ -0,0 +1,31 @@
...
+category: Components
+type: Form Controls
+title: Checkbox
+---
+
+Checkbox.
+
+## When To Use
+
+- Used to select serveral values in multiple options.
```
visvadw: - Used for selecting several values from multiple options.

components/checkbox/index.en-US.md
```diff
@@ -0,0 +1,31 @@
...
+- If you just use only one checkbox, it is the same as `Switch` to switch between two states. The difference is that `Switch` will trigger state change directly, but `Checkbox` just marks state changed and need to be submitted.
```
visvadw: - If you use only one checkbox, it is the same as using `Switch` to toggle between two states. The difference is that `Switch` will trigger state change directly, but `Checkbox` just marks the state as changed and this needs to be submitted.

components/checkbox/index.en-US.md
```diff
@@ -0,0 +1,31 @@
...
+
+## API
+
+### Checkbox
+
+| Property      | Description             | Type      | Default  |
+|----------|------------------|----------|--------|
+| checked | Specifies whether the current is selected | Boolean  | false |
```
visvadw: | checked | Specifies whether the checkbox is selected. | Boolean  | false |

components/checkbox/index.en-US.md
```diff
@@ -0,0 +1,31 @@
...
+| defaultChecked | Initial whether or not selected | Boolean | false |
```
visvadw: | defaultChecked | Specifies the initial state: whether or not the checkbox is selected. | Boolean | false |

components/checkbox/index.en-US.md
```diff
@@ -0,0 +1,31 @@
...
+| onChange | The callback function when the state change | Function(e:Event) | - |
```
visvadw: | onChange | The callback function that is triggered when the state changes. | Function(e:Event) | - |

components/checkbox/index.en-US.md
```diff
@@ -0,0 +1,31 @@
...
+
+### Checkbox Group
+
+| Property      | Description             | Type      | Default  |
+|----------|------------------|----------|--------|
+| defaultValue | Default selected value | Array | [] |
+| value | Used to set the current selected value| Array | [] |
```
visvadw: | value | Used for setting the currently selected value. | Array | [] |

visvadw:
@benjycui Done. Nice job! 👍

visvadw:
@benjycui I would rename `checkbox/demo/controller.md` as `checkbox/demo/controlled.md`.

benjycui:
Yep, `controlled` is better than `controller`, but we must keep the url stable...

PR #3006, opened 2016-09-11
radio/demo/disable.md - English translation

dxps:
English translation of labels


benjycui:
Actually, you can create a PR which includes several commits...

PR #3115, opened 2016-09-21
Heading translation - English translation

kvetoslavnovak:


benjycui:
We are going to provide both of Chinese and English change log in the future. https://github.com/ant-design/ant-design/pull/3087

kvetoslavnovak:
I see. 
But for know the English heading (as well as for generated navigation text) would be helpfull and does not conflict this future intention. 
In the same case of the "Upgrade Notes" page this was accepted - http://beta.ant.design/docs/react/upgrade-notes

benjycui:
The heading will be translated in that PR too. Anyway, thank you for your PR 😄

PR #3443, opened 2016-10-15
Site Title Eng. translation

kvetoslavnovak:


benjycui:
Actually, this is not important... For it will be override by `react-document-title`

kvetoslavnovak:
I see.

PR #3527, opened 2016-10-21
hide scrollbar in ie

zation:
hide scrollbar in ie for issue https://github.com/ant-design/ant-design/issues/3511


zation:
这个修改会导致 mac 下浏览器显示出错，先关闭了

PR #4081, opened 2016-11-30
Fix imports for modules without default export

sologub:
The latest Typescript compiler v2.0.10 gives a "Module has no default export." error for imports of the form: `import React from 'react';`
The correct form is: `import * as React from 'react';`

This PR fixes the imports for modules React, ReactDOM and moment in all TSX files.

### DIFF
--- a/components/affix/index.tsx
+++ b/components/affix/index.tsx
@@ -1,5 +1,5 @@
-import React from 'react';
-import ReactDOM from 'react-dom';
+import * as React from 'react';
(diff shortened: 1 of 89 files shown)


sologub:
It seems I broke something, closing this PR for now...

benjycui:
See: https://github.com/ant-design/ant-design/blob/master/tsconfig.json#L5

sologub:
Yes, thanks.

PR #4267, opened 2016-12-15
Add showZero prop for Badge ant-design/ant-design#4251

jmatsushita:


benjycui:
Could you add test case for this feature? Thanks.

You can write test case with Jest and enzyme, and put them in `components/badge/__tests__/index.test.jsx`

benjycui:
It seems that snapshot can make sure this feature works.

benjycui:
You should PR to branch `feature-2.6`, not `master`. And you should run `npm run jest -- -u` to update snapshot before commit.

Close now, and please create a new PR, thanks. 😄

PR #4285, opened 2016-12-18
Add info about how to turn off pagination [ci skip]

SergioCrisostomo:
Adding info about how to turn off pagination in tables.

### DIFF
--- a/components/table/demo/paging.md
+++ b/components/table/demo/paging.md
@@ -11,7 +11,7 @@ title:
 
 ## en-US
 
-when use pagination in table.
+When using pagination in table, and you can use pagination={false} to turn off pagination.
 
 ````jsx
 import { Table } from 'antd';


components/table/demo/paging.md
```diff
@@ -53,3 +53,4 @@ ReactDOM.render(
   <Table columns={columns} dataSource={data} pagination={pagination} />
 , mountNode);
 ````
+You can use `pagination={false}` to turn off pagination.
```
benjycui: Actually, you can write docs after code ..

components/table/demo/paging.md
```diff
@@ -11,7 +11,7 @@ title:
 
 ## en-US
 
-when use pagination in table.
+when using pagination in table:
```
benjycui: When using pagination in table, and you can use `pagination={false}` to turn off pagination.

SergioCrisostomo: @benjycui updated!

PR #4565, opened 2017-01-11
Change AutoCompleteProps to extend SelectProps

normalser:
According to documentation: https://ant.design/components/auto-complete/

"Since AutoComplete is based on Select, so besides the following API, AutoComplete has the same API as Select"

This PR changes AutoCompleteProps interface to extend SelectProps

Otherwise examples from that URL will not work in TypeScript

eg.

```tsx
<AutoComplete
        dataSource={dataSource}
        style={{ width: 200 }} // -> TS ERROR
        onSelect={onSelect} // -> TS ERROR
        onChange={this.handleChange}
        placeholder="input here" // -> TS ERROR
      />
```

### DIFF
--- a/components/auto-complete/index.tsx
+++ b/components/auto-complete/index.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import Select, { SelectProps, OptionProps, OptGroupProps } from '../select';
+import Select, { AbstractSelectProps, OptionProps, OptGroupProps } from '../select';
 import { Option, OptGroup } from 'rc-select';
 import classNames from 'classnames';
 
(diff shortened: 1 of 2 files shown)


benjycui:
Yep, we can simplify `interface AutoCompleteProps`.

But you had better to do it like [this](https://github.com/ant-design/ant-design/blob/master/components/popover/index.tsx#L7), for AutoComplete is not similar to Select 100%.

wallverb:
Thanks @benjycui 
I extracted all common properites (with exactly the same type) into `AbstractSelectProps` and added `style, placeholder, onSelect` based on the documentation from https://ant.design/components/auto-complete/

Not sure if there are any other fields in SelectProps left that are applicable to AutoComplete as well

benjycui:
NVM, it's a good start point, we can improve it anytime.

PR #4856, opened 2017-02-14
add class ant-form-item-control-wrapper for formItem wrapper

jzlxiaohei:
Close: https://github.com/ant-design/ant-design/issues/4841


benjycui:
👍

PR #5031, opened 2017-02-23
Fix RangePicker TimePicker component locking background to white

megawac:
Because the background here is locked to white and the foreground is set to `@text-color` the TimePicker combo ends up not working for dark themes.

Thank you!


benjycui:
Years ago, I PR to Ramda and you review my PR.

Now, you PR to antd and I review your PR.

Small world 😄

megawac:
Ha! look forward to working to you more in the future

PR #5133, opened 2017-03-01
Added missing getFieldsError in WrappedFormUtils

krosenberg:
The [documentation](https://ant.design/components/form/#Form.create(options)) states that the Form API includes a `getFieldsError ` function, but there it is not included in the [WrappedFormUtils](https://github.com/ant-design/ant-design/blob/master/components/form/Form.tsx#L50) type. This PR adds `getFieldsError ` to the type alias based upon rc-form's function defined [here](https://github.com/react-component/form/blob/master/src/createBaseForm.js), which takes an optional array of field names (strings) as a parameter.

### DIFF
--- a/components/form/Form.tsx
+++ b/components/form/Form.tsx
@@ -48,6 +48,7 @@ export type WrappedFormUtils = {
                           callback?: (erros: any, values: any) => void): void;
   /** 获取某个输入控件的 Error */
   getFieldError(name: string): Object[];
+  getFieldsError(names?: Array<string>): Object[];
   /** 判断一个输入控件是否在校验状态*/
   isFieldValidating(name: string): boolean;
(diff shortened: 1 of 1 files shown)


components/form/Form.tsx
```diff
@@ -48,6 +48,7 @@ export type WrappedFormUtils = {
                           callback?: (erros: any, values: any) => void): void;
   /** 获取某个输入控件的 Error */
   getFieldError(name: string): Object[];
+  getFieldsError(names?: Array<string>): Object[];
```
benjycui: The returned type should be an object, see: https://github.com/react-component/form/blob/master/src/createBaseForm.js#L253-L260

benjycui:
Oops... @infeng

infeng:
:scream: sorry

benjycui:
NVM, just update it @infeng , the type is wrong.

PR #5134, opened 2017-03-02
chore: improve ts definition of Form

### LINKED ISSUE #5131: missing typescript prop in interface for form create options
#### Environment(required)

- antd version: 2.7.4
- OS and its version: Mac OS 10.11.6
- Browser and its version: any

I tried passing onValuesChange as a prop to the Form.create() method in a TypeScript project.

#### What do you expected?

I expected the prop to be valid

#### What happen?

Typescript said that prop was not valid for Form.create

onValuesChange is stated as a valid prop in documentation: https://ant.design/components/form/
onValuesChange is missing as an available prop in the FormCreateOptions interface: https://github.com/ant-design/ant-design/blob/master/components/form/Form.tsx#L12

infeng:
close #5131 and fix [error](https://github.com/ant-design/ant-design/blob/master/components/form/Form.tsx#L51) of Form

### DIFF
--- a/components/form/Form.tsx
+++ b/components/form/Form.tsx
@@ -11,6 +11,7 @@ import { FIELD_META_PROP } from './constants';
 
 export interface FormCreateOption {
   onFieldsChange?: (props: any, fields: Array<any>) => void;
+  onValuesChange?: (props: any, values: any) => void;
   mapPropsToFields?: (props: any) => void;
   withRef?: boolean;
 }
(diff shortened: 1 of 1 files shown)


components/form/Form.tsx
```diff
@@ -13,6 +13,7 @@ export interface FormCreateOption {
   onFieldsChange?: (props: any, fields: Array<any>) => void;
```
benjycui: I prefer to put `onValuesChange` here.. We need to group relative API together.

infeng: you are right

PR #5145, opened 2017-03-02
typescript upgrade to 2.2

ddcat1115:
- ~~[x] change Object to new type object~~ reverted for it's breaking change
- [ ] mixins support

### DIFF
--- a/components/back-top/index.tsx
+++ b/components/back-top/index.tsx
@@ -82,7 +82,7 @@ export default class BackTop extends React.Component<BackTopProps, any> {
     const { visibilityHeight, target = getDefaultTarget } = this.props;
     const scrollTop = getScroll(target(), true);
     this.setState({
-      visible: scrollTop > visibilityHeight,
+      visible: scrollTop > (visibilityHeight as number),
     });
   }
 
--- a/components/grid/row.tsx
+++ b/components/grid/row.tsx
@@ -17,6 +17,7 @@ export default class Row extends React.Component<RowProps, any> {
   static defaultProps = {
     gutter: 0,
   };
+
   static propTypes = {
     type: React.PropTypes.string,
     align: React.PropTypes.string,
@@ -27,16 +28,17 @@ export default class Row extends React.Component<RowProps, any> {
     prefixCls: React.PropTypes.string,
   };
   render() {
-    const { type, justify, align, className, gutter, style, children, prefixCls = 'ant-row', ...others } = this.props;
+    const { type, justify, align, className, gutter, style, children,
+      prefixCls = 'ant-row', ...others } = this.props;
     const classes = classNames({
       [prefixCls]: !type,
       [`${prefixCls}-${type}`]: type,
       [`${prefixCls}-${type}-${justify}`]: type && justify,
       [`${prefixCls}-${type}-${align}`]: type && align,
     }, className);
-    const rowStyle = gutter > 0 ? assign({}, {
-      marginLeft: gutter / -2,
-      marginRight: gutter / -2,
+    const rowStyle = (gutter as number) > 0 ? assign({}, {
+      marginLeft: (gutter as number) / -2,
+      marginRight: (gutter as number) / -2,
     }, style) : style;
     const cols = Children.map(children, (col: React.ReactElement<any>) => {
       if (!col) {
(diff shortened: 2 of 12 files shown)


yesmeck:
Is this a breaking change? If we introduce the new type `object`, people using antd will have to upgrade to TypeScript 2.2 too.

ddcat1115:
ready to merge, leave `mixins support` to later optimization

paranoidjk:
just my point of view，`mixin` may be not a very good pattern,  especially the older `React LifeCycle mixin` use case

components/back-top/index.tsx
```diff
@@ -40,10 +40,6 @@ export interface BackTopProps {
 }
 
 export default class BackTop extends React.Component<BackTopProps, any> {
-  static defaultProps = {
-    visibilityHeight: 400,
-  };
```
yesmeck: Why not use defaultProps?

paranoidjk: does this syntax will auto bind `this.props.visibilityHeight` ? I understand it‘s just something like `default parameter`.
so what if call `this.props.visibilityHeight` in constructor or lifeCycle and other function?  I guess it will be `undefined`, then this will cause problem

paranoidjk: https://github.com/ddcat1115/ant-design/blob/master/components/back-top/index.tsx#L75

paranoidjk: suggest still to use `defaultProps`

yesmeck:
这个先不合到 2.8 里了，对用户没影响的。

components/back-top/index.tsx
ddcat1115: yes, I got it wrong...

It seems even if the props(typed in props interface: `a?: number`) has been defined in `defaultProps ` , for example:

```tsx
static defaultProps = {
	a: 1,
}
```  

I still need to specify the type when I use it?

```tsx
const { a } = this.props; 
const aa = a - 1;  // ERROR: prop 'a' possibly undefined
const add = (a as number) - 1;  //  ✔️
```

paranoidjk: I came across this problem too... 

https://github.com/ant-design/ant-design-mobile/commit/e5d1a708f16076a879fee5f49dbe63e9f6d5c90f#commitcomment-20970016

seems like Typescript type-inference will  not handle React defaultProps, and it's actually
reasonable， since  `defaultProps` do not belongs to any EcmaScript Standard

ddcat1115: Hum... I think I'll use `as number` too 😂   By now I can't find any better solution...

yesmeck: 以 `visibilityHeight` 为例， 因为 `BackTopProps` 里写了 `visibilityHeight?: number;`，`visibilityHeight` 就可能是 `undefined`，但是因为我们在 `defaultProps` 里写了 `visibilityHeight`，所以其实在 `BackTopProps` 可以把 `visibilityHeight` 的 `?` 去掉。

yesmeck: @ @ 去掉也不行。。。

paranoidjk: 应该是不能去掉的，BackTopProps 是typescript的interface，defaultProps只是React的一个特性，而typescript类型推导的时候不考虑react自己的defaultProps规则，这就是我上面说的

components/grid/row.tsx
```diff
@@ -27,26 +28,27 @@ export default class Row extends React.Component<RowProps, any> {
...
     const classes = classNames({
       [prefixCls]: !type,
       [`${prefixCls}-${type}`]: type,
       [`${prefixCls}-${type}-${justify}`]: type && justify,
       [`${prefixCls}-${type}-${align}`]: type && align,
     }, className);
-    const rowStyle = gutter > 0 ? assign({}, {
-      marginLeft: gutter / -2,
-      marginRight: gutter / -2,
+    const rowStyle = (gutter as number) > 0 ? assign({}, {
```
paranoidjk: https://github.com/ant-design/ant-design-mobile/blob/master/components/textarea-item/index.tsx#L73

这样重复代码量少点，也避免分散不好维护。

benjycui: 看到不合理的地方，直接改，毕竟你是 maintainer

components/back-top/index.tsx
paranoidjk: https://github.com/ant-design/ant-design-mobile/issues/1873

components/grid/row.tsx
paranoidjk: 提前赋值来 cast 类型会造成编译后残留 https://github.com/ant-design/ant-design-mobile/issues/1873

Issue #5203, opened 2017-03-07
onCancel not being called for modals


elios264:
#### Environment(required)

- antd version:  2.7.4
- OS and its version: Windows 10
- Browser and its version: chrome  56.0.2924.87

#### What did you do? Please provide steps to re-produce your problem.

        Modal.confirm({
          title: 'Sample title',
          content: `Sample content`,
          onCancel: () => console.log('cancelling') });

#### What do you expected?

a console log when pressing ESC key

#### What happens?

The modal closes but the log is not printed

#### Re-producible online demo

http://codepen.io/elios264/pen/RpoVeE

NE-SmallTown:
I think it's not a bug,maybe it's brower's behavior(i.e.,remove all nodes which are masks ),you can use `alert()` or `confirm()` on the browser and then press "Esc" key,you will see that they will disappear.

Maybe these help?

[Five unexpected uses for the Esc key](http://www.macworld.com/article/1143530/esckey.html)

[bootstrap modal handle esc](https://github.com/twbs/bootstrap/blob/v4-dev/js/src/modal.js)

[bootstrap modal doc](https://getbootstrap.com/javascript/#js-programmatic-api)

[relevant issue1](https://github.com/twbs/bootstrap/issues/13929)

[relevant issue2](https://github.com/twbs/bootstrap/issues/13816)

elios264:
Hi @NE-SmallTown, 

It is not browser behaviour, I checked the underline source code but I couldn't figure out because the library uses rc-dialog underneath, but the rc-dialog does calls onCancel when pressing ESC,

I read "cancel button and x button"  too,  I just think they missed that case.

benjycui:
We need to trigger `props.onCancel` and pass event as argument [here](https://github.com/ant-design/ant-design/blob/6bca3ba87fcde17e845f4cf04f22c6006212fd20/components/modal/confirm.tsx#L33)

Could you try to PR to fix this? @elios264 @NE-SmallTown

NE-SmallTown:
@elios264 Sorry for that I don't check [rc-dialog](https://github.com/react-component/dialog/blob/master/src/Dialog.tsx#L137),it does handle the "ESC" key.

```
 if (props.keyboard && e.keyCode === KeyCode.ESC) {
      this.close(e);
 }
```

It does is a bug like @benjycui says,when you press "ESC" key,the execute order is:

```
[...]

PR #5208, opened 2017-03-08
fix press `ESC` key don't call onCalcel callback function

NE-SmallTown:
fix issue [5203](https://github.com/ant-design/ant-design/issues/5203)


NE-SmallTown:
close,because [#5509](https://github.com/ant-design/ant-design/pull/5209) did better,sorry for that I don't think about the [typescript](https://github.com/ant-design/ant-design/pull/5209/files#diff-05de9ca94fd8b5565d4cc1f682b41616R60) because I don't know it too much

PR #5343, opened 2017-03-16
Document argument order swap in sorter function.

JesperWe:

### DIFF
--- a/components/table/index.en-US.md
+++ b/components/table/index.en-US.md
@@ -93,7 +93,7 @@ One of Property `columns` for descriping column, Column has the same API.
 | filterDropdownVisible | whether filterDropdown is visible | boolean | - |
 | onFilterDropdownVisibleChange | called when filterDropdownVisible is changed | function(visible) {} | - |
 | filteredValue | controlled filtered value | string[] | - |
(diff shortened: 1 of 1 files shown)


yesmeck:
I'v add a link to MDN. :)

JesperWe:
I saw that. I just wanted it to mention the argument swapping because I did implement that myself at first, and then it was always sorting in the same direction because you swapped the args once and I swapped them back again :-D

yesmeck:
@JesperWe I reverted your change, because you don't need care about the order of the arguments.

JesperWe:
Well, my point was that "you don't need to care" is non-intuitive because normally you do...

yesmeck:
Actually, the order of compare function's arguments is not sequential.  Different browsers have different sorting algorithm.

http://stackoverflow.com/questions/234683/javascript-array-sort-implementation

JesperWe:
That's not the point I was trying to document. The sorter takes two arguments (a,b) and compare them. Normally for ascending sort you return 1 if `a > b`, and for descending you return 1 if `a < b`.

Now Ant Design Table will give you the first row as `a` and the following row as `b` for ascending sort, and the opposite for descending. This way the compare function does **not** need to know if it is doing ascending or descending, which is very smart and good. 

But it is not the way Array.sort() usually works. Usually you need to know which way you are sorting **in the compare function**, and change your return value accordingly.

JesperWe:
...or maybe you are not swapping the arguments, but just changing the sign of the return value, which has exactly the same effect :-) !

yesmeck:
> you are not swapping the arguments, but just changing the sign of the return value

Yes, your are right now.

Issue #5334, opened 2017-03-16
Dragger component does not work with formDecorator


mimol91:
#### Environment(required)

- antd version: 2.8.1
- OS and its version Linux mint:
- Browser and its version Chrome :56

#### What did you do? Please provide steps to re-produce your problem.
Trying to use formDecorator with `Dragger` component.

P.S.
`Upload` works fine

#### What happen?
Error: Cannot read property `onChange` of undefined

#### Re-producible online demo
http://codepen.io/anon/pen/XMeyvb?editors=1111

## Hacky solution
Change `<Dragger>` to `<Upload  type={"drag"}>`

benjycui:
`getFieldDecorator` cannot work with functional stateless component.

PR #5358, opened 2017-03-17
Fix upload

benjycui:
Close: #5334


components/upload/Dragger.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
```
benjycui: Git diff is crazy.

The most important thing is just make Dragger as a ES6 style component.

PR #5413, opened 2017-03-21
[WIP] refactor: use injectLocale in Upload

benjycui:
ping @bang88

### DIFF
--- a/components/locale-provider/injectLocale.tsx
+++ b/components/locale-provider/injectLocale.tsx
@@ -8,8 +8,12 @@ export interface ComponentContext {
   antLocale?: { [key: string]: any };
 }
 
+export interface ComponentClass<P> {
+  [key: string]: any;
+}
+
 export default (componentName: string, defaultLocale) => (
-  function<P>(Component: typeof React.Component): React.ComponentClass<P> {
+  function<P>(Component: typeof React.Component): ComponentClass<P> {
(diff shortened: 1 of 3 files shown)


components/upload/index.tsx
```diff
@@ -2,4 +2,9 @@ import Upload from './Upload';
 import Dragger from './Dragger';
 
 Upload.Dragger = Dragger;
+
+function Test() {
+  return <Upload.Dragger />;
```
benjycui: This doesn't work as I expected, `Upload.Dragger` is any, so that we cannot get any tips in VSCode.

bang88: Yep, you are right , but we don't want hard code every static member.

components/locale-provider/injectLocale.tsx
```diff
@@ -8,8 +8,12 @@ export interface ComponentContext {
   antLocale?: { [key: string]: any };
 }
 
+export interface ComponentClass<P> {
+  [key: string]: any;
```
bang88: If we don't extends from `React.ComponentClass<P>` will lose the default static members.

bang88:
这个重构要不就先暂缓一下，等有好一点的解决方案再做，现在实现起来有点不舒服。

benjycui:
目前的确没有好的办法，push 上来只是为了让你可以直观的看到问题而已。

PR #5420, opened 2017-03-21
improve Form.create ts definition of Form

infeng:
when set `strictNullChecks: true` in `tsconfig.json`, `this.props.form` ts definition is `WrappedFormUtils | undefined`, it is unreasonable, so I make it necessary.
image

### DIFF
--- a/components/form/Form.tsx
+++ b/components/form/Form.tsx
@@ -75,14 +75,11 @@ export type WrappedFormUtils = {
 }
 
 export interface FormComponentProps {
-  form?: WrappedFormUtils;
+  form: WrappedFormUtils;
 }
 
-export class FormComponent extends React.Component<FormComponentProps, {}> {
-}
-
-export interface ComponentDecorator {
-  <T extends (typeof FormComponent)>(component: T): T;
+export interface ComponentDecorator<TOwnProps> {
+  (component: React.ComponentClass<FormComponentProps & TOwnProps>): React.ComponentClass<TOwnProps>;
 }
 
 export default class Form extends React.Component<FormProps, any> {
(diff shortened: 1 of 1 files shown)


components/form/Form.tsx
```diff
@@ -75,14 +75,11 @@ export type WrappedFormUtils = {
 }
 
 export interface FormComponentProps {
-  form?: WrappedFormUtils;
-}
-
-export class FormComponent extends React.Component<FormComponentProps, {}> {
+  form: WrappedFormUtils;
```
bang88: This maybe a breaking change.

components/form/Form.tsx
```diff
@@ -109,7 +106,7 @@ export default class Form extends React.Component<FormProps, any> {
 
   static Item = FormItem;
 
-  static create = (options?: FormCreateOption): ComponentDecorator => {
+  static create = <TOwnProps extends any>(options?: FormCreateOption): ComponentDecorator<TOwnProps> => {
```
bang88: I don't understand why we need extends from any.

BANG88 (commented):
By the way, from your snapshot,  `form` is typeof `WrappedFormUtils`, 
```ts
this.props.form as WrappedFormUtils

```
type inference is unnecessary?

infeng:
@bang88 The snapshot is usage now, `form ` is typeof `WrappedFormUtils  | undefined`, so I must set type when I use `this.props.form`,  it is not easy to use. `form` become unnecessary in #2885.

components/form/Form.tsx
infeng: If not add `extends any`, it will cause error, I dont`t know why :sweat_smile:.

infeng:
this is the new usage.
image

components/form/Form.tsx
benjycui: I think this is OK.

bang88: 😢   its `FormComponentProps ` not `FormProps`, my bad.

components/form/Form.tsx
bang88: Try `function<TOwnProps>()` instead of arrow function.

BANG88 (approved):

PR #5453, opened 2017-03-23
Refactor RadioGroup's cloneElement to context

### LINKED ISSUE #5443: Refactor cloneElement to context in RadioGroup
Currently, we use [cloneElement](https://github.com/ant-design/ant-design/blob/master/components/radio/group.tsx#L89) to pass props to `Radio` in `RadioGroup`. This approach leads `Radio` cannot be wrapped by other components or created from HOC.

For example, following code won't work:

```javascript
<RadioGroup onChange={this.onChange} value={this.state.value}>
  <div>
    <img src="A.png" />
    <Radio value={1}>A</Radio>
  </div>  
  <div>
    <img src="B.png" />
    <Radio value={2}>B</Radio>
  </div>  
</RadioGroup>
```

By replacing `cloneElement` with `context`, `Radio` can be free to compose.

Related issue: #5165

yesmeck:
Close #5443

### DIFF
--- a/components/radio/group.tsx
+++ b/components/radio/group.tsx
@@ -1,7 +1,6 @@
-import React from 'react';
+import React, { PropTypes } from 'react';
 import classNames from 'classnames';
-import PureRenderMixin from 'rc-util/lib/PureRenderMixin';
-import assign from 'object-assign';
+import shallowEqual from 'shallowequal';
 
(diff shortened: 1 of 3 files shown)


yesmeck:
Should we treat this change as a feature?

PR #5537, opened 2017-03-28
	Added isFieldTouched and isFieldsTouched types to WrappedFormUtils

krosenberg:
The [documentation](https://ant.design/components/form/#Form.create(options)) states that the Form API includes `isFieldTouched` and `isFieldsTouched` function, but they are not included in the [WrappedFormUtils](https://github.com/ant-design/ant-design/blob/master/components/form/Form.tsx#L50) type. This PR adds `isFieldTouched` and `isFieldsTouched` to the type alias based upon rc-form's function defined [here](https://github.com/react-component/form/blob/master/src/createBaseForm.js).


benjycui:
Thanks!

Issue #5620, opened 2017-04-04
date-picker组件showToday设置为false，年、月份面板显示异常！


denlaku:
### Version
2.9.0

### Environment
chrome

### Reproduction link
[https://github.com/denlaku/react-view/blob/master/2017-04-04_232024.png](https://github.com/denlaku/react-view/blob/master/2017-04-04_232024.png)

### Steps to reproduce
1.设置date-picker组件showToday设置为false；
2.点击年或者月份

### What is expected?
年或者月份能正常显示

### What is actually happening?
年或者月份的内容溢出

PR #5635, opened 2017-04-05
fix: month/year panel of DatePicker[showToday=false], close: #5620

benjycui:
Close: https://github.com/ant-design/ant-design/issues/5620

### DIFF
--- a/components/date-picker/style/MonthPanel.less
+++ b/components/date-picker/style/MonthPanel.less
@@ -1,13 +1,17 @@
 .@{calendar-prefix-cls}-month-panel {
-  left: 0;
-  top: 1px;
-  bottom: 0;
-  right: 0;
-  background: @component-background;
-  z-index: 10;
   position: absolute;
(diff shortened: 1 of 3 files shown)


components/date-picker/style/MonthPanel.less
```diff
@@ -18,10 +22,14 @@
   .calendarPanelHeader(~"@{calendar-prefix-cls}-month-panel");
 }
 
+.@{calendar-prefix-cls}-month-panel-body {
+  height: ~"calc(100% - 34px)";
```
afc163: Consider compatibility.

benjycui: ref: http://caniuse.com/#search=calc

afc163: What about IE9/10

benjycui: It works fine in IE9

image

PR #5640, opened 2017-04-05
Translates icon.md into English (completed)

kenaniah:
Adds a native English translation for the icons design spec.

**Need Help:** The systems icons section is translated, but I'm having a hard time understanding the sections that are not yet translated. Google Translate gives me a translation of "Business Icons", but I'm not sure exactly what that is supposed to mean. Is this talking about "Application Icons", custom icons that users make themselves, or something else? 

The section I'm having trouble with is below:

| Original Text | What Google Translate Says |
| ---- | ---- | 
| 业务图标 | Business icon |
| 业务图标不同于系统图标，本身不带有功能性的操作，而是辅助配合文案的一个抽象化图形。相较于系统图标，业务图标在设计的细节上更为丰富，使用的尺寸也比较大。 | The business icon is different from the system icon, which itself does not have a functional operation, but rather an abstract drawing of the auxiliary document. Compared to the system icon, the business icon in the design details are more abundant, the use of the size is relatively large. |


benjycui:
业务图标 are icons which are designed to indicate special features of your own product.

afc163:
From our designer：

`Bussiness icon`: other than system icons we provided officially, it is more like abstract simple illustration In some ways, which helps users to understand the context more easily and clearly. 

We prefer to call it `Graph icon`. And there is a picture we think it will be helpful.

kenaniah:
@benjycui @afc163 Thank you for your help on this! The translation is now complete.

Based on those definitions, I think the term "pictograph" best describes in English what 业务图标 is attempting to convey.

docs/spec/icon.en-US.md
```diff
@@ -0,0 +1,109 @@
...
+
+<img class="preview-img no-padding" align="right" src="https://zos.alipayobjects.com/rmsportal/uYhuxxiWZlgVwdmfqUli.png" description="Colors demonstrated - @Black = #000000、@White = #FFFFFF、@Blue-6 = #108EE9">
+
+The color of the icon should be consistent the color of the surrounding copy, unless the icon is being used to express state (in which case it should be colored accordingly).
+
+---
+
+## Pictographs
+
+<img class="preview-img no-padding" align="right" src="https://zos.alipayobjects.com/rmsportal/tFBnAjChLpybfxtnotTx.png">
```
afc163: You can keep [this picture](https://zos.alipayobjects.com/rmsportal/sBoyTTymmdhEuYUPEGAd.png) and add the above picture after it.

kenaniah:
I'll re-add that's picture in the next PR then ;-)

PR #5665, opened 2017-04-07
Chinese (Traditional) locale added

GeorgioWan:
Hello,

I added Traditional Chinese locale `zh_TW` and translation.


GeorgioWan:
@yesmeck Hi,
So zh_TW locale will provide on **antd@2.10** but not 2.9.1, right?

I look forward to it and thanks for merging this!! :raised_hands:

PR #5667, opened 2017-04-07
New locale with zh_TW

chenwb-m:
Hi, I am work with ant design and I really need the locale with zh_TW.
Thank you very much.


yesmeck:
#5665 Will be released this weekend.

PR #5809, opened 2017-04-19
fix: vertical tab bar height should follow tabs height

benjycui:
ref: https://github.com/ant-design/ant-design/pull/5784

主要修改为 860879f


components/tabs/style/index.less
```diff
@@ -216,6 +216,7 @@
   &-vertical {
     > .@{tab-prefix-cls}-bar {
       border-bottom: 0;
+      height: 100%;
```
benjycui: 让 TabBar 高度跟随 Tabs

PR #5855, opened 2017-04-21
Add a less variable to opt out of form item colons

megawac:
Colons disabled (@form-item-trailing-colon: false):

image

Colons enabled (@form-item-trailing-colon: true):

image

Our designer wanted to opt out of manadatory colons here

### DIFF
--- a/components/style/themes/default.less
+++ b/components/style/themes/default.less
@@ -194,6 +194,7 @@
 @label-required-color        : @highlight-color;
 @label-color                 : @text-color;
 @form-item-margin-bottom     : 24px;
+@form-item-trailing-colon    : true;
 
 // Input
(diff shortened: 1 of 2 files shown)


components/style/themes/default.less
```diff
@@ -194,6 +194,7 @@
 @label-required-color        : @highlight-color;
 @label-color                 : @text-color;
 @form-item-margin-bottom     : 24px;
+@form-item-trailing-colon    : false;
```
bang88: I think this should default to `true` ?

megawac: Whoops thanks, didnt mean to commit it as false

plandem:
I also remove colons

Issue #5877, opened 2017-04-24
Tabs break change in 2.9.2 with padding added


jaredleechn:
### Version
2.9.2

### Environment
does not matter

### Reproduction link
[https://ant.design/components/tabs-cn/#components-tabs-demo-position](https://ant.design/components/tabs-cn/#components-tabs-demo-position)

### Steps to reproduce
choose postion into left, and inspect the tabs, you will find `div.ant-tabs-nav-container` with a `padding: 32px` property. It was imported in v2.9.2 as a break change.

### What is expected?
no extra padding, no break change.

### What is actually happening?
padding added for `div.ant-tabs-nav-container`

Roughly checking the [pr](https://github.com/ant-design/ant-design/pull/5784/files#diff-eb371173aab3f298300df1fd198e0186R237), I think the author maybe want to add padding to `div.ant-tabs-nav-container.ant-tabs-nav-container-scrolling` indeed

---
@dicklwm #5784

PR #5886, opened 2017-04-25
Fix the break change #5877 

MinLeeV5:
Fix the break change #5877 
I add padding to ```div.ant-tabs-nav-container.ant-tabs-nav-container-scrolling```


benjycui:
👍

PR #5898, opened 2017-04-25
fix: should not reset showDate when value is empty array, close: #5872

benjycui:
Close: https://github.com/ant-design/ant-design/issues/5872


ladjzero:
Please review and merge asap. ;)

afc163 (approved):

PR #5978, opened 2017-05-01
Add @tabs-card-head-background

liekkas:
extract new variable @tabs-card-head-background from tabs's card-style.less for custom theme


liekkas:
@megawac @component-background is used in the active state

PR #6005, opened 2017-05-03
Use @tag-default-bg for multiselect tags

megawac:


components/style/themes/default.less
```diff
@@ -290,6 +290,10 @@
 @table-padding-vertical: 16px;
 @table-padding-horizontal: 8px;
 
+// Tag
+// --
+@tag-default-bg: @background-color-base;
```
afc163: That will change `f3f3f3` to `f7f7f7`.

megawac: `#f3f3f3` and `#f7f7f7` are almost identical colors and this will make the tags a bit more consistent

image
image

megawac:
Rebased

components/style/themes/default.less
afc163: It is not same color, we prefer to keep it `#f3f3f3` which is more accessible for more devices.

megawac: do you want me to make `@tag-default-bg` #f3f3f3 or create a different variable?

afc163: `@select-tag-bg` will be good.

megawac: I'm fine with `@select-tag-bg` but why should `select-tag-bg` not default to the same color as `tag-default-bg` or vice versa. Just seems weird to have such similar variable for such similar things

afc163: The `Tag` has a shader border for visual accessibility which select tag don't have it and `#f7f7f7` looks too light.

PR #6007, opened 2017-05-03
Fix cascader not using @component-background

megawac:
Also, its a bit of a shame that `cascader` and `select` styles aren't consistent for the input elements when they are clearable :(. It'd be really nice if they both implemented the same clearable/dropdown arrow mixins

### DIFF
--- a/components/cascader/style/index.less
+++ b/components/cascader/style/index.less
@@ -22,7 +22,7 @@
     display: inline-block;
     cursor: pointer;
     font-size: @font-size-base;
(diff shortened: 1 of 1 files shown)


benjycui:
According to me, we can add background variable for every component. And these variables extends component background by default:

```less
@component-bg: #fff;
@input-bg: @component-bg;
...
```

megawac:
It seems pretty inconsistent how its done atm, although I'll agree thats eventually how it should be

PR #6018, opened 2017-05-04
fixed issue 5834, autocomplete is not being decorated

kossel:

The cause is because the `.has-error` class is being overwrote by 2 other css rules:
* ant-select style
`.@{select-prefix-cls}-search__field`
* auto-complete style
```
.@{input-prefix-cls} {
 input();
}
```
the `input()` creates duplicated css attributes, because auto-complete component already inject a <Input /> which already has `.ant-input` class (exact same attributes)

Also updated the sample file so you can test

### DIFF
--- a/components/select/style/index.less
+++ b/components/select/style/index.less
@@ -248,7 +248,7 @@
     }
 
     .@{select-prefix-cls}-search__field {
-      border: 0;
+      border: 0 @border-style-base @border-color-base;
       font-size: 100%;
       height: 100%;
       width: 100%;
--- a/components/auto-complete/style/index.less
+++ b/components/auto-complete/style/index.less
@@ -31,7 +31,6 @@
     }
 
     .@{input-prefix-cls} {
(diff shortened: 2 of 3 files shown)


components/select/style/index.less
```diff
@@ -248,7 +248,7 @@
     }
 
     .@{select-prefix-cls}-search__field {
-      border: 0;
+      border: 0 @border-style-base @border-color-base;
```
benjycui: Actually, this line of code is confusing.

kossel: hmm actually it should be just `border-width: 0;` because `border:0` equals to `border: 0px initial initial` which render a black rectangle and overwrites the default gray color.

should I do another PR for this small change?

benjycui: Yep, thanks.

kossel: https://github.com/ant-design/ant-design/pull/6030

PR #6019, opened 2017-05-04
Cannot override default ghost button

iharpeshka:
I needed to override color default ghost button, but I did not want to do this with "css overide"....

### DIFF
--- a/components/button/style/index.less
+++ b/components/button/style/index.less
@@ -142,9 +142,7 @@
   }
 
   &-background-ghost {
-    background: transparent!important;
-    border-color: #fff;
-    color: #fff;
+    .button-variant-ghost(@btn-default-bg);
   }
 
   &-background-ghost&-primary {


components/button/style/index.less
```diff
@@ -142,9 +142,7 @@
   }
 
   &-background-ghost {
-    background: transparent!important;
```
benjycui: 我不懂，你为什么要覆盖 Ghost button 的背景色？重设计定义 Ghost 必须没有背景色。 https://ant.design/components/button-cn/#components-button-demo-ghost

iharpeshka: I want to replace the button color !!

iharpeshka: In the current implementation, it is always white color text and color btn? But why is it so? Ghost warning btn i can override....

iharpeshka:
I do a default (white) button on a white background... and i already define primary ghost button with blue color

components/button/style/index.less
iharpeshka: I do a default (white) button on a white background... and i already define primary ghost button with blue color

xueqingxiao: @iharpeshka Cause we must felllow design spec, if you want chang backgroud color & text color, you can setup styles by yourself.

PR #6030, opened 2017-05-05
Bugfix/5834/make css more specific

kossel:
changed `border: 0 @border-style-base @border-color-base;` to `border-width: 0px`
I also updated the jest snapshot


benjycui:
Please don't update snapshot.

PR #6057, opened 2017-05-08
Use the prop-types package from npm instead of React.PropTypes

deanchao:
Warning: Accessing PropTypes via the main React package is deprecated. Use the prop-types package from npm instead.

Noticed that this has been fixed in [2.9.2](https://ant.design/changelog#2.9.2) and the pull request: [#5723](https://github.com/ant-design/ant-design/pull/5723)

But there are still some remaining uses of PropTypes from React, so I create this pull request to remove the warning.

### DIFF
--- a/components/badge/index.tsx
+++ b/components/badge/index.tsx
@@ -1,4 +1,5 @@
-import React, { PropTypes } from 'react';
+import React from 'react';
+import PropTypes from 'prop-types';
 import Animate from 'rc-animate';
 import ScrollNumber from './ScrollNumber';
 import classNames from 'classnames';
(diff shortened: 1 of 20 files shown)


benjycui:
Could you just replace all the `PropTypes` in antd's codebase?

chaotyler:
@benjycui  You mean the PropTypes in site and docs? If so, just added one more commit.

qq645381995:
@chaotyler  那这个版本说明时候发布？ 发布后就没了 prop-types 的报错？ 
我的代码也需要这么改吗？
```
import React, { Component } from 'react'
import PropTypes from 'prop-types';
```

bernex:
Do you have plan to edit this issue?

chaotyler:
After all the changes, the warning still exits, and following is the reason:

The Babel function `_interopRequireWildcard` iterates through export object properties and this fires deprecation warning because of deprecated property access.

https://github.com/babel/babel/issues/4290

Issue #6119, opened 2017-05-12
Collapse 能否禁用？


petitspois:
### What problem does this feature solve?
不允许指定菜单展开

### What does the proposed API look like?
```js
<Panel disabled>
```

benjycui:
A PR is welcomed

benjycui:
antd 这边还是需要升级下依赖。

PR #6173, opened 2017-05-17
fix: Collapse.Panel should support disabled, close: #6119

benjycui:
Close: https://github.com/ant-design/ant-design/issues/6119

因为底层 rc-collapse 发的是 patch。。。。


benjycui:
Updated.

afc163 (approved):

afc163 (changes requested):
发到 2.11 上。

PR #6220, opened 2017-05-22
Add ValidationRule type for getFieldDecorator validation rules.

mitchelldemler:
Copied types from english documentation. (these might need checking?)


mitchelldemler:
Oops second commit contained a bunch of crap, will open another PR soon.

PR #6268, opened 2017-05-25
Form and Mention TypeScript fixes

jch254:

### DIFF
--- a/components/form/Form.tsx
+++ b/components/form/Form.tsx
@@ -52,7 +52,7 @@ export type ValidationRule = {
   /** transform a value before validation */
   transform?: (value: any) => any;
   /** custom validate function (Note: callback must be called) */
-  validator?: (rule: any, value: any, callback: any) => any;
+  validator?: (rule: any, value: any, callback: any, source?: any, options?: any) => any;
 };
 
 export type ValidateCallback = (erros: any, values: any) => void;
--- a/components/mention/index.tsx
+++ b/components/mention/index.tsx
@@ -5,7 +5,7 @@ import shallowequal from 'shallowequal';
 import Icon from '../icon';
 
 export interface MentionProps {
-  prefixCls: string;
+  prefixCls?: string;
   suggestionStyle?: React.CSSProperties;
   suggestions?: Array<any>;
   onSearchChange?: Function;


components/form/Form.tsx
```diff
@@ -52,7 +52,7 @@ export type ValidationRule = {
   /** transform a value before validation */
   transform?: (value: any) => any;
   /** custom validate function (Note: callback must be called) */
-  validator?: (rule: any, value: any, callback: any) => any;
+  validator?: (rule: any, value: any, callback: any, source?: any, options?: any) => any;
```
infeng: Any optional parameters must follow required parameters.

benjycui: ???

infeng: If `options` is optional, `source ` must be required.

paranoidjk: yep, @infeng is right. https://www.typescriptlang.org/docs/handbook/functions.html#optional-and-default-parameters

benjycui: Oops... I understand, could you find any solution?

benjycui: > We can get this functionality in TypeScript by adding a ? to the end of **parameters** we want to be optional.

infeng: Like this:
```typescript
  validator?(rule: any, value: any, callback: any, source: any): any;
  validator?(rule: any, value: any, callback: any, options: any): any;
  validator?(rule: any, value: any, callback: any, source: any, options: any): any;
```
But i can`t confirm `options` and `source` are all optional.

benjycui: Actually, multi optional parameters works fine, see: [demo](http://www.typescriptlang.org/play/#src=export%20interface%20INotificationService%20%7B%0A%20%20%20%20error(message%3A%20string%2C%20title%3F%3A%20string%2C%20autoHideAfter%3F%20%3A%20number)%3B%0A%7D%0A%0Aclass%20X%20%7B%0A%09error(message%3A%20string%2C%20title%3F%3A%20string%2C%20autoHideAfter%3F%3A%20number)%20%7B%0A%09%09console.log(message%2C%20title%2C%20autoHideAfter)%3B%0A%09%7D%0A%7D%0A%0Anew%20X().error(%22hi%20there%22%2C%20undefined%2C%201000)%3B)

infeng: Multi optional parameters will appear error like this:
```typescript
function foo(name?: string, age?: number) {}
// [ts] Argument of type '100' is not assignable to parameter of type 'string | undefined'.
foo(100);
```

benjycui: Actually... This means that if you want to get `options` from arguments, you must get `source` from arguments at the same time:

```tsx
validator?: (rule: any, value: any, callback: any, source?: any, options?: any) => any;
```

I think that you make a mistake. It is that `async-validator` pass arguments to `validator` and developers get what they want from arguments, not developers pass arguments to `validator`.

paranoidjk: @benjycui Even if the parameter definition is just to display the possible params that users custom validator function may received，there still should not be a optional param follow a optional param, i think the right way may be like this, `any included undefined`

```tsx
validator?: (rule: any, value: any, callback: any, source: any, options: any) => any;
````

benjycui: @paranoidjk you can PR to do so, until we can provide more specific definition for source and options.

PR #6273, opened 2017-05-26
Translate into Polish (WIP)

kocielnik:
My commit is just the last one. Some expressions have been translated without my possibility of seeing them in the original places. Maybe you can provide me with a quick way to preview them?

Thanks for your feedback!


benjycui:
Oops, please clean your PR.

PR #6308, opened 2017-05-31
ComponentDecorator return specific type instead of any

clinyong:
Test.tsx

```tsx
interface TestProps {
    name: string;
}

class Test extends React.Component<TestProps & { form: WrappedFormUtils }, {}> {}

export default Form.create<TestProps>()(Test)
```

index.tsx

```tsx
import Test from './Test'

<Test
    // can get type checking and completion here
/>
```

### DIFF
--- a/components/form/Form.tsx
+++ b/components/form/Form.tsx
@@ -111,7 +111,7 @@ export interface FormComponentProps {
 
 // https://github.com/DefinitelyTyped/DefinitelyTyped/issues/9951
 export interface ComponentDecorator<TOwnProps> {
-  (component: React.ComponentClass<FormComponentProps & TOwnProps>): any;
+  (component: React.ComponentClass<FormComponentProps & TOwnProps>): React.ComponentClass<TOwnProps>;
 }
 
 export default class Form extends React.Component<FormProps, any> {


benjycui:
👍

components/form/Form.tsx
```diff
@@ -111,7 +111,7 @@ export interface FormComponentProps {
 
 // https://github.com/DefinitelyTyped/DefinitelyTyped/issues/9951
 export interface ComponentDecorator<TOwnProps> {
-  (component: React.ComponentClass<FormComponentProps & TOwnProps>): any;
+  (component: React.ComponentClass<FormComponentProps & TOwnProps>): React.ComponentClass<TOwnProps>;
```
infeng: This change will cause `@Form.create()` throw errors, relate [this](https://github.com/ant-design/ant-design/pull/5560).

infeng:
This change will cause @Form.create() throw errors, if you want to return specific type, you can write it like this:
```typescript
interface TestProps {
    name: string;
}

class Test extends React.Component<TestProps & { form: WrappedFormUtils }, {}> {}

export default Form.create<TestProps>()(Test) as React.ComponentClass<TestProps>;
```

Issue #6392, opened 2017-06-07
Documentation: Select with search field


aaronplanell:
Hello,

The documentation of the [Select Component](https://ant.design/components/select/) explains how to search among differents options. The _'Select with search field'_ example shows:
```
  <Select
    showSearch
    style={{ width: 200 }}
    placeholder="Select a person"
    optionFilterProp="children"
    onChange={handleChange}
    filterOption={(input, option) => option.props.value.toLowerCase().indexOf(input.toLowerCase()) >= 0}
  >
    <Option value="jack">Jack</Option>
    <Option value="lucy">Lucy</Option>
    <Option value="tom">Tom</Option>
  </Select>
```

The filter option is right in this example:
```
    filterOption={(input, option) => option.props.value.toLowerCase().indexOf(input.toLowerCase()) >= 0}
```

But, it's right because the value and the children match. However, if they are different it won't work :cry: 

In my humble opinion, I think this example is more accurate (and more useful!):
```
    filterOption={(input, option) => option.props.children.toLowerCase().indexOf(input.toLowerCase()) >= 0}
```

What do you think about it?

Best regards,
Aaron Planell

yesmeck:
It's just a example to show how to implement filterOption.

aaronplanell:
Hi @yesmeck !

I totally agree, but maybe it's a little bit confusing. Probably it will be more useful with `children` instead of `value` and it will work the same way. I think other developers will enjoy this update at the documentation.

Best regards!

yesmeck:
Oh, got your point, could you send a PR?

aaronplanell:
I will try it :smile:

aaronplanell:
Sorry @yesmeck , I suppose that I must use the `feature-2.11` branch, musn't I?

yesmeck:
It ok to use master.

PR #6420, opened 2017-06-09
update document, ref #6392

aaronplanell:
Hello,

The documentation of the [Select Component](https://ant.design/components/select/) explains how to search among differents options. The _'Select with search field'_ example shows:
```
  <Select
    showSearch
    style={{ width: 200 }}
    placeholder="Select a person"
    optionFilterProp="children"
    onChange={handleChange}
    filterOption={(input, option) => option.props.value.toLowerCase().indexOf(input.toLowerCase()) >= 0}
  >
    <Option value="jack">Jack</Option>
    <Option value="lucy">Lucy</Option>
    <Option value="tom">Tom</Option>
  </Select>
```

The filter option is right in this example:
```
    filterOption={(input, option) => option.props.value.toLowerCase().indexOf(input.toLowerCase()) >= 0}
```

But, **it's right because the value and the children match**. However, if they are different it won't work :cry: 

In my humble opinion, I think this example is more accurate (and more useful!):
```
    filterOption={(input, option) => option.props.children.toLowerCase().indexOf(input.toLowerCase()) >= 0}
```

I opened the issue #6392 and @yesmeck ask me to do a PR.

Best regards,
Aaron Planell

### DIFF
--- a/components/select/demo/search.md
+++ b/components/select/demo/search.md
@@ -28,7 +28,7 @@ ReactDOM.render(
     placeholder="Select a person"
     optionFilterProp="children"
     onChange={handleChange}
-    filterOption={(input, option) => option.props.value.toLowerCase().indexOf(input.toLowerCase()) >= 0}
+    filterOption={(input, option) => option.props.children.toLowerCase().indexOf(input.toLowerCase()) >= 0}
   >
     <Option value="jack">Jack</Option>
     <Option value="lucy">Lucy</Option>


u007:
mine example uses attribute
```
<a-select
            v-model="newCountry"
            placeholder="select country"
            :show-search="true"
            :option-filter-prop="children"
            :filter-option="(input, option) => {
              return option.data.attrs.text.toLowerCase().indexOf(input.toLowerCase()) >= 0
              }"
          >
            <a-select-option
              v-for="option in countries"
              :key="option.code"
              :value="option.code"
              :text="option.name"
            >{{option.name}}</a-select-option>
          </a-select>
```

PR #6434, opened 2017-06-12
Allow Number type table keys. Make notification success, error, info, warn and warning non-optional

jch254:
Re-open #6350:

Optional notification functions are frustrating and require a non-null check with strict TS:
```
if (notification.error) {
  notification.error({
    message: 'Oops!',
    description: 'You do not have permission to view that page, please contact the administrator.',
  });
}
```

DefinitelyTyped/React defines the Key type as `string | number` (see https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react/index.d.ts#L31). Antd Table supports Number typed keys but TS interfaces do not.

### DIFF
--- a/components/notification/index.tsx
+++ b/components/notification/index.tsx
@@ -148,18 +148,19 @@ function notice(args) {
   });
 }
 
-const api: {
-  success?(args: ArgsProps): void;
-  error?(args: ArgsProps): void;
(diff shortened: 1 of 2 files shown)


yesmeck:
Thanks!

PR #6435, opened 2017-06-12
Add warning about declaring Table class in render function

jch254:
I followed the 'Using in TypeScript' instructions and hit an issue where the Table always rerendered. This was because I had declared `class UserTable extends Table<IUser> {}` inside the render function. Moving `class UserTable extends Table<IUser> {}` outside the render function fixed the issue.

### DIFF
--- a/components/table/index.en-US.md
+++ b/components/table/index.en-US.md
@@ -154,6 +154,7 @@ const data: IUser[] = [{
   name: 'Jack',
 }];
 
+// Do not declare class inside render function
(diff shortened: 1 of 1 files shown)


benjycui:
...

I think this is common knowledge of React

yesmeck:
Aha, It's not about TypeScript, you should not do this in JavaScript too.

jch254:
Thanks no probs, at least it's searchable now 👍

PR #6536, opened 2017-06-20
Fix importing Collapse in TypeScript

jch254:
The `antd/lib/collapse/index.js` file exports as below:

```
...
exports['default'] = Collapse;

Collapse.Panel = _rcCollapse2['default'].Panel;
Collapse.defaultProps = {
    prefixCls: 'ant-collapse',
    bordered: true,
    openAnimation: Object.assign({}, _openAnimation2['default'], {
        appear: function appear() {}
    })
};
```

Calls to `typeof Collapse` return Object. Moving to a separate file and importing via `index.tsx` fixes this and is the same pattern used in other components (e.g. Form).


benjycui:
Just for code style.

huachaoliu:
Ie browser dose not support Object.assign

Issue #6548, opened 2017-06-20
fix disabled dropdown submenu style error


laispace:
### Version
antd-v2.11.1

### Environment
antd-v2.11.1

### Reproduction link
[https://codepen.io/laispace/pen/weJLjQ](https://codepen.io/laispace/pen/weJLjQ)

### Steps to reproduce
Use `Submenu` that is disabled inside `Dropdown`.

### What is expected?
The `Submenu` with correct disabled style, disabling cursor pointer.

### What is actually happening?
The `Submenu`'s disabled style is not right.

PR #6549, opened 2017-06-20
fix disabled dropdown submenu style error

laispace:
Close: https://github.com/ant-design/ant-design/issues/6548

### DIFF
--- a/components/dropdown/style/index.less
+++ b/components/dropdown/style/index.less
@@ -49,7 +49,7 @@
     background-clip: padding-box;
 
     &-item,
-    &-submenu-title {
+    &-submenu {
       padding: 7px 8px;
       margin: 0;
       clear: both;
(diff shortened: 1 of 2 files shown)


afc163:
image

laispace:
@afc163 已修正这个错误信息

components/dropdown/style/index.less
```diff
@@ -49,7 +49,7 @@
     background-clip: padding-box;
 
     &-item,
-    &-submenu-title {
+    &-submenu {
```
benjycui: Why did you remove this?

laispace: I compared Menu's style and Dropdown's style, finding that this line fixed the problem, without side effects.

laispace: If I am wrong, maybe you can refuse the PR, but I still want to know what's the correct way to fix the problem.

benjycui:
Close in 7ff0573

PR #6619, opened 2017-06-27
fix dataPicker onOk type definition

newraina:
onOk 回调里可以接受参数，详见：[rc-calendar](https://github.com/react-component/calendar/blob/master/src/Calendar.jsx#L165)

但是 TS 定义里没有反应出来，导致使用时如果在回调里带参数会报错。

### DIFF
--- a/components/date-picker/index.tsx
+++ b/components/date-picker/index.tsx
@@ -44,7 +44,7 @@ export interface DatePickerProps extends PickerProps, SinglePickerProps {
     disabledSeconds?: () => [number, number],
   };
   onOpenChange?: (status: boolean) => void;
-  onOk?: () => void;
+  onOk?: (selectedTime: moment.Moment) => void;
   placeholder?: string;
 }
 const DatePicker = wrapPicker(createPicker(RcCalendar)) as React.ClassicComponentClass<DatePickerProps>;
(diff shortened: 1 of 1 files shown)


components/date-picker/index.tsx
```diff
@@ -44,7 +44,7 @@ export interface DatePickerProps extends PickerProps, SinglePickerProps {
     disabledSeconds?: () => [number, number],
   };
   onOpenChange?: (status: boolean) => void;
-  onOk?: () => void;
+  onOk?: (selectedTime: moment.Moment) => void;
```
bang88: Please consider make it optional or this will break the following markup
```
onOK={()=>void}
```

components/date-picker/index.tsx
```diff
@@ -59,7 +59,7 @@ export interface RangePickerProps extends PickerProps {
   defaultValue?: [moment.Moment, moment.Moment];
   defaultPickerValue?: [moment.Moment, moment.Moment];
   onChange?: (dates: [moment.Moment, moment.Moment], dateStrings: [string, string]) => void;
-  onOk?: () => void;
+  onOk?: (selectedTime: moment.Moment) => void;
```
bang88: same here.

BANG88 (changes requested):

components/date-picker/index.tsx
newraina: onOk is a callback function, base on [TS docs/Optional Parameters in Callbacks](https://www.typescriptlang.org/docs/handbook/declaration-files/do-s-and-don-ts.html):

> but there’s no need to make the parameter optional to accomplish this – it’s always legal to provide a callback that accepts fewer arguments.

No need to make `selectedTime` optional, you can try this: [playground](https://www.typescriptlang.org/play/index.html#src=const%20eat%20%3D%20(cb%3A%20(something%3A%20string)%20%3D%3E%20void)%20%3D%3E%20%7B%20%7D%0D%0Aconst%20something%20%3D%20()%20%3D%3E%20%7B%7D%0D%0A%0D%0Aeat(something))

components/date-picker/index.tsx
bang88: Yep, a few confusing, but you are right.

BANG88 (approved):

PR #6646, opened 2017-06-28
Establish a ref to the select component

nachocodoner:
I have just established a ref to the `select` component in order to access to the `rs-select` component definition and be able to customize the behaviors for such component when rendering in a HOC. I think `refs` should be provided broadly for all components for that purpose.

### DIFF
--- a/components/select/index.tsx
+++ b/components/select/index.tsx
@@ -148,6 +148,7 @@ export default class Select extends React.Component<SelectProps, any> {
       <RcSelect
         {...restProps}
         {...modeConfig}
+        ref="select"
         prefixCls={prefixCls}
         className={cls}
         optionLabelProp={optionLabelProp || 'children'}


benjycui:
Why? There is no methods on instances of Select.

Gywem:
There are methods on the instance of the `rc-select` that antd uses. You [can check out](https://github.com/react-component/select/blob/master/src/Select.jsx) the implementation and see that you have a full API to control different stuff around the select.
That way, we can tweak the select in a HOC to behave differently, for instance, to close the popup all time we select an item and many things more.

benjycui:
> There are methods on the instance of the rc-select that antd uses.

For example?



> You can check out the implementation and see that you have a full API to control different stuff around the select.

It is not recommend use these methods, they are not documented, and they will be refactor anytime.

Gywem:
There are methods like [`setOpenState`](https://github.com/react-component/select/blob/master/src/Select.jsx#L521) that may be useful to tweak the behavior of the select in a HOC layer. For instance, what I want to accomplish is to close the popup when selecting an item.

> It is not recommend use these methods, they are not documented, and they will be refactor anytime.

That is true, but a component should be open to be extensible, and is in other's responsibility to risk on modify them and face possible refactors of the component's API in the future. What I just propose is to put a reference to raise possibilities and not restrictions.

What I feel is just a pity to lose the possibility to have a select already implemented and with antd styling, but not being able to extend the behavior.  The `rc-select` implementation is open to extensibility, as you can see they use reference for every element, but antd layer restrict that communication in this moment.

benjycui:
Nope, this PR is dangerous, so I will not merge it.



> There are methods like setOpenState that may be useful to tweak the behavior of the select in a HOC layer.

We need to make Select's open to be controllable, but it is difficult for the code is to complicated.

PR #6711, opened 2017-07-03
Update Avatar Page documentation Fix

llaski:
Quick fix for updating the API property section - currently `type` property but it should be `shape`


yesmeck:
Thanks.

PR #6788, opened 2017-07-10
Fixing typos

NoHomey:
It should be "Добре" instead of "Дибре".

Just like `filterConfirm`

Reference: https://translate.google.bg/#en/bg/Ok


yesmeck:
Thanks.

PR #6813, opened 2017-07-13
Fix backTop noAny

chenshuai2144:
Update TypeScript definition for the component.


benjycui:
Please clean your PR, you should not include other's commit in your PR.

afc163 (changes requested):
like @benjycui said

PR #6880, opened 2017-07-18
Modal.confirm supports zIndex prop #6722

Alex1990:


components/modal/Modal.tsx
```diff
@@ -40,6 +40,7 @@ export interface ModalProps {
   transitionName?: string;
   className?: string;
   getContainer?: (instance: React.ReactInstance) => HTMLElement;
+  zIndex?: boolean;
```
benjycui: I think you also need to add `zIndex` to [here](https://github.com/ant-design/ant-design/blob/edf8a4f239df597b34ad3289586a411dcf7a93f2/components/modal/Modal.tsx#L48-L60).

benjycui:
@Alex1990 下次记得补充文档 e552978

Alex1990:
好得

Issue #5627, opened 2017-04-05
Enable noImplicitAny in tsconfig.json


yesmeck:
There are many implicit `any` types in our codebase, we should give these variables a appropriate type declaration for better TypeScript support.

Ref: https://github.com/ant-design/ant-design/pull/5609#discussion_r109763481

---

If someone is interested in this issue, just reply this issue which component you are going to handle:

* [x] back-top https://github.com/ant-design/ant-design/pull/6835
* [x] carousel
* [x] dropdown
* [x] input
* [x] message
* [x] progress
* [x] spin
* [x] tag
* [x] tree-select
* [x] badge
* [x] cascader
* [x] form
* [x] input-number
* [x] modal
* [x] radio
* [x] steps
* [x] time-picker
* [x] upload
* [x] affix
* [x] breadcrumb
* [x] checkbox
* [x] grid
* [x] layout
* [x] notification
* [x] rate
* [x] timeline
* [x] version
* [x] alert @bang88  https://github.com/ant-design/ant-design/pull/5836
* [x] anchor
* [x] button
* [x] col
* [x] icon
* [x] locale-provider
* [x] pagination
* [x] row
* [x] switch
* [x] tooltip
* [x] calendar
* [x] collapse
* [x] mention
* [x] popconfirm
* [x] select
* [x] table
* [x] transfer
* [x] auto-complete
* [x] card
* [x] date-picker
* [x] menu
* [x] popover
* [x] slider
* [x] tabs
* [x] tree

bang88:
Any updates ?

bang88:
```sh
11 function getTargetRect(target): ClientRect {
                          ~~~~~~

components/affix/index.tsx(11,24): error TS7006: Parameter 'target' implicitly has an 'any' type.
```
source code
```tsx
function getTargetRect(target): ClientRect {
  // this check is for SSR ?
  return target !== window ?
    target.getBoundingClientRect() :
    { top: 0, left: 0, bottom: 0 };
}
```

benjycui:
> this check is for SSR ?

Yep

bang88:
Should be provide a function to do this work. Don't check target is window.
We just need to know the current environment is server or client ?

yesmeck:
canUseDom? https://github.com/react-component/util#domcanusedom

bang88:
Yep, some thing like this.

benjycui:
If someone is interested in this issue, you can reply here that which component you are going to handle:

[...]

PR #6925, opened 2017-07-22
remove `any` from Card component #5627

djyde:


djyde:
The declaration for `rc-util/addEventListener` should be in `custom-typings`, but I just found a problem https://github.com/ant-design/ant-design/issues/6926

PR #6935, opened 2017-07-24
add rc-util declaration to custom-typings

djyde:

### DIFF
--- a/typings/custom-typings.d.ts
+++ b/typings/custom-typings.d.ts
@@ -12,6 +12,17 @@ declare module "omit.js";
 
 declare module 'rc-animate';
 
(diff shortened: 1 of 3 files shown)


typings/custom-typings.d.ts
```diff
@@ -12,6 +12,17 @@ declare module "omit.js";
 
 declare module 'rc-animate';
 
+declare module 'rc-util/lib/Dom/addEventListener' {
```
benjycui: Just PR to rc-util?

djyde: There are lots of methods in `rc-util` that I am not able to cover them in a short time.

benjycui: Don't worry, take your time.

Issue #6537, opened 2017-06-20
Table fixed column don't align when scroll


shenqihui:
### Version
2.11.1

### Environment
mac 10.12.4
chrome 58.0.3029.110 (Official Build) (64-bit)

### Reproduction link
[https://codepen.io/shenqihui/pen/eRvopB](https://codepen.io/shenqihui/pen/eRvopB)

### Steps to reproduce
method 1: not move mouse and direct scroll the page , large probability to reproduce 
method 2: change chrome into ipad simulate and scroll, feel like it 100% reproduce
方式一：不移动鼠标直接滚动，很大概率重现。
方式二：将chrome改成ipad模拟器调试形式，感觉是100%几率重现。

### What is expected?
fixed columns fix align with other columns.
fixed 的列和其他列应该同时滚动，对齐

### What is actually happening?
fixed columns  don't fix align with other columns
fixed 列和其他列不对齐。

shenqihui:
i found the problem at rc-table project at file [https://github.com/react-component/table/blob/master/src/Table.jsx#L692](https://github.com/react-component/table/blob/master/src/Table.jsx#L692)

---

when it scroll, if will detect the `e.target` and `this.scrollTarget`, if it is the same, and set it scroll will take effect.

but at first ,it should first set `this.scrollTarget` at `onMouseOver` or `onTouchStart` event. 

So this explain that when i don't move the mouse and just scroll and it won't trigger `onMouseOver` or `onTouchStart`  event.

shenqihui:
Wow i resolve this problem  with violence by replace that if condition . 
As I'm using `roadhog`, so I config it‘s `webpack.config.js` , code like this:

```
module.exports = (config, env) => {
  config.module.loaders.push({
    test: /rc-table\/lib\/Table\.js$/,
    loader: 'string-replace',
    query: {
      search: 'e.target !== _this.scrollTarget',
      replace: 'false',
    },
  });
retutn config;
}
```

And i dont know why it should first detact `e.target` and `this.scrollTarget`.
Could You Tell Me why and how to Reproduction ? 
I will try to resolve that problem.

shenqihui:
目前采用三楼的替换之后，发现纵向有滚动的时候，横向进行滚动，会回到头部。

所以再加了一个判断处理，避免横向滚动的时候的问题。
```
  config.module.loaders.push({
    test: /rc-table\/lib\/Table\.js$/,
    loader: 'string-replace',
    query: {
[...]

PR #6937, opened 2017-07-24
Fix Table fixed column don't align when scroll

shenqihui:
Desc at https://github.com/ant-design/ant-design/issues/6537 .

Fix at https://github.com/react-component/table/pull/152 and publish at
rc-table 5.4.2 .


yesmeck:
Don't need update patch version, just reinstall your node_modules.

PR #6976, opened 2017-07-27
check this.container exists updating WiderPadding

neekey:
before doing updating wider padding for Card, always check if the container exists first.

Extra checklist:

**if** *isBugFix* **:**

### DIFF
--- a/components/card/index.tsx
+++ b/components/card/index.tsx
@@ -41,17 +41,19 @@ export default class Card extends Component<CardProps> {
   }
   @throttleByAnimationFrameDecorator()
   updateWiderPadding() {
-    // 936 is a magic card width pixer number indicated by designer
-    const WIDTH_BOUDARY_PX = 936;
-    if (this.container.offsetWidth >= WIDTH_BOUDARY_PX && !this.state.widerPadding) {
-      this.setState({ widerPadding: true }, () => {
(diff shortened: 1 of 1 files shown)


neekey:
I got this error when using Card with a loading component that I created:

image

It is caused by the absence of the `this.container` when `updatingWiderPadding(0` was invoked.

When dealing with actual element, always check if it exists.

neekey:
hmm... I will take a look at why the line [46](https://codecov.io/gh/ant-design/ant-design/pull/6976/src/components/card/index.tsx#L46) is not covered.

components/card/index.tsx
```diff
@@ -41,17 +41,19 @@ export default class Card extends Component<CardProps> {
   }
   @throttleByAnimationFrameDecorator()
   updateWiderPadding() {
-    // 936 is a magic card width pixer number indicated by designer
```
afc163:
```jsx

if (!this.container) {
  return;
}
```

yesmeck:
A more properer fix is cancel the throttle invoking before component unmount.

afc163:
fixed in https://github.com/ant-design/ant-design/commit/b54a70bc8e1ea8dda83661eb831f7ce52039730c

PR #7072, opened 2017-08-03
Object cloning in flatArray is performed only if object has 'childrenName' field.

dbolotin:
`spread`-style object cloning leads to object `prototype` change, which makes it impossible to use custom class objects as `dataSource` for Table component.

At least, cloning is not required if there is no `childrenName` field.

I can also change:
```
const newItem = { ...item };
```
to 
```
const newItem = Object.create(item);
```
to preserve `prototype` even in case with nested rows, as suggested [here](https://stackoverflow.com/a/28152032/769192), but I'm not sure whether it will brake something.

### DIFF
--- a/components/table/util.tsx
+++ b/components/table/util.tsx
@@ -4,11 +4,15 @@ export function flatArray(data: Object[] = [], childrenName = 'children') {
   const result: Object[] = [];
   const loop = (array) => {
     array.forEach(item => {
-      const newItem = { ...item };
-      delete newItem[childrenName];
-      result.push(newItem);
-      if (item[childrenName] && item[childrenName].length > 0) {
-        loop(item[childrenName]);
+      if (item[childrenName]) {
+        const newItem = { ...item };
+        delete newItem[childrenName];
+        result.push(newItem);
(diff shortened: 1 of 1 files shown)


benjycui:
Why do you care about type? Use case?

dbolotin:
I'am using antd through scala-js, and using array of scala-js classes as data source for the table.

Exact problem I encounter was the following: I set `rowKey` as lambda that extracts particular `id` field value from data object. This failed because [here](https://github.com/ant-design/ant-design/blob/master/components/table/Table.tsx#L645) data is flatten (which leads to type stripping), then in `SelectionCheckboxAll` `data` elements are used to extract keys using `this. getRecordKey`. So my lambda (scala-js lambda) receives object with stripped type, which leads to crash, because the object don't have prototype scala-js expects.

The same issue will lead to problem if one uses methods to extract certain values from data objects.

I checked behaviour with corrected version, and now everything works as expected, I can use scala-js lambdas.

P.S. Aside form functionality, this solution is more performant, because in most cases (I believe so) users of antd don't use nested rows, and there is no need whatsoever to make a copy of each data element on each render.

PR #7125, opened 2017-08-07
Add sr_RS locale (Serbian)

paunovic-stefan:
Added Serbian locale.

Separate PR created for rc-pagination component with the corresponding locale.
https://github.com/react-component/pagination/pull/102
merged in v1.10.4

Separate PR created for rc-calendar component with the corresponding locale.
https://github.com/react-component/calendar/pull/287
merged in v9.1.3


paunovic-stefan:
Serbian localization was added to rc-calendar in v9.1.3, but package.json for ant.design has it set to "rc-calendar": "~9.0.0" and thus not allowing the installation of 9.1.x versions.

I'm guessing there was a reason to set it to v9.0.x.

Is there a plan to start using v9.1.x of rc-calendar, or should I move the localization done in rc-calendar directly to ant.design?

paunovic-stefan:
I've added sr_RS locale from rc-calendar in components/date-picker/locale/sr_RS.tsx instead of importing it from rc-calendar so that the build will succeed until rc-calendar dependency is updated to equal or grated than v9.1.3 that contains the sr_RS locale

components/date-picker/locale/sr_RS.tsx
```diff
@@ -0,0 +1,45 @@
+// import CalendarLocale from 'rc-calendar/lib/locale/sr_RS';
```
benjycui: rc-calendar@9.0.1 which includes sr_RS locale had been published.

PR #7226, opened 2017-08-16
[BugFix] Make sure props's className are after ant's className.

iineva:


benjycui:
Why? It's not important.

PR #7592, opened 2017-09-14
Added react and react-dom as peerDeps

pastelsky:
Adds `react` and `react-dom` as peerDependencies for `antd`. Users that try to install antd without react will get helpful install-time warnings.

Some tools like [bundlephobia.com](https://bundlephobia.com) also use the `peerDependency` field to understand which dependencies are external to the package, and shouldn't be accounted for when calculating bundle size. Without this hint, it would fail.

### DIFF
--- a/package.json
+++ b/package.json
@@ -35,8 +35,8 @@
   "typings": "lib/index.d.ts",
   "license": "MIT",
   "peerDependencies": {
-    "react": ">=15.0.0",
-    "react-dom": ">=15.0.0"
+    "react": "~0.14.0 || >=15.0.0",
+    "react-dom": "~0.14.0 || >=15.0.0"
   },
   "dependencies": {
     "array-tree-filter": "~1.0.0",
--- a/package.json
+++ b/package.json
@@ -34,6 +34,10 @@
   ],
   "typings": "lib/index.d.ts",
   "license": "MIT",
+  "peerDependencies": {
+    "react": ">=15.0.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -34,6 +34,10 @@
   ],
   "typings": "lib/index.d.ts",
   "license": "MIT",
+  "peerDependencies": {
+    "react": ">=15.0.0",
+    "react-dom": ">=15.0.0"
```
yesmeck: I think we should set version to `*`.

pastelsky: Are older react versions supported?

yesmeck: I'm not sure, but antd doesn't require a particular version of react currently.

pastelsky: Setting it to `>=15.0.0` should cover all modern react versions. I picked it up from the `devDependecies` of `antd`'s package.json itself. I'm guessing all of the tests for `antd` assume `react>=15.0.0`. If you feel otherwise, we could downgrade.

yesmeck: I tested antd against React 0.14, all tests can pass after https://github.com/react-component/calendar/pull/298 merged.

yesmeck: We can set it to `"~0.14.0 || >=15.0.0"`

nikogu: @pastelsky can you set it to "~0.14.0 || >=15.0.0" and repull

pastelsky:
Done.

PR #7694, opened 2017-09-21
update for Dutch locale-provider: added missing translations

kstiopin:
Not a bugfix or a feature so I'm not sure to which branch this PR should be made but it's not too much staff here so I hope you could clarify for me if I have to do smth different.


benjycui:
Thanks.

PR #7853, opened 2017-10-11
List Optimize fixed: #4905

nikogu:
fixed: https://github.com/ant-design/ant-design/issues/4905#issuecomment-332442843


nikogu:
这下应该好了 =。-

PR #7874, opened 2017-10-12
feat: duration argument is optional

monkindey:
fix https://github.com/ant-design/ant-design/issues/7857

**Any wrong, please let me know.**

### DIFF
--- a/components/message/index.tsx
+++ b/components/message/index.tsx
@@ -29,7 +29,7 @@ type NoticeType = 'info' | 'success' | 'error' | 'warning' | 'loading';
 
 function notice(
   content: React.ReactNode,
-  duration: number = defaultDuration,
+  duration: (() => void) | number = defaultDuration,
   type: NoticeType,
   onClose?: () => void,
 ) {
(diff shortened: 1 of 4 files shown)


components/message/index.tsx
```diff
@@ -29,7 +29,7 @@ type NoticeType = 'info' | 'success' | 'error' | 'warning' | 'loading';
 
 function notice(
   content: React.ReactNode,
-  duration: number = defaultDuration,
+  duration: (() => void) | number = defaultDuration,
```
infeng: Supply multiple function types for the same function as a list of overloads will be better.

monkindey: Thanks for your review, you mean ?

```js
duration: (() => void) | number = defaultDuration,
```

to 

```js
const fn = () => void;
...

duration: fn | number = defaultDuration,
```

abenhamdine: > Supply multiple function types for the same function as a list of overloads will be better.  

why is it better ?

infeng: Sorry, I got it wrong, you are right.

abenhamdine: No problem, just asking ;)

abenhamdine (approved):

monkindey:
Thanks to be merged

PR #7879, opened 2017-10-12
Add new locale (ar-EG: Arabic Egypt) #2

mseada94:
This is a pull request for the same changes here #7841 , after rebase changes from  0fb3b6e to **antd-3.0**.
I have no idea about conflict in **components/upload/Upload.tsx** , this change comes because of rebase.

If there still any problem, i can add my changes in new commit starting from the currunt **antd-3.0**, and make a new pull request, so please tell me if that was the case.

--------
First of all, thank you for your contribution! :-)

Please makes sure that these checkboxes are checked before submitting your PR, thank you!

* [x] Make sure that you propose PR to right branch: bugfix for `master`, feature for latest active branch `feature-x.x`.
* [x] Make sure that you follow antd's [code convention](https://github.com/ant-design/ant-design/wiki/Code-convention-for-antd).
* [x] Run `npm run lint` and fix those errors before submitting in order to keep consistent code style.
* [x] Rebase before creating a PR to keep commit history clear.
* [x] Add some descriptions and refer relative issues for you PR.

Extra checklist:

**if** *isBugFix* **:**

  * [ ] Make sure that you add at least one unit test for the bug which you had fixed.

**elif** *isNewFeature* **:**

  * [ ] Update API docs for the component.
  * [ ] Update/Add demo to demonstrate new feature.
  * [ ] Update TypeScript definition for the component.
  * [ ] Add unit tests for the feature.

PR #7888, opened 2017-10-13
Add new locale (ar-EG: Arabic Egypt) #3

mseada94:
This is a pull request for the same changes in #7841 and #7879 , with clean history and no conflict.

Extra checklist:

**if** *isBugFix* **:**

**elif** *isNewFeature* **:**


mohamed-seada-1994:
I think the snapshot will be updated if i move the new local to the end of the array, but that does not work.
It seems that the snapshot should be generated with the `--updateSnapshot` flag.
I have some problems with my environment, so i can't re-generate the snapshot.
Please re-generate it and commit the new one, Thanks!

PR #8077, opened 2017-10-31
Detect Sider even if it's not first child of layout

### LINKED ISSUE #7579: how to implement sidebar in <div>
### Version
2.13.1

### Environment
Windows 10 x64, Chrom 61, react 15.4, react-router 4

### Reproduction link
[https://codepen.io/anon/pen/XeWPog?editors=001](https://codepen.io/anon/pen/XeWPog?editors=001)

### Steps to reproduce
Create CustomSider

```
const CustomSider = ({sensors, dispatch}) => {
  // console.log({sensors.sensors});
  return (
    <Sider style={{background: '#fff', borderRight: 'solid', borderColor:'#ddd', borderWidth:'5px'}}>
      <Menu mode="inline" theme="light">
          <Menu.Item key="dev1">Device 1</Menu.Item>
          <Menu.Item key="dev2">Device 2</Menu.Item>
          <Menu.Item key="dev3">Device 3</Menu.Item>
      </Menu>
    </Sider>
  );
}
```

Then add `CustomSider.__ANT_LAYOUT_SIDER = true`.

For the DOMRender, add

```
ReactDOM.render(
  <Layout>
    <div><CustomSider></CustomSider></div>
    <Content>Content</Content>
  </Layout>
, mountNode);
```

That's all.

### What is expected?
Content to the right of Sider

Content under Sider

---
This happened when I want to put CustomSider in my <Route/> to hide the CustomSider in some routes.

yasinuslu:
close #7579

### DIFF
--- a/components/layout/Sider.tsx
+++ b/components/layout/Sider.tsx
@@ -43,6 +43,14 @@ export interface SiderProps {
   breakpoint?: 'xs' | 'sm' | 'md' | 'lg' | 'xl';
 }
 
+const generateId = (() => {
+  let i = 0;
+  return (prefix: string = '') => {
+    i += 1;
+    return `${prefix}${i}`;
+  };
+})();
+
 export default class Sider extends React.Component<SiderProps, any> {
   static __ANT_LAYOUT_SIDER: any = true;
 
@@ -60,10 +68,16 @@ export default class Sider extends React.Component<SiderProps, any> {
     siderCollapsed: PropTypes.bool,
   };
 
+  static contextTypes = {
+    siderHook: PropTypes.object,
+  };
+
   private mql: any;
+  private uniqueId: string;
 
   constructor(props) {
     super(props);
+    this.uniqueId = generateId('ant-sider-');
     let matchMedia;
     if (typeof window !== 'undefined') {
       matchMedia = window.matchMedia;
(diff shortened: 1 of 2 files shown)


yasinuslu:
Just noticed failing tests, working on it.

yasinuslu:
I don't know what we can do about unit tests, even if i call `addSider` and `removeSider` in `constructor`, `Layout` still renders without `ant-layout-has-sider` class because `render` in enzyme doesn't support rerendering. 

If we want to use this method to detect `Sider` `ant-layout-has-sider` class is added to layout only after rerendering because we register `Sider` to `Layout` by adding it to state.

ddcat1115:
Try `npm run test -- -u` and commit the updated snap.

yasinuslu:
The problem is that, if i update snapshots this div will not have `ant-layout-has-sider` but the right behaviour is that it should have it. In browser environment it has this class, if we call hook in `componentWillMount` instead of `componentDidMount` it will also have it in server-side rendering too, but still it would not match this snapshot in tests.

image

So is it ok for me to update snapshots and call it a day ?

ddcat1115:
You can update the snapshot and add a new test case for the class detecting, by using `mount` in enzyme.

ddcat1115:
@yasinuslu ping~

afc163:
Added test case: https://github.com/ant-design/ant-design/commit/0eca38ba2ff9fa154505595f78422c8c17bc78b9

PR #8166, opened 2017-11-10
Add a less variable to control processing colors

megawac:
The designers that be at my company would love the ability to control the `processing` colors of steps, progress bars and badges (they wanted progress to be green instead of blue :)

This pull request adds a less variable similar to other states (e.g. success, error, warning, etc) used throughout antd.

### DIFF
--- a/components/steps/style/index.less
+++ b/components/steps/style/index.less
@@ -2,7 +2,7 @@
 @import "../../style/mixins/index";
 
 @steps-prefix-cls: ~"@{ant-prefix}-steps";
-@process-icon-color: @primary-color;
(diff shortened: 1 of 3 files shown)


components/steps/style/index.less
```diff
@@ -2,7 +2,7 @@
 @import "../../style/mixins/index";
 
 @steps-prefix-cls: ~"@{ant-prefix}-steps";
-@process-icon-color: @primary-color;
+@process-icon-color: @processing-color;
```
megawac: In 3.0 can we renaming `process` -> `processing` (or vice versa) for consistency?

PR #8177, opened 2017-11-12
Added a few explicit types.

abramobagnara:
I've added a few trivial explicit types whose lack cause build errors with tsc --noImplicitAny when using ant design as a library.


yesmeck:
I think's they should have a more specific type.

abramobagnara:
From what I see in the context the related uses have "any" type.
However the PR is a step forward (as it fixes build issues), a better type might be specified later.

yesmeck:
Fixed in `antd-3.0` branch.

PR #8351, opened 2017-11-28
Remove deprecations

yesmeck:


benjycui:
Note: 需要更新 changelog https://github.com/ant-design/ant-design/issues/8233

yesmeck:
Updated

components/form/index.en-US.md
```diff
@@ -9,7 +9,7 @@ Form is used to collect, validate, and submit the user input, usually contains v
 
 ## Form
 
-You can align the controls of a `form` using one of the following attributes：
+You can align the controls of a `form` using one of the `layout` prop：
```
megawac: This would be better written as

> You can align the controls of the `form using the `layout` prop:

benjycui (approved):

components/input/index.en-US.md
```diff
@@ -26,7 +26,7 @@ Keyboard and mouse can be used for providing or changing data.
 | prefix | The prefix icon for the Input. | string\|ReactNode |  |
 | size | The size of the input box. Note: in the context of a form, the `large` size is used. Available: `large` `default` `small` | string | `default` |
 | suffix | The suffix icon for the Input. | string\|ReactNode |  |
-| type | The type of input, `text` or `textarea`(`type=textarea` are deprecated after `2.12`, please use `Input.TextArea`) | string | `text` |
+| type | The type of input (`type=textarea` has removed after `3.0`, please use `Input.TextArea`) | string | `text` |
```
megawac: Would this be better kept just in the changelog? I think this will be confusing to have in the documentation for input

megawac (changes requested):

components/form/index.en-US.md
yesmeck: Updated. 😄

components/input/index.en-US.md
```diff
@@ -26,7 +26,7 @@ Keyboard and mouse can be used for providing or changing data.
 | prefix | The prefix icon for the Input. | string\|ReactNode |  |
 | size | The size of the input box. Note: in the context of a form, the `large` size is used. Available: `large` `default` `small` | string | `default` |
 | suffix | The suffix icon for the Input. | string\|ReactNode |  |
-| type | The type of input, `text` or `textarea`(`type=textarea` are deprecated after `2.12`, please use `Input.TextArea`) | string | `text` |
+| type | The type of input (use `Input.TextArea` instead of `type="textarea"`) | string | `text` |
```
megawac: So can the `type` attribute be entirely removed in 3.0?

megawac (approved):

components/input/index.en-US.md
yesmeck: `type` has other values like `password`, `email` etc.

megawac: Ahhh, right. Would it be worth including the types that are valid (I think the `use `Input.TextArea` is worth including!)

yesmeck: It's a long list, and `type` is just a standard HTML `attribute`, so I'm keeping it as it is.

megawac: 👍 maybe a link to https://developer.mozilla.org/docs/Web/HTML/Element/input#Form_%3Cinput%3E_types would be worth referencing

yesmeck: 👍 Added.

megawac:
LGTM 👍!

PR #8511, opened 2017-12-08
Add missing type definitions for table #8507

### LINKED ISSUE #8507: defaultExpandAllRows missing in typescript TableProps<T>
### Version
3.0.0

### Environment
windows 10

### Reproduction link
[http:///localhost.com](http:///localhost.com)

### Steps to reproduce
```  
return <Table
            columns={this.props.columns}
            dataSource={this.props.data}
            defaultExpandAllRows={true}
            expandedRowRender={(record) => this.expandedRowRender(record.key)}
        />
```

// typescript compiler complains because missing in definition

### What is expected?
Need to be added to TableProps<T> in interface.tsx.

error given by typescript compiler when used

danedavid:
Fixes #8507

### DIFF
--- a/components/table/interface.tsx
+++ b/components/table/interface.tsx
@@ -82,10 +82,12 @@ export interface TableProps<T> {
   rowKey?: string | ((record: T, index: number) => string);
   rowClassName?: (record: T, index: number) => string;
   expandedRowRender?: any;
+  defaultExpandAllRows: boolean;
   defaultExpandedRowKeys?: string[] | number[];
   expandedRowKeys?: string[] | number[];
   expandIconAsCell?: boolean;
   expandIconColumnIndex?: number;
+  expandRowByClick: boolean;
   onExpandedRowsChange?: (expandedRowKeys: string[] | number[]) => void;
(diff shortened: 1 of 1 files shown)


components/table/interface.tsx
```diff
@@ -82,10 +82,12 @@ export interface TableProps<T> {
   rowKey?: string | ((record: T, index: number) => string);
   rowClassName?: (record: T, index: number) => string;
   expandedRowRender?: any;
+  defaultExpandAllRows: boolean;
```
yesmeck: should be optional

components/table/interface.tsx
```diff
@@ -82,10 +82,12 @@ export interface TableProps<T> {
...
   defaultExpandedRowKeys?: string[] | number[];
   expandedRowKeys?: string[] | number[];
   expandIconAsCell?: boolean;
   expandIconColumnIndex?: number;
+  expandRowByClick: boolean;
```
yesmeck: same here

yesmeck:
Could you create a another PR? These two props should be optional.

danedavid:
Ok I thought I should do it here. Will raise another

PR #8550, opened 2017-12-11
Revise contributing document

yesmeck:
Can we add this document to website?

[Rendered](https://github.com/ant-design/ant-design/blob/contributing/.github/CONTRIBUTING.md)


benjycui:
> Can we add this document to website?

A link is enough, jump to GitHub directly.

PR #8726, opened 2017-12-25
docs: import antd style for practical-project

zhaohuxing:
According to practical-project document operation is completed,
did not show the expected results.

Because antd style did not load in the project.


yesmeck:
style is imported by `babel-plugin-import` automatically.

PR #8780, opened 2017-12-29
Add algolia search

yesmeck:
Close #8780 

Config https://github.com/algolia/docsearch-configs/blob/master/configs/ant_design.json


yesmeck:
放到 new-design-guide 分支了。

Issue #8799, opened 2018-01-03
List组件中每次grid属性中设置gutter后都会在水平方向出现滚动条，希望能够配置不出现滚动条


hitobear:
### Reproduction
The reproduction' link is https://codesandbox.io/s/o5q1z4wv89

![image](https://user-images.githubusercontent.com/14918822/34507896-7cbca2e8-f075-11e7-89cf-49da1a529c28.png)

### What problem does this feature solve?
使用List列表时关于滚动条的界面美观

### What does the proposed API look like?
如添加scroll属性，可以为false或true

PR #8830, opened 2018-01-04
Fixed list grid has scrollbar in ie #8799

nikogu:
https://github.com/ant-design/ant-design/issues/8799

### DIFF
--- a/components/list/style/index.less
+++ b/components/list/style/index.less
@@ -9,6 +9,9 @@
   * {
     outline: none;
   }
+  .ant-spin-container {
+    overflow: visible;
+  }
   &-pagination {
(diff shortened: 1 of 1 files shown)


components/list/style/index.less
```diff
@@ -9,6 +9,9 @@
   * {
     outline: none;
   }
+  .ant-spin-container {
+    overflow: visible;
```
nikogu: default is `auto`, will show scroll bar in ie.

ChauVV:
How to pull refresh on list?

nikogu:
@ChauVV When dataSource updated, the List will updated too, control by your data.

PR #9054, opened 2018-01-21
Pass name prop down to the table selection input

### LINKED ISSUE #9050: Ability to pass name of the checkbox input to getCheckedProps in rowSelection in Table
Currently, if we define a function `getCheckedProps` in `rowSelection` for `Table` component, we can only define `disabled` value there. My feature request is about being able to define `name` for the checkbox too. My use-case is that I'm doing a form (wrapping the table), so I want to give specific `name` to the selection checkboxes in my table. 

The result of the function `getCheckboxProps` could have also (optional) `name` argument, not only `disabled`.

mgrdevport:
Resolves #9050 
This PR adds the name prop to the table getCheckboxProps function to pass it down to the checkbox/radio input per row.

### DIFF
--- a/components/table/SelectionBox.tsx
+++ b/components/table/SelectionBox.tsx
@@ -45,26 +45,24 @@ export default class SelectionBox extends React.Component<SelectionBoxProps, Sel
   }
 
   render() {
-    const { type, rowIndex, disabled, onChange } = this.props;
+    const { type, rowIndex, ...rest } = this.props;
     const { checked } = this.state;
 
     if (type === 'radio') {
       return (
         <Radio
-          disabled={disabled}
-          onChange={onChange}
-          value={rowIndex}
           checked={checked}
+          value={rowIndex}
+          {...rest}
+        />
+      );
+    } else {
+      return (
+        <Checkbox
+          checked={checked}
+          {...rest}
         />
       );
     }
-
-    return (
-      <Checkbox
-        checked={checked}
-        disabled={disabled}
-        onChange={onChange}
-      />
-    );
   }
 }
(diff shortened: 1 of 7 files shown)


yesmeck:
I think we can pass the whole object which getCheckboxProps returns to Checkbox.

yesmeck:
Please PR to feature-3.2 branch.

yesmeck:
Ping @mgrdevport

mgrdevport:
I had to add the jest parameter "mapCoverage" to get accurate coverage results (locally and for codecov).
Is there any reason why the parameter was not added before?
Now I can see the correct line mappings for the code coverage result on each file.

yesmeck:
> I had to add the jest config param "mapCoverage" to get accurate coverage results (locally and for codecov).

I'v added it yesterday https://github.com/ant-design/ant-design/commit/4a9277277b686768218a7caf94392c543d220e67

components/table/SelectionBox.tsx
```diff
@@ -45,26 +45,23 @@ export default class SelectionBox extends React.Component<SelectionBoxProps, Sel
...
-    const { type, rowIndex, disabled, onChange } = this.props;
+    const { type, ...rest } = this.props;
     const { checked } = this.state;
 
     if (type === 'radio') {
       return (
         <Radio
-          disabled={disabled}
-          onChange={onChange}
-          value={rowIndex}
```
yesmeck: Show we keep `value`?

.jest.js
```diff
@@ -38,6 +38,7 @@ module.exports = {
     '!components/*/locale/index.tsx',
     '!components/*/__tests__/**/type.tsx',
   ],
+  mapCoverage: true,
```
yesmeck: Can you remove it? I've added this option.

mgrdevport:
@yesmeck 
I should have a look into the master commits more often - thanks.
Value prop on the radio is back again.

yesmeck:
Thanks!

PR #9055, opened 2018-01-21
Remove span.ant-tag-text in Tag, and remove type casting in Dropdown

zheeeng:
tow commits:

1. Warp React.ReactNode with span element is not suggested. It may cause anti-specification problem: `<span><span>I'm spec breaker</span></span>`. span is not a general tags container.

    Another benefit from this change is keeping the same structure with CheckableTag.

    After inspecting the removing of the style of .ant-tag-text, seems to bring no problems. The old example employed this CSS class has gone long long time ago. See: https://github.com/ant-design/ant-design/commit/0635877a5199a521c5477d6c659030736bc7f19a

2. By codes, the children and the overlay of Dropdown must be **single** and **valid React.ReactElement**. React.Children.only takes it and report more friendly React internal built error messages.

    At present: `<Dropdown>hello<a /></Dropdown>`, causes unfriendly error messages.
---

Extra checklist:

**if** *isBugFix* **:**

**elif** *isNewFeature* **:**

### DIFF
--- a/components/tag/style/index.less
+++ b/components/tag/style/index.less
@@ -29,14 +29,6 @@
     color: @tag-default-color;
   }
 
-  &-text {
-    a:first-child:last-child {
-      display: inline-block;
-      margin: 0 -8px;
-      padding: 0 8px;
-    }
-  }
-
   .@{iconfont-css-prefix}-cross {
     .iconfont-size-under-12px(10px);
     cursor: pointer;
--- a/components/tag/style/index.less
+++ b/components/tag/style/index.less
@@ -29,6 +29,12 @@
     color: @tag-default-color;
   }
 
+  >a:first-child:last-child {
+    display: inline-block;
+    margin: 0 -8px;
+    padding: 0 8px;
+  }
+
   .@{iconfont-css-prefix}-cross {
     .iconfont-size-under-12px(10px);
     cursor: pointer;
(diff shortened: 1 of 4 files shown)


zheeeng:
Withdraw this RP. I missed the related snapshots. I will create another PR.

components/tag/style/index.less
```diff
@@ -29,14 +29,6 @@
...
   }
 
-  &-text {
-    a:first-child:last-child {
-      display: inline-block;
-      margin: 0 -8px;
-      padding: 0 8px;
-    }
-  }
-
```
nikogu: @afc163 it's seems unnecessary ?  https://ant.design/components/tag-cn/#components-tag-demo-basic  remove this style, work fine

afc163: It is useful for `<Tag><a href="">link</a></Tag>`:

image

The hack code will fill the whole area inside Tag, otherwise it could be unclickable at the edge.

zheeeng: Yeah, I inspected these styles before PR. The .ant-tag-text is necessary.
BTW, @afc163 is this RP in progress?

nikogu: @zheeeng  take a look for `<Tag><a href="">link</a></Tag>`, the `a` must be fill the Tag block.

PR #9066, opened 2018-01-22
fix(tag): this should be the true `afterclose`

gy134340:
got this problem because I want do something when the tag is close(like set some somes state in the outer component), but meet some issues

Extra checklist:

**if** *isBugFix* **:**

**elif** *isNewFeature* **:**


yesmeck:
@gy134340 Can you demonstrate your issue on https://codesandbox.io?

PR #9067, opened 2018-01-22
Removed unnecessary section

m4jing:
As the title says, there's no need for that part of codes for this demo.


yesmeck:
Thanks!

m4jing:
My pleasure.

PR #8598, opened 2017-12-13
Pass id prop to DatePicker input field

mgrdevport:
This PR passes the DatePicker prop id to the rendered input field as requested in #8305.

PR #9123, opened 2018-01-25
Add and pass ids to modal buttons

mgrdevport:
This PR enables us to define an id for the default modal buttons and makes automated browser testing much easier.

- add id property to Button / ActionButton
- add okId and cancelId properties to modal and pass them down to the buttons
- update modal documentation (en)

I could use some help with the zh translation of the documentation...

### DIFF
--- a/components/locale-provider/__tests__/index.test.js
+++ b/components/locale-provider/__tests__/index.test.js
@@ -87,7 +87,11 @@ const App = () => (
 
 describe('Locale Provider', () => {
   beforeAll(() => {
-    MockDate.set(moment('2017-09-18T03:30:07.795Z').valueOf() + (new Date().getTimezoneOffset() * 60 * 1000));
+    const mockDate = moment('2017-09-18T03:30:07.795Z');
+    let offset = new Date().getTimezoneOffset();
+    offset = mockDate.isDST() ? 2 * offset : offset;
+    const mockNow = mockDate.valueOf() + (offset * 60 * 1000);
+    MockDate.set(mockNow);
   });
 
   afterAll(() => {
--- a/components/modal/Modal.tsx
+++ b/components/modal/Modal.tsx
@@ -39,6 +39,7 @@ export interface ModalProps {
   destroyOnClose?: boolean;
   style?: React.CSSProperties;
   wrapClassName?: string;
+  wrapProps?: object;
   maskTransitionName?: string;
   transitionName?: string;
   className?: string;
@@ -151,6 +152,7 @@ export default class Modal extends React.Component<ModalProps, {}> {
       <div>
         <Button
           onClick={this.handleCancel}
+          aria-label="cancel"
         >
           {cancelText || locale.cancelText}
         </Button>
@@ -158,6 +160,7 @@ export default class Modal extends React.Component<ModalProps, {}> {
           type={okType}
           loading={confirmLoading}
           onClick={this.handleOk}
+          aria-label="ok"
         >
           {okText || locale.okText}
         </Button>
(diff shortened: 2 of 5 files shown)


yesmeck:
Same as https://github.com/ant-design/ant-design/pull/8598#discussion_r158230999

I think we should only pass id to the root wrapper. `okId` is too specific, if we introduce `okId`, should we introduce `headerId`, `closeId`, `bodyId` ?

mgrdevport:
@yesmeck 
Then we would have 4 options to target the buttons:
wrapperid -> button[0] (cancel) / wrapperid -> button[1] (ok)

or even without any wrapper id:
.ant-modal-footer  -> button[0] / .ant-modal-footer  -> button[1]

or via the default button class:
.ant-modal-footer ->:not(.ant-btn-primary) / .ant-modal-footer -> .ant-btn-primary

or the last one (by content):
.ant-modal-footer -> text === cancelText / .ant-modal-footer -> text === okText

The test would depend on the correct array position / className / content of the button.
Don't you think that could lead to unstable test scenarios?

If that's ok for you, I would like to propose to reject this PR.

mgrdevport:
@yesmeck 
Good hint, I've used the already implemented wrapProps of the dialog to pass down an id to the wrapper.

zheeeng:
Reference element through id seems anti react philosophy. If there is a real situation that we have to access element instance, I take the side that using nested css selectors or collecting children refs and mounting them on the Modal instance.

zheeeng:
IMO, adhering id on an element is the dev's choice, should not be supported Ant Design.

mgrdevport:
You're right, react is using (and asking for) unique keys when it needs some kind of reference (i.e. the rendering of list entries). Never ids.
But I don't want to use ids in the react reference context.

I'm working on a project and I love Ant Design so far.
But when it comes to automated browser testing with i.e. Selenium, it's getting a bit messy.

You can try to write a test to target the first two date pickers in the fourth form example:
```js
const pickers = document.querySelector('#components-form-demo-time-related-controls')
                        .querySelectorAll('.ant-calendar-picker');
const first = pickers[0];
const second = pickers[1];
```
And now let's say you want to add a third one between the two. Now the second one becomes the third array entry and you would have to update the index in your test.
IMO that's not maintainable test code.

Maybe you can show me a way to produce maintainable test code with the current setup?
Perhaps I've overlooked something.

zheeeng:
`.ant-calendar-picker` is unit tested by AntD. You just need to test the tasks `#components-form-demo-time-related-controls` take. I think, the position of elements, cloud also be tested as an unit of `#components-form-demo-time-related-controls`. If it is an integration test, just make sure '#components-form-demo-time-related-controls' contains the content you certainly want.

```
[...pickers].some(assert)
```

mgrdevport:
You're right again, I'm writing integration tests and I don't want to unit test AntD components :)

Open some URL -> input form values -> submit form (api call & redirect is happening) -> ensure new entry submitted by the form is displayed as part of a list.

How would you pass the desired content (in our example the dates) to the component fields, so you could use your assertion (without the array index as indicator)?
I don't have access to the component props (selenium), so I have to find the affected inputs (via css selector) and type in the date as string (or simulate a click on the right date).

zheeeng:
Don't understand much your meaning, someway somehow, if your main codes don't depend on id selector, in test scene the element index is equivalent to id. Once you change your main code, modification on testing is necessary, after all, your solution can not handle element add/remove/change id very well.

mgrdevport:
I understand for you it doesn't matter if you work with array indices or ids in your integration test and that it has no effect on the maintainability or effort.
I'm sorry that you don't understand my intentions. If I can do anything further to bring more clarity in to this, please let me know.

PR #9172, opened 2018-01-29
Vars to theme Tabs

mrgeorgegray:
- Moves existing vars for Tabs to themes/default.less
- Adds new vars to theme Tabs
______________

* [x] Changes for `master`

### DIFF
--- a/components/style/themes/default.less
+++ b/components/style/themes/default.less
@@ -384,7 +384,18 @@
 // ---
 @tabs-card-head-background: @background-color-light;
 @tabs-card-height: 40px;
+@tabs-card-active-color: @primary-color;
 @tabs-title-font-size: @font-size-base;
+@tabs-ink-bar-bg-color: @primary-color;
+@tab-bar-margin: 0 0 16px 0;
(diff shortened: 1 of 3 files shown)


components/style/themes/default.less
```diff
@@ -384,7 +384,18 @@
 // ---
 @tabs-card-head-background: @background-color-light;
 @tabs-card-height: 40px;
+@tabs-card-active-color: @primary-color;
 @tabs-title-font-size: @font-size-base;
+@tabs-ink-bar-bg-color: @primary-color;
+@tab-bar-margin: 0 0 16px 0;
```
ryanhoho: should be 4 different vars? there 4 different tab bar positions.

mrgeorgegray: `@tab-bar-margin` was just [moved from here](https://github.com/ant-design/ant-design/blob/master/components/tabs/style/index.less#L7) and remains the same. This PR just aims to enable altering the key colors of the `<Tab />` component using the theme file.

Issue #9068, opened 2018-01-22
Mentions with autosizing input


synapse:
### What problem does this feature solve?
For example if we have a one line chat input that resizes while typing   mentions (checkout the Slack chat input)

### What does the proposed API look like?
```
<Mention
    style={{ width: '100%' }}
    onChange={onChange}
    defaultValue={toContentState('@afc163')}
    suggestions={['afc163', 'benjycui', 'yiminghe', 'RaoHai', '中文', 'にほんご']}
    onSelect={onSelect}
    autosize
  />
```

PR #9307, opened 2018-02-08
fix: make Mention autosize height by default, close: #9068

benjycui:
Close: https://github.com/ant-design/ant-design/issues/9068


benjycui:
If uses want to lock height, just set it like https://ant.design/components/mention/#components-mention-demo-multilines

PR #9408, opened 2018-02-24
do somthing after slide up animation of Alert component

Hughen:
I difficultly use Alert.onClose to beautifully close it.

problem case:
```javascript
  constructor(props) {
    super(props);

    this.state = {
      show: true
    };
  }

  render() {
    return (
      <div>
        {
          this.state.show ? (
            <Alert
              closable
              message="Alert Message..."
            />
          ):null
        }
      </div>
    )
  }
}
```
In above code, the alert slide-up animation can not be working, because the parent state(ProblemView) distribute update action maybe faster than itself animation during the closing effect.

So, antd should move `onClose` to bottom of `animationEnd` or add `onAfterClosed` property to fixed it, just like this:
```javascript
handleClose = (e) => {
  e.preventDefault();
  let dom = ReactDOM.findDOMNode(this) as HTMLElement;
  dom.style.height = `${dom.offsetHeight}px`;
  dom.style.height = `${dom.offsetHeight}px`;

  this.setState({
    closing: false,
  });
  // I cancel call function of onClose from `handleClose`
};
animationEnd = () => {
  this.setState({
    closed: true,
    closing: true,
  });
  (this.props.onAfterClosed || this.props.onClose || noop)();
};
```

Q: why I can do it?
A: The `onClose` feature is too little avail to slide up a alert message, most of the time I just want to smooth close the alert, because the alert has a slide-up animation during it's closing.

### DIFF
--- a/CHANGELOG.en-US.md
+++ b/CHANGELOG.en-US.md
@@ -21,8 +21,8 @@ If you want to read change logs before `2.0.0`, please visit [GitHub](https://gi
 
 `2018-02-26`
 
-- 🐞 Fix issue resulting in SubMenu can not been shown correctly when `defaultOpenKeys` includes nonexistent key. [#8475](https://github.com/ant-design/ant-design/issues/8475)
-- 🐞 Fix issue resulting in DatePicker's value can not been controlled correctly. [#8885](https://github.com/ant-design/ant-design/issues/8885)
+- 🐞 Fix issue resulting in SubMenu can not be shown correctly when `defaultOpenKeys` includes a nonexistent key. [#8475](https://github.com/ant-design/ant-design/issues/8475)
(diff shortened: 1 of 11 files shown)


yesmeck:
Why do you want to unmount `<Alert />`?

Hughen:
I need to keep user actions, and shown or hiden `<Alert />` is an action.

On the side, the slide-up animation in `<Alert />` that's not how it was designed. I think what antd should expose `onAfterClosed` to `<Alert />`

Hughen:
I think I should fork this repository, and pull request again.

It's OK?

yutingzhao1991:
@Hughen yes, you should. ref https://ant.design/docs/react/contributing#Sending-a-Pull-Request and make a clear PR.

Hughen:
@yutingzhao1991 you are right, I will close this PR, and I pull a official requests later.

PR #9424, opened 2018-02-27
Updated button text color setting

jkim506:
If span color is set globally, @btn-primary-color and other variables from default.less will not be used because span is more specific. Set the colour on the span of the text instead of setting it on the button.


jkim506:
Closing due to lint issues

PR #9426, opened 2018-02-27
Setting button text colour on span

jkim506:
If there is a global span colour specified, the colour of the button will not be set even if variables such as @btn-primary-color are set by the consumer of the library.

To fix this, the colour should be set on the span directly.

### DIFF
--- a/components/button/style/mixin.less
+++ b/components/button/style/mixin.less
@@ -87,6 +87,11 @@
   color: @color;
   background-color: @background;
   border-color: @border;
+  
+  > span {
+    color: @color;
+  }
+  
   // a inside Button which only work in Chrome
   // http://stackoverflow.com/a/17253457
   > a:only-child {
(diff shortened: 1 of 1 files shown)


yutingzhao1991 (approved):

yesmeck:
I think you should avoid global style.

jkim506:
Yes that is good advice but somtimes it's out of your control. And as a consumer of a library, if I set a variable intended to set the colour of something, I'd expect it to do so to the best of its abilities. And in this case, its by being as specific as possible with css selectors.

yesmeck:
As a general-purpose component, Button should not assume it's children. What if I do want to specify the color of the span? The best way I think is overriding the global style on yourself.

```css
.ant-btn > span {
  color: currentColor;
}
```

jkim506:
That's a good point. I'll close this request.

PR #9539, opened 2018-03-06
When treeNode is disabled, its switcher is highlight and clickabled

valleykid:


yesmeck:
`rc-tree` 的版本也要升级下。

valleykid:
已升级到`1.7.11`

yesmeck:
https://github.com/ant-design/ant-design/blob/master/package.json#L77

这里的

PR #9592, opened 2018-03-09
fixed modal content overflowed

Hughen:
The modal content will be overflowed, if the content is a long english word.
Just like this:
```javascript
Modal.confirm({
  title: "Modal very long title:1:abcdefghijklmnopqrstuvwxyz:it:is:so:long:long:long:long:maybe:over:mask:view",
  content: "Modal content is very:long:long:long:long:abcdefghijklmnopqrstuvwxyz:string:filled:it:and:it:is:terrible",
  maskClosable: true,
});
```

I solved this problem by adding broken words

### DIFF
--- a/components/modal/__tests__/Modal.test.js
+++ b/components/modal/__tests__/Modal.test.js
@@ -44,4 +44,13 @@ describe('Modal', () => {
     const wrapper = mount(<ModalTester footer={null} />);
     expect(wrapper.render()).toMatchSnapshot();
   });
+
+  it('render with very long word fill in title', () => {
+    const wrapper = mount(
+      <ModalTester
+        title="Modal very long title:1:abcdefghijklmnopqrstuvwxyz:it:is:so:long:maybe:over:mask:view"
+      />
+    );
+    expect(wrapper.render()).toMatchSnapshot();
+  });
 });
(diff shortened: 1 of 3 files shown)


components/modal/__tests__/Modal.test.js
```diff
@@ -44,4 +44,13 @@ describe('Modal', () => {
...
   });
+
+  it('render with very long word fill in title', () => {
+    const wrapper = mount(
+      <ModalTester
+        title="Modal very long title:1:abcdefghijklmnopqrstuvwxyz:it:is:so:long:maybe:over:mask:view"
+      />
+    );
+    expect(wrapper.render()).toMatchSnapshot();
+  });
```
afc163: This test case is not useful to test work break situation.

Hughen: I dont know what I should do for testing it, and I have offered a problem demo, ex:
[Edit n3pk656j5p](https://codesandbox.io/s/n3pk656j5p)

yesmeck: We don't have tests for CSS right now.

yesmeck:
Merged in 53abf888cbe5941e2fe14aeda5d51395dea4bc0b

PR #9528, opened 2018-03-05
Fix moment require

yesmeck:
Fix #9502

PR #9594, opened 2018-03-09
Fix moment import (default instead of wildcard)

### LINKED ISSUE #8876: Uncaught TypeError: moment.locale is not a function
### Version
3.1.1

### Environment
Windows 10, Chrome 63

### Reproduction link
[https://github.com/ant-design/parcel-antd](https://github.com/ant-design/parcel-antd)

### Steps to reproduce
1. Change index.js of parcel sample to:
```
import React from 'react';
import ReactDOM from 'react-dom';
import { Button } from 'antd';
import { LocaleProvider } from 'antd';
// The default locale is en-US, but we can change it to other language
import frFR from 'antd/lib/locale-provider/fr_FR';
import moment from 'moment';
import 'moment/locale/fr';

ReactDOM.render(
  <LocaleProvider locale={frFR}>
  <Button type="primary">Button</Button>
  </LocaleProvider>
, document.getElementById('mountNode'));
```
2. Run app

### What is expected?
Application will run successfully with locale loaded

Error thrown in console
parcel-antd.js:27172 Uncaught TypeError: moment.locale is not a function
    at Object.require.6.moment (parcel-antd.js:27172)
    at newRequire (parcel-antd.js:42)
    at localRequire (parcel-antd.js:48)
    at Object.require.4.antd/es/locale-provider (parcel-antd.js:31853)
    at newRequire (parcel-antd.js:42)
    at require.183 (parcel-antd.js:67)
    at parcel-antd.js:72

fathyb:

## What ?
This PR replaces all occurences of `import * as moment from 'moment'` to `import moment from 'moment'`. It also sets the following TypeScript options : 
- `allowSyntheticDefaultImports: true` to allow default imports even when typings don't declare them
- `esModuleInterop: true` to add ES6 interoperability code, this makes `allowSyntheticDefaultImports` not dangerous.

## Why ?
ES6 does not allow calling a namespace (in `import * as namespace from module`), because it is an object. This is problematic because Parcel uses Babel to transform ES6 modules to CommonJS, which strictly follow this standard. For more context see [moment#4488](https://github.com/moment/moment/issues/4488) and [parcel#530](https://github.com/parcel-bundler/parcel/pull/530).

It fixes #8876 and should make `antd` work on any bundler.

Extra checklist:

    - I don't know if we can add a unit test for this as it depend on the bundler. I manually tested it on Parcel and Webpack

### DIFF
--- a/components/calendar/Header.tsx
+++ b/components/calendar/Header.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import * as moment from 'moment';
+import moment from 'moment';
 import { PREFIX_CLS } from './Constants';
 import Select from '../select';
 import { Group, Button, RadioChangeEvent } from '../radio';
--- a/components/calendar/index.tsx
+++ b/components/calendar/index.tsx
@@ -1,6 +1,6 @@
 import * as React from 'react';
 import PropTypes from 'prop-types';
-import * as moment from 'moment';
+import moment from 'moment';
 import FullCalendar from 'rc-calendar/lib/FullCalendar';
 import LocaleReceiver from '../locale-provider/LocaleReceiver';
 import { PREFIX_CLS } from './Constants';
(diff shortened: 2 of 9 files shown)


yesmeck:
Fixed in #9528

PR #9602, opened 2018-03-09
Fix <Switch /> small margin

bryanberger:
This should be 13px. Size of the circle (12px) + spacing to match the left size(1px).
https://github.com/ant-design/ant-design/blob/master/components/switch/style/index.less#L110

**Issue using current value of -12.5px:**
screen shot 2018-03-09 at 2 26 22 pm

**Issue solved:**
screen shot 2018-03-09 at 2 26 17 pm

* [x] branch master

### DIFF
--- a/components/switch/style/index.less
+++ b/components/switch/style/index.less
@@ -107,7 +107,7 @@
     &:before,
     &:after {
       left: 100%;
-      margin-left: -12.5px;
+      margin-left: -(@switch-sm-height - 3px);
     }
 
     .@{switch-prefix-cls}-inner {
--- a/components/switch/style/index.less
+++ b/components/switch/style/index.less
@@ -107,7 +107,7 @@
     &:before,
     &:after {
       left: 100%;
-      margin-left: -(@switch-sm-height - 3px);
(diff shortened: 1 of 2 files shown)


bryanberger:
Hi, any thoughts on this?

yutingzhao1991 (approved):

components/switch/style/index.less
```diff
@@ -107,7 +107,7 @@
     &:before,
     &:after {
       left: 100%;
-      margin-left: -12.5px;
+      margin-left: -(@switch-sm-height - 3px);
```
valleykid: Sorry to have a little doubt: Why not define a new variable but use the `switch-sm-height`(Inconsistent with single responsibility)?
ps: `.5px` may be related to browser compatibility. (cc @afc163 )

bryanberger: Ah, sure thing. I noticed currently that variable was being used with multiple responsibility.

bryanberger:
@yutingzhao1991 @valleykid I added a new var for single responsibility. Not sure what CodeCov is failing for, thoughts?

yesmeck:
Thanks @bryanberger

PR #9636, opened 2018-03-12
MenuItem: avoid rendering Tooltip unless necessary

whtsky:

Extra checklist:

**if** *isBugFix* **:**

**elif** *isNewFeature* **:**

### DIFF
--- a/components/menu/MenuItem.tsx
+++ b/components/menu/MenuItem.tsx
@@ -21,7 +21,7 @@ class MenuItem extends React.Component<any, any> {
     const item = <Item {...props} ref={this.saveMenuItem} />;
     if (inlineCollapsed && props.level === 1) {
       return <Tooltip
-        title={inlineCollapsed && props.level === 1 ? props.children : ''}
+        title={props.children}
         placement="right"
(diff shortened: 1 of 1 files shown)


yesmeck:
@whtsky What issue does this PR resolve? Could you add a test case?

whtsky:
Right now `MenuItem` is always wrapped in `Tooltip`, which is only useful when user set `inlineCollapsed=true` and MenuItem's level is `1`.  Since `Tooltip` does not implement `shouldComponentUpdate`, this results in a lot of unnecessary React Node & updates. ( And a lot of complaining message when using [why-did-you-update](https://github.com/maicki/why-did-you-update) )
IMO we can have slight performance improvement ( and better DX w/ why-did-you-update ) by not wrapping `MenuItem` in `Tooltip` when possible.

Since it's not a bug fix nor new feature, I don't know how to write a precise test case. any ideas?

yesmeck:
Thanks!

yesmeck:
@whtsky I have to revert this change due to it breaks the menu transition animation. After adding this conditional rendering, React will create new DOM node when `inlineCollapsed` changges.

PR #9704, opened 2018-03-16
Fix table scroll x

yesmeck:
See https://codesandbox.io/s/9lx4jj4n7r

`scroll: { x: true }` does not work when set width of `.ant-table-scroll table` to `auto`.


yesmeck:
https://github.com/ant-design/ant-design/issues/2203

PR #9824, opened 2018-03-27
Release 2.13.14

ojab:
Fix for build with less-3 is already in 2.x-stable for two weeks, but not in any release yet.


yesmeck:
Thanks, I will make a release this weekend.

ojab:
As far as I can see it wasn't published to npm, can it also be done?

PR #9853, opened 2018-03-28
Fixes LocaleProvider `moment` usage

klujanrosas:
In #9528 LocaleProvider was left untouched, and it also needs to make use of the `interopDefault` when invoking `moment`.

Otherwise an error of `moment.locale is not a function` will be thrown.


klujanrosas:
@yesmeck Let me know if changes are needed.

yesmeck:
Thanks!

Issue #9923, opened 2018-04-03
Please allow Collapse settings to be overridden through the default.less settings


davidhatten:
- [X] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
I want to override the padding, margins, and other styles of various elements of the Collapse component. Doing this via css classes is rather cumbersome, as I have to specify the entire chain of classes to override styles. For example, if I want to change the style on a Collapse Panel header, I have to use this selector:

`.ant-collapse .ant-collapse-item .ant-collapse-header`

I think the right way to do this is like other component, and provide settings in the `default.less` file to override them in a local theme.

### What does the proposed API look like?
After looking at how the other component settings work, something like:
 
```
@collapse-header-padding
@collapse-item-padding
@collapse-item-background-color
```

For me, I only want padding and selected color options, but I'm not sure what other users may find useful.

ddcat1115:
You're welcome to raise a PR to add some less variables~

davidhatten:
I have raised a PR and I believe my changes are valid, but I'm catching some kind of cascading lint error across a bunch of files I haven't touched. I don't know, maybe the error makes more sense to you, but I can't make heads or tails of it.

davidhatten:
This has been fixed by PR #9943. I am closing the issue. Thank you.

PR #9924, opened 2018-04-03
push collapse styling into overridable defaults

davidhatten:

[Related Issue](https://github.com/ant-design/ant-design/issues/9923)

This pushes color and padding for the Collapse component into defaults so they can easily be overridden.

I set up the values such that the Collapse will not change appearances for anyone currently using the defaults (or any values those defaults were using), but allows a user to remove all of the padding from both the header and the content of a Collapse.

Properties added to `default.less`:

```
@collapse-header-padding          
@collapse-header-background-color 
@collapse-content-padding         
@collapse-content-background-color
@collapse-content-box-padding     
```

I couldn't see any tests around default styles for any of the other components, but if these changes are testable I'll be happy to go back and add them. 

I don't think this requires any updates to documentation.

Extra checklist:

  ~~* [ ] Update API docs for the component.~~
  ~~* [ ] Update/Add demo to demonstrate new feature.~~
  ~~* [ ] Update TypeScript definition for the component.~~
  ~~* [ ] Add unit tests for the feature.~~

### DIFF
--- a/components/collapse/style/index.less
+++ b/components/collapse/style/index.less
@@ -3,6 +3,9 @@
 
 @collapse-prefix-cls: ~"@{ant-prefix}-collapse";
 
+@collapse-header-bg: @background-color-light;
+@collapse-active-bg: @background-color-base;
+
 .collapse-close() {
   transform: rotate(0);
 }
@@ -12,6 +15,7 @@
 
 .@{collapse-prefix-cls} {
   .reset-component;
+  background-color: @collapse-header-bg;
   border-radius: @border-radius-base;
   border: @border-width-base @border-style-base @border-color-base;
   border-bottom: 0;
--- a/components/collapse/style/index.less
+++ b/components/collapse/style/index.less
@@ -32,8 +32,7 @@
 
     > .@{collapse-prefix-cls}-header {
       line-height: 22px;
-      padding: @collapse-header-padding;
-      background-color: @collapse-header-background-color;
+      padding: 12px 0 12px 40px;
       color: @heading-color;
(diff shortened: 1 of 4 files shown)


davidhatten:
I am unable to replicate the build errors locally. They refer to TimelineItem and UploadList, two components I did not touch in this change. As such, I'm not sure how to approach the build failures.

davidhatten:
Okay, so the error on TimelineItem was due to the difference between `null` and `undefined` (no idea why it wasn't failing before), and that seems to be fixed.

The error on UploadList however, is a little more complicated. I think it's angry at line 144:

`pointerEvents: 'none',`

I'm not familiar with TypeScript, but I *think* TypeScript wants `pointerEvents: "none",`, but eslint wants single quotes for that value, so that's a conflict.

Again, I don't know why these weren't failing before, or what part of my change triggered these errors, but I seem to have fixed one of them and hit a wall on the second.

davidhatten:
Okay well now it's complaining about newlines in strings that don't exist, I might have taken this as far as I can.

davidhatten:
Running `npn run lint` locally passes with no errors, and the lines that travis is saying "unexpected newline in string" look like this 

```
return '#' + this.toHex(allow3Char);
```

So, I guess I need help to figure this out.

davidhatten:
So.

This pull request has changes from the branch `origin/feature-3.4.0`. Yet, it fails. It fails with the first issues that I was unable to reproduce locally.

Can someone please take a look at this? I have no file changes but the build is still failing.

PR #10023, opened 2018-04-12
Feature  onTag api on Tag

zheeeng:

Extra checklist:

**if** *isBugFix* **:**

**elif** *isNewFeature* **:**

In using Typescript, Tag Component doest receive DOM attributes but these attributes indeed be passed down as properties to the inner div element. onTag do the same thing as onRow does in Table Component. If this PR is accepted, plz consider remove spreading div props in inner div in next break change version.

### DIFF
--- a/components/tag/CheckableTag.tsx
+++ b/components/tag/CheckableTag.tsx
@@ -6,6 +6,7 @@ export interface CheckableTagProps {
   className?: string;
   checked: boolean;
   onChange?: (checked: boolean) => void;
+  onTag?: React.DOMAttributes<HTMLDivElement>;
 }
 
 export default class CheckableTag extends React.Component<CheckableTagProps> {
@@ -16,13 +17,12 @@ export default class CheckableTag extends React.Component<CheckableTagProps> {
     }
   }
   render() {
-    const { prefixCls = 'ant-tag', className, checked, ...restProps } = this.props;
+    const { prefixCls = 'ant-tag', className, checked, onTag, onChange: _onChange, ...divProps } = this.props;
     const cls = classNames(prefixCls, {
       [`${prefixCls}-checkable`]: true,
       [`${prefixCls}-checkable-checked`]: checked,
     }, className);
 
-    delete (restProps as any).onChange; // TypeScript cannot check delete now.
-    return <div {...restProps as any} className={cls} onClick={this.handleClick} />;
+    return <div {...divProps} {...onTag} className={cls} onClick={this.handleClick} />;
   }
 }
--- a/components/tag/demo/control.md
+++ b/components/tag/demo/control.md
@@ -82,7 +82,7 @@ class EditableTagGroup extends React.Component {
         )}
         {!inputVisible && (
           <Tag
-            onClick={this.showInput}
+            onTag={{ onClick: this.showInput }}
             style={{ background: '#fff', borderStyle: 'dashed' }}
           >
             <Icon type="plus" /> New Tag
--- a/components/tag/index.en-US.md
+++ b/components/tag/index.en-US.md
@@ -30,3 +30,4 @@ Tag for categorizing or markup.
 | -------- | ----------- | ---- | ------- |
 | checked | Checked status of Tag | boolean | `false` |
 | onChange | Callback executed when Tag is checked/unchecked | (checked) => void | - |
+| onTag | Same as onTag with Tag. Pass DOM attributes to Tag element | DOM attributes | {} |
--- a/components/tag/index.en-US.md
+++ b/components/tag/index.en-US.md
@@ -22,6 +22,7 @@ Tag for categorizing or markup.
 | closable | Whether Tag can be closed | boolean | `false` |
 | color | Color of the Tag | string | - |
 | onClose | Callback executed when tag is closed | (e) => void | - |
+| onTag | Pass DOM attributes to Tag element | DOM attributes | {} |
 
 ### Tag.CheckableTag
 
--- a/components/tag/index.tsx
+++ b/components/tag/index.tsx
@@ -85,7 +85,7 @@ export default class Tag extends React.Component<TagProps, TagState> {
     const {
       prefixCls, closable, color, className, children, style,
       onTag = {}, onClose: _onClose, afterClose: _afterClose,
-      ...divProps
+      ...divProps,
     } = this.props;
     const closeIcon = closable ? <Icon type="cross" onClick={this.close} /> : '';
     const isPresetColor = this.isPresetColor(color);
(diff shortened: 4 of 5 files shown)


yesmeck:
`export interface TagProps extends React.HTMLAttributes<HTMLDivElement>`

zheeeng:
@yesmeck Have thought of that, It is likely to mix up the DOM attributes with TagProps, make the editor/IDE intelligence hints a lot of trivial available props.
Another problem is that If we hoist the DOM attributes to the Component Props' root, may cause some naming conflicts between DOM Attributes and Component. This is why we have to drop props such as `onChange`, `onClick`, `value`.

yesmeck:
See https://github.com/ant-design/ant-design/commit/82b721a432f8814ed1ae35362bb7cdb6d378dedc

We can omit `onChange`、`onClick`

`TagProps extends Omit<React.HTMLAttributes<HTMLDivElement>, 'onChange' | 'onClick'>`

yesmeck:
I don't fully understand how does `onTag` solve the problem, you still can not pass your own `onClick` to `CheckableTag` after introducing `onTag`.

zheeeng:
onTag prop usage:
```
const t = <Tag onTag={{ onClick (e) { console.log(e.altKey) }}} />
```

Advantage 1: Not pollute TagProps.
if use `TagProps extends TagProps extends React.DOMAttributes<HTMLDivElement>`:
image
There are too many useless props(defined generic React.DOMAttributes) to prevent us find Tag props.

Advantage 2: CheckableTag's `onChange` passes us `true` or `false`, its `onTag.onChange` passes use the synthetic event. It is for designing/extend some complicated features.

yesmeck:
I don't think typing system should affect the API design.

> its onTag.onChange passes us the synthetic event.

`div` don't have `onChange` event right?

zheeeng:
I second your point. But in using TS, if we don't define `onClick`, `onDoubleClick`, 'role' such things, we will never get TS passing if using these props, but javascript could handle it. If use interface extending, we can't precisely be hinted with the Tag own props. Sometimes, Antd hicked `onChange`, `onSubmit`,  in these situations, we indeed cant access synthetic event by `onChange`(as I mentioned with `CheckableTag.onChange` example).

zheeeng:
The main problem is that if we want to add event handlers to like 'click', 'doubleClick', 'touch', 'mouseover', in writing in TS, we have on efficient ways.

The current workaround is wrap Tag with a div element and appending event handler on it:
```
<div onClick={this.handleClickTag} onDoubleClick={this.handleDoubleClickTag} onMouseOver={this.handleMouseOverTag}>
    <Tag onDoubleClick={???} />  // In Typescript, it is not allowed.
</div>
```

It brings reduent div element, moreover, it is not a generic solution. If the Component props are spreading on inner deep elements.

One day, if we want to spread props to part elements of a component, 'onXXX' api designing could be used as
```
const Card = ({ onHeader, onFooter, onBody }) => 
    <div>
        <div {...onHeader} />
        <div {...onBody} />
        <div {...onFooter} />
    </div>
```
devlopers can expend more awesome features.

yesmeck:
1. Can not pass other HTML attributes `Tag` in TypeScript, because of `TagProps` didn't define them.

We can make `TagProps` extends `React.DOMAttributes<HTMLDivElement>`.

2. Extending `React.DOMAttributes<HTMLDivElement>` will introduce many insignificant props to intelligence hints, so we should introduce another prop `onTag` to pass HTML attributes.

I don't think typing system should affect the API design.

3. Can not access `event` in `onChange`.

We can pass `event` as a second parameter to `onChange`.


Hope I clarified my option.

zheeeng:
@yesmeck Yes, your opinion is clarified.

If we use `TagProps` extends `React.DOMAttributes<HTMLDivElement>`, how can we take the advantage of TypeScript intelligence to find the Tag' own props, such as `onClose`, 'afterClose', 'closable' efficiently? Do we have to look up the usage doc every time? The potential problem I have screenshotted above.

If there are new demands that access DOM events on Card/Modal/List/Table's header, content, footer, how will we add these new APIs? I admit such demands is over-designed, but it indeed can re-leverage the onXXX API designing).

afc163:
Is it any way we can highlight or stick up Tag own props without adding new props?

zheeeng:
I afraid there is no an elegant more/typing friendly way to achieve this goal.

IMO, 'onXXX'  APIs run on collecting -> spreading procedure, it is easy to managing/expanding/maintaining /implementing monkey patches.

And more, in the old API design way, we already have some potential problems.
For instance, `InputNumber` receives prop `onChange: function(value: number | string)`. What if the official API `onChange` is added after some developers had already used  the `InputNumber` like:
```
<InputNumber onChange={e => console.log(e.target.value) }
```
New API may brings back compatibility issues, right?

The more users passing DOM attributes in the old way, the fewer chance we have that adding new high-level APIs like `EmailInput::onChange as function(value: EmailStructureTuple)`, `Greeting::onClick as function(value: string)`, `RandomNumGenertor::onClick as function(value: number)`

yesmeck:
> <InputNumber onChange={e => console.log(e.target.value) } />

Can you find any exists API may encounter this situation?

zheeeng:
@yesmeck For instance, look at `<TextArea onChange={e => console.log(e.target.value) } />`, for prevent backward compatibility issues, we can't update its signature to `onChange: (value: string) => void`.

No matter whether my solution is will be accepted, I think we should keep a consistent API designing style and siganture application. 

Let's compare these inconsistencies:
1. DOM attributes spreading vs Dom attributes collecting to spreading: 
```
// CheckableTag.tsx
const { ...restProps } = this.props
<div {...restProps as any} className={cls} onClick={this.handleClick} />
```
vs
```
// rc-table/TableRow.js
const { onRow } = this.props
const rowProps = this.props.onRow(record, index)         // <-- We have used such onXXX api
<BodyRow  {...rowProps} /> // 
``` 

I perfer the latter one for its typing-friendly, and expected such changes keep going on.
Should note that according to the Antd Changelog, the synetic API `onRow` comes from 'onRowClick' -- the singualr responsiblity API 

2. props extends vs naked props definition 
```
// col.tsx
interface ColProps extends React.HTMLAttributes<HTMLDivElement> {}
```
vs
```
interface interface IconProps {}
```

Col Componet is nice for typing, but we encounterd the overwhelming props hints:
image
We lost the advantage that typing is self-documented.

The IconProps limited we setting DOM attributes in TS, need more enhances.

zheeeng:
@KagamiChan 

    after typing something on the list will be narrowed

Actually it is a beatiful assumption, we don't know how to narrow hints if we don't rember these Component APIs.

PR #10070, opened 2018-04-16
Allow to use date format provided by locale

Trufinol:
Currently there is no option to use date format provided by locale. Since the locale is always available from `this.getDefaultLocale` we can use it as a default value for the `format` prop and omit of setting the default 'YYY-MM-DD' value.

Extra checklist:

**if** *isBugFix* **:**

**elif** *isNewFeature* **:**


yesmeck:
Thanks for your contribution, but I think it's a breaking change, the default date format will change from `'YYYY-MM-DD` to `'M/D/YYYY'`.

Trufinol:
If we can add a flag like `useLocale={true}` in order to use the locale value, and if it's not set - use the default value. Would that be a solution?

yesmeck:
I'd prefer to keep it as it is and change it in next major version. Thanks for your contribution!

PR #10111, opened 2018-04-17
fix bug in getting-started - date is null when clear the DatePick

opennext:
DatePicker在界面中会显示清除小叉号。当点击小叉号清除时会报错：Uncaught TypeError: Cannot read property 'toString' of null


yesmeck:
Thanks.

PR #10158, opened 2018-04-19
Add getPopupContainer method decl to MenuProps interface

abdurahmanus:
see ant-design#10145

### DIFF
--- a/components/menu/index.tsx
+++ b/components/menu/index.tsx
@@ -51,7 +51,7 @@ export interface MenuProps {
   inlineCollapsed?: boolean;
   subMenuCloseDelay?: number;
   subMenuOpenDelay?: number;
-  getPopupContainer: (triggerNode: Element) => HTMLElement;
+  getPopupContainer?: (triggerNode: Element) => HTMLElement;
 }
 
 export interface MenuState {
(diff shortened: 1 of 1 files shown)


components/menu/index.tsx
```diff
@@ -51,6 +51,7 @@ export interface MenuProps {
   inlineCollapsed?: boolean;
   subMenuCloseDelay?: number;
   subMenuOpenDelay?: number;
+  getPopupContainer: (node: Node) => Node | null
```
yesmeck: Plz refer https://github.com/ant-design/ant-design/blob/master/components/select/index.tsx#L53

components/menu/index.tsx
```diff
@@ -51,6 +51,7 @@ export interface MenuProps {
   inlineCollapsed?: boolean;
   subMenuCloseDelay?: number;
   subMenuOpenDelay?: number;
+  getPopupContainer: (triggerNode: Element) => HTMLElement;
```
yesmeck: Need a trailing `?`

abdurahmanus: Of course, thanks!

PR #10250, opened 2018-04-25
Fix typo

marcelabomfim:


zhenghangjie:
**Customize** is a mid-20th-century American creation, so the z spelling is the original. The word is so new that some dictionaries, including the main British ones, have yet to recognize **customise** etc., but that spelling is now very common in all sorts of texts from outside North America. In recent British news stories, for instance, **customise** appears about 20 times for every customize. In Canada and the U.S., meanwhile, customize appears about 500 times for every **customise**.

zhenghangjie:
@afc163

PR #10266, opened 2018-04-26
Fix miss `getFieldProps ` on Form.

WingGao:
Fix miss `getFieldProps ` on Form.


yesmeck:
`getFieldProps` has been deprecated.

PR #10426, opened 2018-05-08
Omit 'ref' from ButtonProps otherwise typing conflicts occur.

### LINKED ISSUE #10405: Button.d.ts seems incorrect. Error: "Types of property 'ref' are incompatible."

### Version
3.5.0

### Environment
OSX, Chrome

### Reproduction link
https://github.com/cdeutsch/antd-issue/
See reproduction below as well.

### Steps to reproduce
See the code and error in `MyButton.tsx` in reproduction.

### What is expected?
You should be able to spread all ButtonProps to a Button component.

ButtonProps contains incorrect properties.

---
The bug was introduced in #10229  /cc @whtsky

cdeutsch:
Fixes #10405

Extra checklist:

**if** *isBugFix* **:**

Can't add test because tests are in Javascript, and test would require Typescript.

### DIFF
--- a/components/button/button.tsx
+++ b/components/button/button.tsx
@@ -39,7 +39,7 @@ export type ButtonType = 'default' | 'primary' | 'ghost' | 'dashed' | 'danger';
 export type ButtonShape = 'circle' | 'circle-outline';
 export type ButtonSize = 'small' | 'default' | 'large';
 
-export interface BaseButtonProps<T> extends Omit<React.HTMLProps<T>, 'size'> {
+export interface BaseButtonProps<T> extends Omit<React.HTMLProps<T>, 'ref' | 'size'> {
   type?: ButtonType;
   htmlType?: string;
   icon?: string;


cdeutsch:
An alternative to this fix would be to revert #10229, and accept anything instead.

Right now you can't do something like:
```jsx
<Button data-test="mycomponent">Hello</Button>
```

```typescript
export interface ButtonProps {
    type?: ButtonType;
    htmlType?: string;
    icon?: string;
    shape?: ButtonShape;
    size?: ButtonSize;
    loading?: boolean | {
        delay?: number;
    };
    prefixCls?: string;
    className?: string;
    ghost?: boolean;

    // accept any:
    [elemName: string]: any;
}

```

components/button/button.tsx
```diff
@@ -39,7 +39,7 @@ export type ButtonType = 'default' | 'primary' | 'ghost' | 'dashed' | 'danger';
 export type ButtonShape = 'circle' | 'circle-outline';
 export type ButtonSize = 'small' | 'default' | 'large';
 
-export interface BaseButtonProps<T> extends Omit<React.HTMLProps<T>, 'size'> {
+export interface BaseButtonProps<T> extends Omit<Omit<React.HTMLProps<T>, 'ref'>, 'size'> {
```
yesmeck: `Omit<React.HTMLProps<T>, 'size' | 'ref'>`

whtsky:
Ant Design's button currently does not forward native dom ref to the parent component ( https://reactjs.org/docs/refs-and-the-dom.html#exposing-dom-refs-to-parent-components ), so while this is a working fix, you'll still get the ref of antd's button component instead of actual dom.

cdeutsch:
@whtsky it's not about forwarding the ref, it's about getting a ref to the React component (`<Button>`) I'm declaring. That's why the Ref's don't match.

components/button/button.tsx
cdeutsch: Thanks @yesmeck. Fixed 👍

PR #10467, opened 2018-05-11
[On Development] Promise-like message interface

zhujinxuan:
## Feature:
Related to https://github.com/ant-design/ant-design/issues/10421

This PR provides the then(onClose) promise-like interface for the message, I have to change the hide function interface between I cannot work around the typescript type system.  I tried the following work around, 
```
interface MessageFunction {
  (): void;
}
interface MessageType extends MessageFunction {
  then: (fill: any, reject: any) => Promise<any>;
}
....
const result = () => {
....
} 
result.hide = (fill, reject) => closePromise.then(fill, reject);
return result
```
But it throws an error that 
```
components/message/index.tsx(88,3): error TS2322: Type '() => void' is not assignable to type 'MessageType'.
  Property 'then' is missing in type '() => void'.
```

## Check Lists

* [o] Make sure that you propose PR to right branch: bugfix for `master`, feature for latest active branch `feature-x.x`.
* [?] Make sure that you follow antd's [code convention](https://github.com/ant-design/ant-design/wiki/Code-convention-for-antd).
* [o] Run `npm run lint` and fix those errors before submitting in order to keep consistent code style.
* [o] Rebase before creating a PR to keep commit history clear.
* [o] Add some descriptions and refer relative issues for you PR.

Extra checklist:

**elif** *isNewFeature* **:**

  * [?] Update API docs for the component.
  * [o] Update/Add demo to demonstrate new feature.
  * [?] Update TypeScript definition for the component.
  * [o] Add unit tests for the feature.


zhujinxuan:
My mistake; I mistakenly introduced master. I will reopen another PR.

PR #10643, opened 2018-05-24
Fix issue #10635 Parent element set text-align break Spin

### LINKED ISSUE #10635: Parent element set text-align break Spin
### Version
3.5.3

### Environment
WIN10 - Chome66 - React 16

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/v34mq4mz95)

### Steps to reproduce
Parent element set text-align: center, and use Spin

### What is expected?
Spin at the center of parent

Spin at the right of parent

tfwww:
Fix [#10635](https://github.com/ant-design/ant-design/issues/10635)

Extra checklist:

*isBugFix* **:**

  * [ ] Make sure that you add at least one unit test for the bug which you had fixed..

No need to add the unit test since it's a CSS style issue.

### DIFF
--- a/components/spin/style/index.less
+++ b/components/spin/style/index.less
@@ -23,12 +23,12 @@
   &-nested-loading {
     position: relative;
     > div > .@{spin-prefix-cls} {
+      display: block;
       position: absolute;
       height: 100%;
       max-height: 320px;
       width: 100%;
       z-index: 4;
-      display: block;
       .@{spin-prefix-cls}-dot {
         position: absolute;
         top: 50%;
--- a/components/spin/style/index.less
+++ b/components/spin/style/index.less
@@ -28,6 +28,7 @@
       max-height: 320px;
       width: 100%;
       z-index: 4;
+      display: block;
       .@{spin-prefix-cls}-dot {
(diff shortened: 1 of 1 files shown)


wmzhong:
@afc163 
For the failed check result, it's not relevant to my fixed issue. I only changed one line code.

Also I have explored the failed test, I'm not sure it's a bug or the test case is not good enough.
Here's the detail:
1. The test case:
``` javascript
  it('should not throw error when is not a valid id', () => {
    const Form1 = Form.create()(({ form }) => (
      <Form>
        <Form.Item label="label 1">
          {form.getFieldDecorator('member[0].name.firstname')(<input />)}
        </Form.Item>
      </Form>
    ));
    const wrapper = mount(<Form1 />);
    expect(() => {
      wrapper.find('Form label').at(0).simulate('click');
    }).not.toThrow();
    expect(wrapper.find('Form input').at(0).getDOMNode()).toBe(document.activeElement);
  });
```
2. The result:
image

3. The id property is inconsistent so this test case complains failed. 
---
So maybe you can merge this PR directly or other better advice?

PR #10684, opened 2018-05-28
[Modal] Fix breakpoint for mobile view

ApacheEx:
I think for `max-width` breakpoint, it will be more correct to use `@screen-sm-max` instead of `@screen-md`. 

_p.s. Don't forget that iPad has 768px_  😅


ApacheEx:
seems unrelated issue in CI, let's try to rerun 🤞

PR #10727, opened 2018-05-31
Fix spin delay issue if mounts with spinning=true

dreamerblue:

Extra checklist:

**if** *isBugFix* **:**

If a spin component mounts with spinning=true and delay (e.g. in case of rendering a component with data loading initially), it won't be delayed. This PR just fix for it.

### DIFF
--- a/components/spin/index.tsx
+++ b/components/spin/index.tsx
@@ -56,7 +56,7 @@ export default class Spin extends React.Component<SpinProps, SpinState> {
     return !!(this.props && this.props.children);
   }
 
-  componentWillMount() {
+  componentDidMount() {
     const { spinning, delay } = this.props;
     if (spinning && delay && !isNaN(Number(delay))) {
       this.setState({ spinning: false });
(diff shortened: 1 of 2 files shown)


yesmeck:
@dreamerblue Could you add a test case?

dreamerblue:
@yesmeck OK, I have added a test case.

components/spin/index.tsx
```diff
@@ -56,6 +56,14 @@ export default class Spin extends React.Component<SpinProps, SpinState> {
     return !!(this.props && this.props.children);
   }
 
+  componentWillMount() {
```
yesmeck: Can we use `componentDidMount`? Because `componentWillMount` will be marked as unsafe in the feature versions of react.

dreamerblue: Forgot to reply there. The method has been changed to "componentDidMount".

yesmeck:
CI failed

PR #10748, opened 2018-06-02
chore: update typescript@~2.9.1

picodoth:


picodoth:
require https://github.com/ant-design/antd-tools/pull/70

Issue #10773, opened 2018-06-05
Can't compile with TypeScript 2.8.4


whtsky:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Version
3.6.1

### Environment
TypeScript 2.8.4

### Reproduction link
[https://github.com/whtsky/antd-3.6.1-ts-2.8.4-compile-error](https://github.com/whtsky/antd-3.6.1-ts-2.8.4-compile-error)

### Steps to reproduce
yarn start

### What is expected?
success

### What is actually happening?
ERROR in [at-loader] ./node_modules/antd/lib/menu/index.d.ts:56:22
    TS1005: ';' expected.

ERROR in [at-loader] ./node_modules/antd/lib/menu/index.d.ts:56:23
    TS1003: Identifier expected.

ERROR in [at-loader] ./node_modules/antd/lib/menu/index.d.ts:56:41
    TS1005: ';' expected.

ERROR in [at-loader] ./node_modules/antd/lib/menu/index.d.ts:56:16
    TS2304: Cannot find name 'import'.

---
https://travis-ci.com/whtsky/antd-3.6.1-ts-2.8.4-compile-error/builds/75328387

yesmeck:
Weird, ts compiled `'light' as MenuTheme` to `theme: import("../layout/Sider").SiderTheme;`

whtsky:
downgrade to typescript 2.8.4 seems to fix this.

bow3nwang:
`react-scripts-ts@2.13.0` getting this error.
```
antd\lib\menu\index.d.ts
(56,22): error TS1005: ';' expected.
```

duhongjun:
same as above

whtsky:
@zombieJ can you release a patch version for this?

zombieJ:
@whtsky, we will release this weekend.
@yesmeck

PR #10818, opened 2018-06-06
adjust defaultProps def

zombieJ:
To avoid `as` in `defaultProps`.
I'm not familiar with typescript. Pls help on check this.

ref: https://github.com/ant-design/ant-design/issues/10773


zombieJ:
related: https://github.com/ant-design/ant-design/pull/10778

PR #10850, opened 2018-06-08
Bug upload

Blackgan3:
When I use the beforeUpload function of the upload component, when the function returns false, I want the picture box of the component to be red, so add the code
```js
file. Status = 'error' 
```


yesmeck:
自己在  beforeUpload 里给 file 设置个 error 即可。

PR #9214, opened 2018-02-01
add custom close icon prop to tag component

cyxn:
related issue: https://github.com/ant-design/ant-design/issues/9203

so now you can pass additional prop customCloseIcon to Tag component to render something other instead of close icon. In my case I needed Popconfirm above close icon, so with that pr it's easily possible to do.

* [x] Make sure that you propose PR to right branch: bugfix for `master`, feature for latest active branch `feature-x.x`.
* [x] Make sure that you follow antd's [code convention](https://github.com/ant-design/ant-design/wiki/Code-convention-for-antd).
* [x] Run `npm run lint` and fix those errors before submitting in order to keep consistent code style.
* [x] Rebase before creating a PR to keep commit history clear.
* [x] Add some descriptions and refer relative issues for you PR.

Extra checklist:

**if** *isBugFix* **:**

  * [ ] Make sure that you add at least one unit test for the bug which you had fixed.

**elif** *isNewFeature* **:**

  * [x] Update API docs for the component.
  * [x] Update/Add demo to demonstrate new feature.
  * [x] Update TypeScript definition for the component.
  * [ ] Add unit tests for the feature.

ps I've got some issue to render my demo locally at your website, idk how it works, I guess I need to mention new .md file somewhere in configs

PR #10890, opened 2018-06-12
Add Tag[visible]

### LINKED ISSUE #9203: customize Tag component (close icon)
So I have an items that previously have been picked by user, I render them as <Tags />. also user can close this items (closable), in my case this is delete action. This delete action is sort of dangerous action, like it's gonna do some other changes related to this items. I'd like to show popconfirm before closing(deleting) this item, with some short message. I think this is also actual for select(tag mode). The end user experience is to show that is is not a simple action, and he has to be careful with this close action.

My thoughts about api is to make some 'customCloseIcon' prop - Function(onClose):ReactNode
```javascript
<Tag
    key={item.id}
    color={'#108eed'}
    closable
    onClose={() => this.deleteDangerousItem(item.id)}
    customCloseIcon={(onClose) => (
        <Popconfirm
            title="Are you sure?"
            onConfirm={onClose}
            onCancel={() => {}}
            okText="Yes"
            cancelText="No"
        >
            <Icon type="cross" />
        </Popconfirm>
    )}
>
    <span>{item.name}</span>
</Tag>
```
as far as I know it won't break api, it's still going to be animated and stuff like that

yesmeck:
Close #9203

### DIFF
--- a/components/tag/index.tsx
+++ b/components/tag/index.tsx
@@ -3,6 +3,7 @@ import * as ReactDOM from 'react-dom';
 import Animate from 'rc-animate';
 import classNames from 'classnames';
 import omit from 'omit.js';
+import { polyfill } from 'react-lifecycles-compat';
 import Icon from '../icon';
 import CheckableTag from './CheckableTag';
 
@@ -28,7 +29,7 @@ export interface TagState {
   visible: boolean;
 }
 
-export default class Tag extends React.Component<TagProps, TagState> {
+class Tag extends React.Component<TagProps, TagState> {
   static CheckableTag = CheckableTag;
   static defaultProps = {
     prefixCls: 'ant-tag',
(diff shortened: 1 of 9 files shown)


picodoth:
I don't see how this solves #9203...

yesmeck:
@picodoth https://github.com/ant-design/ant-design/pull/9214#issuecomment-362200918

components/tag/index.tsx
```diff
@@ -33,21 +35,37 @@ export default class Tag extends React.Component<TagProps, TagState> {
     closable: false,
   };
 
-  constructor(props: TagProps) {
-    super(props);
+  static getDerivedStateFromProps(nextProps: TagProps) {
+    return  ('visible' in nextProps) ? { visible: nextProps.visible } : null;
```
picodoth: 返回null的时候是覆盖state为null 还是不修改state?

yesmeck: 不修改

picodoth (approved):
+1 after fixing ci

Issue #11164, opened 2018-07-05
table filterIcon click the first second times will call onChange,but just the first second times ,the second second ....times will not 


FightingJane:
- [ ] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Version
3.6.4

### Environment
win7 google 版本 63.0.3239.132（正式版本）

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/4rool37l3w)

### Steps to reproduce
click the filterIcon,show filterDropdown,
click  the filterIcon  again ,will alert('test')

### What is expected?
not call onChange

### What is actually happening?
call onChange

---
how to prevent from  calling onChange?

filterDropDown.js 
the judge seem to be a bit of a problem:
~~~
{
        key: 'confirmFilter',
        value: function confirmFilter() {
            if (this.state.selectedKeys !== this.props.selectedKeys) {
                this.props.confirmFilter(this.props.column, this.state.selectedKeys);
            }
        }
~~~

adybionka:
I've experienced the same bug and I fixed it and created pull request https://github.com/ant-design/ant-design/pull/11166

adybionka:
Some background: this bug was introduced with bug fix for these issues: https://github.com/ant-design/ant-design/issues/10289 and https://github.com/ant-design/ant-design/issues/10209. 

This is fix for this bug: 
```
        if ('selectedKeys' in nextProps && !shallowequal(this.props.selectedKeys, nextProps.selectedKeys)) {
            newState.selectedKeys = nextProps.selectedKeys;
        }
```

as we can see selectedKeys is updated in the state only when this.props.selectedKeys and nextProps.selectedKeys has different content. In case when both are empty objects but are not the same objects a `confirmFilter` prop function will be fired according to this condition: `if (this.state.selectedKeys !== this.props.selectedKeys)`.

FightingJane:
@adybionka    Oh, I see. 3Q

IssuehuntBot:
@0maxxam0 funded this issue with $5. [Visit this issue on Issuehunt](https://issuehunt.io/repos/34526884/issues/11164)

IssuehuntBot:
@loadbalance-sudachi-kun funded this issue with $256. [Visit this issue on Issuehunt](https://issuehunt.io/repos/34526884/issues/11164)

PR #11166, opened 2018-07-05
Bugfix/table onchange event fired everytime when close filterdropdown

adybionka:
This pull request fix this bug https://github.com/ant-design/ant-design/issues/11164

### DIFF
--- a/components/table/filterDropdown.tsx
+++ b/components/table/filterDropdown.tsx
@@ -4,6 +4,7 @@ import Menu, { SubMenu, Item as MenuItem } from 'rc-menu';
 import closest from 'dom-closest';
 import classNames from 'classnames';
 import shallowequal from 'shallowequal';
+import isEmpty from 'lodash/isEmpty';
 import Dropdown from '../dropdown';
 import Icon from '../icon';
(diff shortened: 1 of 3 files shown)


components/table/filterDropdown.tsx
```diff
@@ -108,11 +109,17 @@ export default class FilterMenu<T> extends React.Component<FilterMenuProps<T>, F
   }
 
   confirmFilter() {
-    if (this.state.selectedKeys !== this.props.selectedKeys) {
-      this.props.confirmFilter(this.props.column, this.state.selectedKeys);
+    const { selectedKeys } = this.state;
+
+    if (this.hasSelectedKeys() && !shallowequal(selectedKeys, this.props.selectedKeys)) {
```
yesmeck: Don't need `this.hasSelectedKeys()`, if `state.selectedKeys` and `props.selectedKeys` are both empty, they are shallow equal, otherwise, if one is empty and the other one is not, they are absolutely not shallow equal.

yesmeck:
Merged in 968371b22a29dbc86f4e3a6009304e1a8c8f5591

PR #11189, opened 2018-07-06
Add prop into select component

thisJJ:
Add prop


components/select/index.tsx
```diff
@@ -56,6 +57,7 @@ export interface SelectProps extends AbstractSelectProps {
   tokenSeparators?: string[];
   getInputElement?: () => React.ReactElement<any>;
   autoFocus?: boolean;
+  id?: string;
```
yesmeck: Don't need this. `AbstractSelectProps` already has `id`.

yesmeck: @thisJJ

yesmeck (changes requested):

yesmeck:
Thanks.

Issue #11404, opened 2018-07-22
Multiple row selection in table component


RaphaelChauveau:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
When working with the table component, I sometimes have to select a large number of contiguous rows. To do that, we currently have to select each row independently. This process can be very long and painstaking.

In order to facilitate the selection or unselection of multiple contiguous rows, I propose a new feature to allow multiple row selection/unselection. 

To (un)select multiple rows, click the checkbox of the first row and shift-click the checkbox of the last row. All the rows located between the first and the last rows shall be (un)selected. If the last row was selected, the rows shall be unselected and if it was unselected, the rows shall be selected. The described behavior is the same as the Google Gmail application. 

### What does the proposed API look like?
The rowSelection config should handle a new type of callback : onSelectMulti.

This callback should be called when rows are (un)selected using the multiple row (un)selection feature. The callback should have the same type as the onSelectAll callback (Function(selected, selectedRows, changeRows))

PR #11406, opened 2018-07-22
Multiple row selection

RaphaelChauveau:
This PR intends to solve issue #11404 (Multiple row selection in table component).

The Chinese documentation is missing from the PR. Could someone please add it ?


components/table/Table.tsx
```diff
@@ -124,6 +124,7 @@ export default class Table<T> extends React.Component<TableProps<T>, TableState<
       // 减少状态
       filters: this.getFiltersFromColumns(),
       pagination: this.getDefaultPagination(props),
+      pivot: void(0),
```
yesmeck: I think we can use `undefined` safely.

yesmeck (approved):

PR #11474, opened 2018-07-26
Update Pagination Typescript definition to include "role"

kagd:
A [PR on rc-pagination](https://github.com/react-component/pagination/pull/135/files) added the ability to define `data-`, `aria-` and `role` attributes. This PR is just adding the `role` attribute to the definition.

Definitions for `data-` and `aria-` are not required since they are already allowed as props on Pagination.


yesmeck:
Thanks!

PR #11567, opened 2018-08-01
Use tabular-nums font variant instead of switching to Tahoma for figures

tibdex:
Using a completely different font for figures leads to inconsistencies: https://github.com/ant-design/ant-design/issues/9422.
Most modern font systems default to [tabular figures](https://www.fonts.com/content/learning/fontology/level-3/numbers/proportional-vs-tabular-figures).
One exception to this is Apple's San Francisco which defaults to [proportional figures](http://martiancraft.com/blog/2015/10/san-francisco-part-2/#special-features-numerals).
There is, however, an official and clean way to switch to tabular figures. It's the [`font-variant` CSS property](https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-numeric) that can be set to `tabular-nums`.

This commit removes the "Monospaced Number" font altogether and use `font-variant: tabular-nums` instead where it was used.
It also moves closer to the sans-serif and monospace [system font stacks used by GitHub](http://markdotto.com/2018/02/07/github-system-fonts/).

---

Here is the before/after on Windows where Segoe UI is now used for figures too. The difference is subtle but you can see it. Especially for the figure 1.

#### Before

before

#### After

after

### DIFF
--- a/docs/spec/font.en-US.md
+++ b/docs/spec/font.en-US.md
@@ -22,16 +22,15 @@ While defining the font system for a visual system, we propose to start from the
 In order to implement a good font system, the first thing is to choose an appropriate font family. Ant Design prefers the system default font family and then also provides a set of alternative font libraries to maintain readability for screens on different platforms and browsers and to make sure it's always user friendly, stable and professional to end user.
 
 ```css
-font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
-             "Helvetica Neue", Helvetica, "PingFang SC", "Hiragino Sans GB", "Microsoft YaHei",
-             SimSun, sans-serif;
+font-family: "Chinese Quote", -apple-system, BlinkMacSystemFont, "Segoe UI", "PingFang SC", "Hiragino Sans GB", "Microsoft YaHei", "Helvetica Neue", Helvetica, Arial, sans-serif,
+ "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol";
 ```
 
-> References：https://www.smashingmagazine.com/2015/11/using-system-ui-fonts-practical-guide/
+> References：https://www.smashingmagazine.com/2015/11/using-system-ui-fonts-practical-guide/ and http://markdotto.com/2018/02/07/github-system-fonts/
 
-In addition, in a lot of applications, numbers often need to be displayed vertically. We set the numeric font to `Tahoma` specifically to make it a monospaced font.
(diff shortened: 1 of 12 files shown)


docs/spec/font.en-US.md
```diff
@@ -22,16 +22,15 @@ While defining the font system for a visual system, we propose to start from the
 In order to implement a good font system, the first thing is to choose an appropriate font family. Ant Design prefers the system default font family and then also provides a set of alternative font libraries to maintain readability for screens on different platforms and browsers and to make sure it's always user friendly, stable and professional to end user.
 
 ```css
-font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto,
-             "Helvetica Neue", Helvetica, "PingFang SC", "Hiragino Sans GB", "Microsoft YaHei",
-             SimSun, sans-serif;
+font-family: "Chinese Quote", -apple-system, BlinkMacSystemFont, "Segoe UI", "PingFang SC", "Hiragino Sans GB", "Microsoft YaHei", "Helvetica Neue", Helvetica, Arial, sans-serif,
```
tibdex: I removed `SimSun` here because [it wasn't actually used](https://github.com/ant-design/ant-design/pull/11567/files#diff-4b7c05557620faa918b0919c047d049a).

I didn't change the `font.zh-CN.md` file, but the PR is configure to "Allow edits from maintainers.". Feel free to edit the Chinese page and to tweak the line breaking if you want shorter lines in the docs.

Gyokur0: Hi @tibdex the adding of "Chinese Quote" replaced Segoe UI everywhere in IE11  :( 
Had to fix it manually



(chrome looks ok)

tibdex: Hi @Gyokur0, as you can see in my comment above, I only added `"Chinese Quote"` in the docs to keep them in sync with the actual font stack used in Less files.

This is the font stack before my commit:

https://github.com/ant-design/ant-design/blob/1d924300e7b682a6b8c7d723ad0a0a7721895f7d/components/style/themes/default.less#L37-L39

Note the presence of `"Chinese Quote"`.

And after:

https://github.com/ant-design/ant-design/blob/b6e9646a6ad5fef497c48d955cc0e6141e6146f2/components/style/themes/default.less#L37-L39

Still starts with `"Chinese Quote", -apple-system, BlinkMacSystemFont, "Segoe UI"`.

I don't see how what you described would have happened.

Besides, I just visited https://ant.design/components/table/, which ships with this commit, on IE 11 and it displays with Segoe UI as expected.

afc163:
https://github.com/ant-design/ant-design/issues/11456
https://github.com/ant-design/ant-design/issues/11843

`font-variant: tabular-nums;` will reduce render perfermance of long word in chome. I fix this by resetting this in Select in https://github.com/ant-design/ant-design/commit/1a30e5de7878eb8e956ed8a7a1cc4df0d747316d for now. Looking forword to further solution or wait for chrome fix.

PR #11705, opened 2018-08-10
fix eslint in package.json

NE-SmallTown:
#11701 left some fix, sorry


.eslintignore
```diff
@@ -1,7 +1,6 @@
 components/**/*.js
 components/**/*.jsx
 components/*/__tests__/type.tsx
-!.eslintrc.js
```
NE-SmallTown: We have set `!.*.js` below so we don't need this

Issue #11699, opened 2018-08-10
antd3.8.0 upload组件interface.d.ts,RcFile接口报错


baggiomygod:
- [ ] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Version
3.8.0

### Environment
macos

### Reproduction link
[https://github.com/ant-design/ant-design/tree/master/components/upload](https://github.com/ant-design/ant-design/tree/master/components/upload)

### Steps to reproduce
引入antd import {} from antd

### What is expected?
正常运行

### What is actually happening?
[ts]
接口“RcFile”错误扩展接口“File”。
  属性“lastModifiedDate”的类型不兼容。
    不能将类型“Date | undefined”分配给类型“Date”。
      不能将类型“undefined”分配给类型“Date”。

---
源码interface.d.ts
```
export interface RcFile extends File {
    uid: string;
    lastModifiedDate?: Date; // 原码是：lastModifiedDate?: Date;可选 File接口里原码是：lastModifiedDate: Date 必须
}
```
以上代码改成： lastModifiedDate: Date后正常
```
export interface RcFile extends File {
    uid: string;
    lastModifiedDate: Date;
}
```

windwang:
+1

andycall:
#11709

PR #11709, opened 2018-08-10
Fix typescript compile error

andycall:
fix obviously typescript@3.0.1 compile error.
image


andycall:
There must be some potential bugs in ant.design.
It's impossible to cause logic error after typescript typing changed.

image

afc163:
close #11699

PR #11767, opened 2018-08-16
Update TreeProps.onCheck function signature

yallie:
This is a small fix for the TypeScript definition.

The `onCheck` function is called with either `string[]` data for `checkedKeys` argument,
or with the `{ checked: string[], halfChecked: string[] }` object, when `checkStrictly` is enabled.

Here is the related line of code in the underlying rc-tree component:

https://github.com/react-component/tree/blob/master/src/Tree.jsx#L464

### DIFF
--- a/components/tree/Tree.tsx
+++ b/components/tree/Tree.tsx
@@ -100,7 +100,7 @@ export interface TreeProps {
   /** 展开/收起节点时触发 */
   onExpand?: (expandedKeys: string[], info: AntTreeNodeExpandedEvent) => void | PromiseLike<any>;
   /** 点击复选框触发 */
-  onCheck?: (checkedKeys: string[], e: AntTreeNodeCheckedEvent) => void;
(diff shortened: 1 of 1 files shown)


components/tree/Tree.tsx
```diff
@@ -100,7 +100,7 @@ export interface TreeProps {
   /** 展开/收起节点时触发 */
   onExpand?: (expandedKeys: string[], info: AntTreeNodeExpandedEvent) => void | PromiseLike<any>;
   /** 点击复选框触发 */
-  onCheck?: (checkedKeys: string[], e: AntTreeNodeCheckedEvent) => void;
+  onCheck?: (checkedKeys: string[] | { checked: string[]; halfChecked: string[] }, e: AntTreeNodeCheckedEvent) => void;
```
picodoth: it feels like we should make a change in rc-tree, instead of doing:
```
(checkedKeys: string[] | { checked: string[]; halfChecked: string[] }
```
We should keep it consistent to 
```
{ checked: string[]; halfChecked: string[] }
```
Just pass empty array to halfChecked if necessary.

@zombieJ

zombieJ (approved):

PR #12049, opened 2018-09-04
Added Mongolian locale provider and tested.

digz6666:
Added Mongolian locale provider and tested.
Translated calender, date-picker, time-picker components.

Also added Mongolian translation of rc-calender and rc-pagination components.
https://github.com/react-component/calendar/pull/411
https://github.com/react-component/pagination/pull/144

### DIFF
--- a/components/calendar/locale/mn_MN.tsx
+++ b/components/calendar/locale/mn_MN.tsx
@@ -1,0 +1,2 @@
+import mn_MN from '../../date-picker/locale/mn_MN';
+export default mn_MN;
--- a/components/date-picker/locale/mn_MN.tsx
+++ b/components/date-picker/locale/mn_MN.tsx
@@ -1,0 +1,19 @@
+import CalendarLocale from 'rc-calendar/lib/locale/mn_MN';
+import TimePickerLocale from '../../time-picker/locale/mn_MN';
+
+// Merge into a locale object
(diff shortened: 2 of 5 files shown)


zombieJ:
Hi,
CI failed, pls check

digz6666:
@zombieJ When I check the test output it matches with my defined values.
It says following in the log and I don't understand :)
```
  ● Locale Provider › should display the text as mn-mn
    expect(value).toMatchSnapshot()
    New snapshot was not written. The update flag must be explicitly passed to write a new snapshot.
    This is likely because this test is run in a continuous integration (CI) environment in which snapshots are not written by default.
```

zombieJ:
test case will check current code match snapshot or not. Since this locale is first added. Just run follow code to update snapshot:
```bash
npm test -- -u
```

digz6666:
@zombieJ I ran the npm test -- -u and npm test on my local clone and it passes.

Result of npm test -- -u:
local_test

Result of npm test:
local_test_2

What can I do?

digz6666:
Seems like people experiencing similar issues: https://github.com/facebook/jest/issues/5270

zombieJ:
Let me handle this : )

digz6666:
@zombieJ thank you, you rock!

PR #12057, opened 2018-09-04
Add vars for Alert types

mrgeorgegray:
Expose less vars for `<Alert />` types. 

Rather than use the hardcoded `colorPalette` mixin for [Alert Types](https://ant.design/components/alert/#components-alert-demo-style) we would like to provide our own colors in the Theme override so I have moved these into the `default.less` theme.

### DIFF
--- a/components/alert/style/index.less
+++ b/components/alert/style/index.less
@@ -31,34 +31,34 @@
   }
 
   &-success {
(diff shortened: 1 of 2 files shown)


bryanberger:
@afc163 we need this too, any chance of getting a review on this?

PR #12059, opened 2018-09-04
Fix less config in documentation

dmitrylabs:
There is a bug in documentation which produces an error:

```
Failed to compile.

./node_modules/antd/es/button/style/index.lessModule build failed:

// https://github.com/ant-design/ant-motion/issues/44
.bezierEasingMixin();^
Inline JavaScript is not enabled. Is it set in your options?
      in /.../node_modules/antd/es/style/color/bezierEasing.less (line 110, column 0)
```

Solution found here:
https://github.com/ant-design/ant-design/issues/7927#issuecomment-400368810


zombieJ:
Thanks for contribution. Code merged

Issue #12158, opened 2018-09-10
Table组件排序图标选中和悬浮的样式希望能改成不一样的


ty-bt:
- [ ] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
当我悬浮在排序箭头按钮上的时候，无法区分当前是否已开启排序

### What does the proposed API look like?
排序箭头开启和悬浮的样式不一样

---------------

Translate: Table title support customize sort icon

zombieJ:
PR is welcomed 😄

IssuehuntBot:
@yaodingyd has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/34526884/issues/12158)

IssuehuntBot:
@yaodingyd has stopped working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/34526884/issues/12158)

IssuehuntBot:
@afc163 has started working. [Visit this issue on Issuehunt](https://issuehunt.io/repos/34526884/issues/12158)

IssuehuntBot:
@afc163 has submitted [output](https://github.com/ant-design/ant-design/pull/12264). [Visit this issue on Issuehunt](https://issuehunt.io/repos/34526884/issues/12158)

IssuehuntBot:
@afc163 has rewarded. [Visit this issue on Issuehunt](https://issuehunt.io/repos/34526884/issues/12158)

PR #12246, opened 2018-09-14
Increase selected sorter size to be different from hover style  (#12158)

yaodingyd:
Minor style fix to make selected sorter style more distinctive from hover style


afc163:
We will change this style to this way soon:

image

yaodingyd:
Feel free to close if already plan for different implementation.

PR #12369, opened 2018-09-22
Get the filtered data after executing the filter function on the dataSource

zombieJ:
ref: https://github.com/ant-design/ant-design/issues/10298

add extra arg on `onChange` event in table


BurgersBacon:
> If we set a filter programmatically, onChange method won't triggered.

any solution for this one?

PR #12448, opened 2018-09-28
fix incorrect prop passed to checkbox in docs examples

marcosfede:
prop should be `checked` instead of `value`


zombieJ:
Nice catch!
CI broke, pls check.

PR #12463, opened 2018-09-28
feature: Added Hebrew translations

orenk86:


orenk86:
Okay, rc-calendar and rc-pagination libraries have been updated and everything seems okay. This PR is ready as far as I'm concerned 😊

orenk86:
Worth mentioning that the failing tests seem unrelated to this PR. They also fail locally for me, with the same errors.
When I run `npm test -- -u` the snapshots get updated and all of the tests pass locally for me.

Should I include an update for the snapshots in this PR? Seems a bit unrelated.
Perhaps it would be better to open another one...

docs/react/i18n.en-US.md
```diff
@@ -27,6 +27,7 @@ Supported languages:
 
 |Language|Filename|
 |---|---|
+|Hebrew|he_IL|
```
zombieJ: Pls order with `Filename` field.

zombieJ (changes requested):

orenk86:
@afc163 @zombieJ May I ask when this will be included in a release? Thanks!

PR #12521, opened 2018-10-05
onBlur event to Popover

vsoren:
Added the onBlur event to Popover


zombieJ:
Currently we pass all the rest props into internal trigger component but we don't hope user use such prop on the Popover. User should handle `onBlur` in the children component instead of Popover:
```jsx
<Popover content="content" title="title">
  <Button onBlur={...}>Hover me</Button>
</Popover>
```

PR #12771, opened 2018-10-22
Migrate Cascader to new lifecycle methods

natanielcz:
Part of #9792 

Maybe it's worth to consider using [memoize](https://github.com/alexreardon/memoize-one) to cache flattenOptions ([recomendation from React team](https://reactjs.org/blog/2018/06/07/you-probably-dont-need-derived-state.html#what-about-memoization)). 

I inlined flattenOptions to render method to avoid complexity caused by introducing getDerivedStateFromProps.

### DIFF
--- a/components/cascader/index.tsx
+++ b/components/cascader/index.tsx
@@ -4,6 +4,7 @@ import arrayTreeFilter from 'array-tree-filter';
 import classNames from 'classnames';
 import omit from 'omit.js';
 import KeyCode from 'rc-util/lib/KeyCode';
+import { polyfill } from 'react-lifecycles-compat';
 import Input from '../input';
 import Icon from '../icon';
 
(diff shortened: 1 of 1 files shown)


components/cascader/index.tsx
```diff
@@ -165,22 +165,20 @@ export default class Cascader extends React.Component<CascaderProps, CascaderSta
...
       popupVisible: props.popupVisible,
-      flattenOptions:
-        props.showSearch ? this.flattenTree(props.options, props.changeOnSelect, props.fieldNames) : undefined,
     };
   }
 
-  componentWillReceiveProps(nextProps: CascaderProps) {
-    if ('value' in nextProps) {
-      this.setState({ value: nextProps.value || [] });
+  componentDidUpdate(prevProps: CascaderProps) {
```
zombieJ: Do not use `componentDidUpdate` instead of `componentWillReceiveProps`. Use `getDerivedStateFromProps` instead.

natanielcz: Why componentDidUpdate isn't the right place for this code?

zombieJ: `getDerivedStateFromProps` is the replacement function of `componentWillReceiveProps`. It's better use it here if you don't have effect code.

PR #12907, opened 2018-10-31
Improve Affix

ztplz:
Just a little improve for `Affix`:
  - remove unused variable
  - more type-safety


ztplz:
@afc163  Hm... it seems not change coverage,  that's make me confused.

Issue #12743, opened 2018-10-20
Allow Tooltip above slider to permanently be visible


8shaks:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
This feature would allow the slider tooltip to permanently be visible instead of only being visible when users are dragging on the slider handle. This would be useful since it would allow devs to customize their website more. It would also allow the user to always be able to see the value of the slider.

### What does the proposed API look like?
`<Slider
defaultToolTip={true}/>`

zy410419243:
if you don't want to wait for upgrading,  you can change code in index.tsx which is in node_modules/antd/slider, around line 85 to 93, as below
``` javascript
var _this$props = _this.props,
   tooltipPrefixCls = _this$props.tooltipPrefixCls,
   tipFormatter = _this$props.tipFormatter,
   defaultToolTip = _this$props.defaultToolTip;

   var visibles = _this.state.visibles;

   var visible = tipFormatter && visibles[index] || dragging || defaultToolTip : false;

   return React.createElement(
```
then, you can try <Slider defaultToolTip /> as you suppose.

If this issue will be upgraded, I will take it.

8shaks:
So how long would it be until it is upgraded?

8shaks:
Also how I would make so if users pull my code from GitHub and do an npm install they get the customized Index.tsx that I made?

8shaks:
Thanks

PR #12915, opened 2018-11-01
Add tooltipVisible to set ToolTip shown

orzyyyy:
[#12743](https://github.com/ant-design/ant-design/issues/12743)


8shaks:
I don't see this in the new Antd changelog? Is there a reason why this is not there? How would I make the tooltip always visible now?

yutingzhao1991:
@8shaks the feture branch not released yet.

8shaks:
Oh ok, do you know when the feature branch will be released?

zkeyword:
so same

PR #13206, opened 2018-11-20
Cascader support limit filtered item count

zombieJ:
Add `limit` prop in `showSearch`. Default value: 50:
```jsx
<Cascader
  {...props}
  showSearch={{ limit: 30 }}
/>
```

ref: #13097

### DIFF
--- a/components/cascader/index.en-US.md
+++ b/components/cascader/index.en-US.md
@@ -50,7 +50,7 @@ Fields in `showSearch`:
 | Property | Description | Type | Default |
 | -------- | ----------- | ---- | ------- |
 | filter | The function will receive two arguments, inputValue and option, if the function returns true, the option will be included in the filtered set; Otherwise, it will be excluded. | `function(inputValue, path): boolean` |  |
-| limit | Set the count of filtered items | number | 50 |
+| limit | Set the count of filtered items | number \| false | 50 |
 | matchInputWidth | Whether the width of result list equals to input's | boolean |  |
 | render | Used to render filtered options. | `function(inputValue, path): ReactNode` |  |
 | sort | Used to sort filtered options. | `function(a, b, inputValue)` |  |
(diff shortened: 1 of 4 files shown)


components/cascader/index.tsx
```diff
@@ -97,6 +98,8 @@ export interface CascaderState {
   flattenOptions: CascaderOptionType[][] | undefined;
 }
 
+const defaultLimit = 50;
```
JLHwung: Please document this variable in comments as it is too far from lines where it is used.

components/cascader/index.tsx
```diff
@@ -317,10 +320,30 @@ export default class Cascader extends React.Component<CascaderProps, CascaderSta
...
+      flattenOptions.some((path) => {
+        const match = filter(this.state.inputValue, path, names);
+        if (match) {
+          filtered.push(path);
+          matchCount += 1;
+        }
+        return matchCount >= limit;
+      });
+    } else {
+      filtered = flattenOptions.filter((path) => filter(this.state.inputValue, path, names));
```
JLHwung: As `limit` is backed by a default value, a user can only specify
```
<Cascader showSearch={{ filter, limit: -1 }} />
```
to go into this branch.

I suggest just cap the `limit` to be non-negative and remove this branch.

JLHwung (changes requested):

JLHwung:
I can tell from the implementation details that this change is actually a breaking change. The following code
```
<Cascader
  {...props}
/>
```
will now have at most 50 filtered item unless they modify their code and specify that
```
<Cascader
  {...props}
  showSearch={{ limit: Infinity }}
/>
```

It is a good idea to add default limit to `Cascader`. I hope that the limit itself is configurable in a project level, so people can opt in to the new behaviour without breaking other's code.

components/cascader/index.en-US.md
```diff
@@ -50,6 +50,7 @@ Fields in `showSearch`:
 | Property | Description | Type | Default |
 | -------- | ----------- | ---- | ------- |
 | filter | The function will receive two arguments, inputValue and option, if the function returns true, the option will be included in the filtered set; Otherwise, it will be excluded. | `function(inputValue, path): boolean` |  |
+| limit | Set the count of filtered items | boolean | 50 |
```
JLHwung: Shouldn't it be of the type `number`? 🤔

afc163 (approved):

JLHwung (approved):

Issue #1668, opened 2016-05-12
Modal 关闭时样式问题


codering:
点击右上角关闭时，会出现边框

![2016-05-12 15-24-33](https://cloud.githubusercontent.com/assets/2271825/15207066/2000bf7e-1856-11e6-8b2d-97343a0091e4.png)

PR #13220, opened 2018-11-20
fix(module:alert): padding for overlapping close button

cohancarpentier:
## PR Type
What kind of change does this PR introduce?

```
[x] Code style update (formatting, local variables)
```

## What is the current behavior?

Issue Number: [#1668](https://github.com/NG-ZORRO/ng-zorro-antd/issues/1668)

## What is the new behavior?

Close button doesn't overlap alert's content anymore.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

### DIFF
--- a/components/alert/style/index.less
+++ b/components/alert/style/index.less
@@ -15,6 +15,7 @@
 
   &&-no-icon {
     padding: 8px 15px;
+    .clearfix;
   }
 
   &-icon {
@@ -63,11 +64,9 @@
 
   &-close-icon {
     font-size: @font-size-sm;
-    position: absolute;
-    right: 16px;
-    top: 8px;
-    line-height: 22px;
+    line-height: 21px;
     overflow: hidden;
+    float: right;
     cursor: pointer;
 
     .@{iconfont-css-prefix}-close {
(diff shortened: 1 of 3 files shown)


components/alert/style/index.less
```diff
@@ -111,6 +111,11 @@
     font-size: @font-size-base;
   }
 
+  &-message {
```
ilanus: could you place it on line `86` above `&-with-description` to not interrupt the `&-with-description` block?

cohancarpentier: Done!

ilanus: although the overflow still happens if the element is wider than `16px` not sure if `padding` is the correct solution here:
screen shot 2018-11-20 at 17 29 53

cohancarpentier: You're right. I made a new commit using float right instead and that seems to fix it.

ilanus: rebase and update snapshops

cohancarpentier: I'm clueless as to how to update snapshots; do I do that by hand or is there some sort of command? Thanks!

ilanus: `yarn test -u` || `npm test -- -u`

ilanus: now `&-message` is fixed, but `<Alert description="Warning Text Warning Text Warning TextW arning Text Warning Text Warning TextWarning Text" />` still have the issue 

screenshot 2018-11-21 at 21 54 19

PR #13242, opened 2018-11-22
Fixes #7927 #11097 - Convert inline JavaScript to @plugin

### LINKED ISSUE #7927: import the antd theme, webpack build show .bezierEasingMixin error ?

matthew-dean:
This makes styles compatible with Less 3.x default settings (and also makes the code cleaner and eliminates global leaking).

You're welcome to take this and organize the plugin differently. I essentially concatted the inline JS in the other .less files.

Couple notes:

1. Less already has a bunch of color manipulation functions, so I'm not sure why the `tinycolor` library is pasted in there.
2. Previously, `tinycolor` and `colorEasing` were leaked as globals, but I wasn't sure if they were actually consumed anywhere. `tinycolor` is referenced in `colorPalette`, so that sort of made sense. But `colorEasing`? 

I didn't have time to do linting or look at code conformity. This should get you 99% to using `@plugin` how you want though. Basically, in Less, you can register any functions you want, which accept nodes (not JS values - you'll notice in the code for the `colorPalette` function that the index is a node with a value property, not a number; likewise, the color is a parsed Less Color node). Then, all you have to do is return a Less node from that function.

Less tree nodes are here: https://github.com/less/less.js/tree/master/lib/less/tree


chenshuai2144:
这件事总是需要做的，我们要不做了吧。这样和社区的生态也好对接

PR #13276, opened 2018-11-25
Add Input.Password

orzyyyy:
For Issue #13267.


ilanus:
Great job! we are missing `eye-close` type of icon

hswm

Maybe use `show / hide` text instead of icon?

components/input/index.tsx
```diff
@@ -2,13 +2,16 @@ import Input from './Input';
 import Group from './Group';
 import Search from './Search';
 import TextArea from './TextArea';
+import Password from './Password';
 
 export { InputProps } from './Input';
 export { GroupProps } from './Group';
 export { SearchProps } from './Search';
 export { TextAreaProps } from './TextArea';
+export { PasswordProps } from './Password'
```
ilanus: `;`

components/input/Input.tsx
```diff
@@ -5,6 +5,7 @@ import omit from 'omit.js';
 import Group from './Group';
 import Search from './Search';
 import TextArea from './TextArea';
+import Password from './Password'
```
ilanus: `;`

zy410419243:
@ilanus Thanks for review and advice. I would have some try as below,
  
- add `Button` that the text can be customed by user.
  
- when click `Button` or `Icon`, the show of text should have a transition.

zy410419243:
#13281

PR #13278, opened 2018-11-25
Update Docs

ztplz:
ref: #13274 
Remind users that Ant Design's solution of using react-app-rewired would not work anymore since it doesn't support CRA 2.0.
@afc163


abenhamdine:
Actually react-app-rewired still work with cra 2.1 (and antdesign, typescript, babel plugin import etc..). But the author doesn't use cra anymore and didn't test it with cra 2, thus does not want to officially support it.
Otherwise a good alternative is craco, with corresponding plugins for antdesign.

PR #13285, opened 2018-11-26
Check textarea length with js

### LINKED ISSUE #12318: Textarea with maxLength

### Version
3.9.2

### Environment
Chrome 69.0.3497.81

### Reproduction link
[https://codepen.io/gonenoob/pen/vzvEvY](https://codepen.io/gonenoob/pen/vzvEvY)

### Steps to reproduce
Copy the contents in the Table and paste them into TextArea. The maxLength attribute of TextArea does not play a correct role.I limit the maximum length to 28, but the content length can be up to 29.

### What is expected?
MaxLength of TextArea works correctly.

The maxLength attribute of TextArea does not play a correct role.

ztplz:
Fix #12318
WebKit Bug: [https://bugs.webkit.org/show_bug.cgi?id=182725](https://bugs.webkit.org/show_bug.cgi?id=182725)

### DIFF
--- a/components/input/TextArea.tsx
+++ b/components/input/TextArea.tsx
@@ -92,6 +92,10 @@ export default class TextArea extends React.Component<TextAreaProps, TextAreaSta
     if (!('value' in this.props)) {
       this.resizeTextarea();
     }
+    const { maxLength } = this.props;
+    if (maxLength && e.target.value.length > maxLength) {
+      e.target.value = e.target.value.substring(0, maxLength);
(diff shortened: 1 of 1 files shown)


zombieJ:
I don't think we should help browser to handle this.
Since help to modify `value` of `target` is do modify the dom element's value directly which should be avoid in React.

ztplz:
@zombieJ Yes, users also can handle this situation too.

PR #13384, opened 2018-12-01
Fix input type slow

ztplz:
There is no need to consider the mobile browser, which may cause some problems on the desktop browser.


ztplz:
fix #13347

Issue #13372, opened 2018-11-30
【Tree】Support customize toggle collapse/expand icon


nick-ChenZe:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
In our project, we need to change this icon according to the design and we don't want reset this by override the font icon in css. I found prop `switcherIcon` in `rc-tree`  which not open in  the `Tree`  component of antd. 
So which i want to request is to reopen the property. If that is allowed, i can make the pull request.
Thanks for your work.

### What does the proposed API look like?
switcherIcon

mrcaobin:
same problem，expand icon is too small , hope to support customize .

IssuehuntBot:
@issuehuntfest has funded $20.00 to this issue. [See it on IssueHunt](https://issuehunt.io/repos/34526884/issues/13372)

IssuehuntBot:
@nick-chenze has submitted a pull request. [See it on IssueHunt](https://issuehunt.io/repos/34526884/issues/13372)

IssuehuntBot:
@afc163 has rewarded $14.00 to @nick-chenze. [See it on IssueHunt](https://issuehunt.io/repos/34526884/issues/13372)

- :moneybag: Total deposit: $20.00
- :tada: Repository reward(20%): $4.00
- :wrench: Service fee(10%): $2.00

PR #13592, opened 2018-12-12
 feat: add switcherIcon prop to Tree

nick-ChenZe:
#13372


nick-ChenZe:
seem `table-pipe-alignment` option not reasonable for docs

PR #13625, opened 2018-12-13
Set apostrophes to uniform style in documentation

DoubleCookies:
Some of apostrophes in documentation looks different.


ztplz:
Looks good to me. Thanks.

PR #13766, opened 2018-12-21
fix(Table):fix arguments and type definition in TableRowSelection

hahabazinga:
* [x] Make sure that you propose pull request to right branch: bugfix for `master`, feature for branch `feature`.
* [x] Rebase before creating a pull request to keep commit history clear.
* [x] Add some descriptions and refer relative issues for you pull request.

Extra checklist:

**if** *isBugFix* **:**
fix issue [13761](https://github.com/ant-design/ant-design/issues/13761)
1. since interface `TableRowSelection` has already used Generics `T`, we can use `T` to avoid type 
 assertion when we call member functions in `TableRowSelection`. 
2. Besides, I noticed that function `onSelectInvert` receives `selectedRowKeys` as an argument in [`table.tsx`](https://github.com/ant-design/ant-design/blob/master/components/table/Table.tsx), not `selectedRows`，so I fix it

### DIFF
--- a/components/table/interface.tsx
+++ b/components/table/interface.tsx
@@ -85,11 +85,11 @@ export type TableSelectWay = 'onSelect' | 'onSelectMultiple' | 'onSelectAll' | '
 export interface TableRowSelection<T> {
   type?: RowSelectionType;
   selectedRowKeys?: string[] | number[];
-  onChange?: (selectedRowKeys: string[] | number[], selectedRows: Array<T>) => void;
+  onChange?: (selectedRowKeys: string[] | number[], selectedRows: T[]) => void;
   getCheckboxProps?: (record: T) => Object;
   onSelect?: SelectionSelectFn<T>;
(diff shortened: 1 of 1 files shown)


components/table/interface.tsx
```diff
@@ -85,12 +85,12 @@ export type TableSelectWay = 'onSelect' | 'onSelectMultiple' | 'onSelectAll' | '
 export interface TableRowSelection<T> {
   type?: RowSelectionType;
   selectedRowKeys?: string[] | number[];
-  onChange?: (selectedRowKeys: string[] | number[], selectedRows: Object[]) => void;
+  onChange?: (selectedRowKeys: string[] | number[], selectedRows: Array<T>) => void;
```
yesmeck: prefer `T[]` instead of `Array<T>`

hahabazinga: have changed it to fit in unified style

yesmeck (approved):

PR #13828, opened 2018-12-25
Update Santa.jsx

sfturing:
Update Santa.jsx Chrismas to Christmas


acerest:
Aha, find too, late for issuing

yutingzhao1991 (approved):

PR #13831, opened 2018-12-25
fix: Christmas typo

yutingzhao1991:


zy410419243:
https://github.com/ant-design/ant-design/pull/13828

重了

PR #13931, opened 2018-12-26
fix(table): added z-index to fixed columns

### LINKED ISSUE #13930: Cause z-index error when used Badge in fixed Table columns

### Version
3.11.6

### Environment
macOS Mojave, Chrome 71, React 16

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/6l3mxqkvxz)

### Steps to reproduce
Scroll the table

### What is expected?
Correct vertical order

Incorrect vertical order
image

kenve:
Solved #13930 

* [x] Make sure that you propose pull request to right branch: bugfix for `master`, feature for branch `feature`.
* [x] Rebase before creating a pull request to keep commit history clear.
* [x] Add some descriptions and refer relative issues for you pull request.

Extra checklist:

### DIFF
--- a/components/table/style/index.less
+++ b/components/table/style/index.less
@@ -593,6 +593,7 @@
     overflow: hidden;
     transition: box-shadow 0.3s ease;
     border-radius: 0;
+    z-index: 20;
     table {
       width: auto;
       background: @component-background;


yutingzhao1991:
I am worried about whether it will cause other bug if we modify the `z-index`. Is there are any `z-index` rules in antd?

kenve:
Yeah, i am also worried that it will bring other problems. I just tested this pr #13797, it works fine! 

And it  also works well with table filter panel (`z-index:1050`) when i used the fixed columns. As long as the `z-index` of the table content is less than 20, and the filter , there will be no problem. 

This problem is caused by the pr #13143 was Merged,  and  the `count` of the Badge is set to `z-index: 10`.

So you better collect all the `z-index` to sort.

kenve:
Close, ref #14036

PR #14003, opened 2018-12-28
Dropdown overlay support function callback

zombieJ:
### This is a ...

- [x] New feature

### What's the background?

1. when there are lots of items that need right click menu, like a datagrid editor. create a menu element for every single grid cell during rendering is not necessary. Creating it when showing the menu would be much more efficient.

2. sometimes menu can be affected by other states, some global, some local, it's a waste of time if component has to re-renderer every time a state changes (not react state), even when the popup is not shown.

ref: #13944
  
### Realize of API
  
overlay support `() => React.ReactNode`:
```jsx
const menu = (
  <Menu>
    <Menu.Item>
      <a target="_blank" rel="noopener noreferrer" href="http://www.alipay.com/">1st menu item</a>
    </Menu.Item>
    <Menu.Item>
      <a target="_blank" rel="noopener noreferrer" href="http://www.taobao.com/">2nd menu item</a>
    </Menu.Item>
    <Menu.Item>
      <a target="_blank" rel="noopener noreferrer" href="http://www.tmall.com/">3rd menu item</a>
    </Menu.Item>
  </Menu>
);

ReactDOM.render(
  <Dropdown overlay={() => menu}>
    <a className="ant-dropdown-link" href="#">
      Hover me <Icon type="down" />
    </a>
  </Dropdown>,
  mountNode
);
```
  
### What's the affect?

changlog:
* Dropdown's `overlay` prop support function callback.
* Dropdown 的 `overlay` 属性支持函数调用。

### Additional Plan?

None

### DIFF
--- a/components/dropdown/dropdown.tsx
+++ b/components/dropdown/dropdown.tsx
@@ -16,9 +16,12 @@ const Placements = tuple(
   'bottomRight',
 );
 type Placement = (typeof Placements)[number];
+
+type OverlayFunc = () => React.ReactNode;
+
 export interface DropDownProps {
   trigger?: ('click' | 'hover' | 'contextMenu')[];
-  overlay: React.ReactNode;
+  overlay: React.ReactNode | OverlayFunc;
   onVisibleChange?: (visible?: boolean) => void;
   visible?: boolean;
   disabled?: boolean;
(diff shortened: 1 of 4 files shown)


chenshuai2144 (commented):
👍

chenshuai2144 (approved):
💪

PR #14181, opened 2019-01-08
Standardize @input-bg across all .ant-input uses

sarendipitee:
Makes it consistent when overriding @input-bg and @input-border variables
### This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Branch merge
- [x] Other (about what?)

### What's the background?

> 1. Describe the source of requirement.

It should be possible to _consistently_ style all "input" like components with "@input-x" less vars. 

> 2. Resolve what problem.

Currently AutoComplete/Select have different styles and hardcoded `background: transparent` making this difficult.

> 3. Related issue link.
 N/A

Could potentially affect styling if user had to create custom overrides to get around this

> `@input-bg` and `@input-border` less vars are now applied consistently to Input, InputNumber, Select, DatePicker, and AutoComplete components

Only style changes.

### Self Check before Merge


components/auto-complete/style/index.less
```diff
@@ -44,7 +44,7 @@
     }
 
     .@{input-prefix-cls} {
-      background: transparent;
```
afc163: `background: transparent;` could be useful, we can't change to white.

components/input-number/style/index.less
```diff
@@ -73,7 +70,7 @@
     -moz-appearance: textfield;
     height: @input-height-base - 2px;
     transition: all 0.3s linear;
-    background-color: transparent;
+    background-color: @input-bg;
```
afc163: https://github.com/ant-design/ant-design/commit/37ec04b21a662dd00fb7e729c6df7d53a082b5b1

components/auto-complete/style/index.less
afc163: same as other `transparent` changes.

components/input-number/style/index.less
JonDum: I'll have to look into this one a bit more.

JonDum:
@afc163 For context, trying to achieve a look like this:
screen shot 2019-01-11 at 1 05 04 am
is not possible with variables alone. There's no way to set a "global" @input-bg, which is what this is trying to rectify.

components/auto-complete/style/index.less
JonDum: Perhaps we can change value of `@input-bg` in default theme to `transparent`? Should allow more maximum use-cases / flexibility since default `@component-background` is already `#fff`

JonDum:
@afc163 Went ahead and made `@input-bg` default to `transparent` and everything looks good across Input, InputNumber, Select/AutoComplete, etc

zombieJ:
default background should keep `#FFF` or it will break some user with colored background:

JonDum:
@zombieJ Although that looks pretty cool, the whole point of this is that you can fix that across all input-like inputs by just overriding @input-bg. 

For example, to get a decent look all I had to do was override:
```
@body-background: #333;
@component-background: #333;
@border-color-base: #555; // base border outline a component
@input-bg : #444;
```
And I end up with:
screen shot 2019-02-14 at 1 14 35 am

Without this PR, that is not possible at all through variables and you'd have to have several custom global selectors to override it for AutoComplete, Select, Input, InputNumber, etc, etc

zombieJ:
I mean it's ok to use set background color by the same `@input-bg`. But should not change `@input-bg: #fff;` to `@input-bg: transparent;`.

JonDum:
> I mean it's ok to use set background color by the same @input-bg. But should not change @input-bg: #fff; to @input-bg: transparent;.

@zombieJ Why is that a problem? If user has changed background, then they're likely already doing CSS/variable overrides so it should be fine to expect them to also change @input-bg.

zombieJ:
> @zombieJ Why is that a problem? If user has changed background, then they're likely already doing CSS/variable overrides so it should be fine to expect them to also change @input-bg.

antd always set component background color as #FFF. And input under other component should be transparent cause it should pass the background by container background color. So it is like:
* @input-bg: @component-background (#FFF)
* Not change internal input background color (keep transparent) and should change @component-background instead:
  * components/auto-complete/style/index.less
  * components/date-picker/style/Calendar.less
  * components/input-number/style/index.less
  * components/select/style/index.less

PR #14200, opened 2019-01-09
should not call afterClose when change visible from false to true

tangjinzhou:
### This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Branch merge
- [ ] Other (about what?)

### What's the background?

Tag should not call afterClose when change visible from false to true

### DIFF
--- a/components/tag/__tests__/index.test.js
+++ b/components/tag/__tests__/index.test.js
@@ -35,6 +35,18 @@ describe('Tag', () => {
     expect(wrapper.find('div.ant-tag:not(.ant-tag-hidden)').length).toBe(1);
   });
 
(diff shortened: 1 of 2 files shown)


dengfuping:
@tangjinzhou `afterClose` does not be called when `visible` changes from `false` to `true` in this [demo](https://codesandbox.io/s/3xzop645o1). Could you provide a re-producible demo to show the problem?

tangjinzhou:
https://codesandbox.io/s/p319y5pykq
3.11.4版本重构tag之后引入问题

PR #14600, opened 2019-01-28
fix destroyFns not reduce when instance.destroy()

xu-snow:
### This is a ...

- [x] Bug fix

### What's the background?

> 1. in confirm.destroy(), `destroyFns` will never contain `destroy` , it is simple code problem, `destroyFns` contain `close` instead
> 2. it will make the  `destroyFns` not change when user manually close confirm dialog

> 1. fixbug,confirm `destroy()` has unreachable code
> 2. fixbug, 销毁对话框函数  `destroy()` 有永远不会执行到的代码块

### Self Check before Merge

- [x] Doc is or not needed
- [x] Demo not needed
- [x] TypeScript not needed
- [x] Changelog not needed

### DIFF
--- a/components/modal/__tests__/confirm.test.js
+++ b/components/modal/__tests__/confirm.test.js
@@ -1,4 +1,5 @@
 import Modal from '..';
+import { destroyFns } from '../Modal'
 
 const { confirm } = Modal;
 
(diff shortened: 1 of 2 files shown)


ycjcl868:
close is right.

image

PR #14602, opened 2019-01-28
fix IE 11 bug

phoenisx:

### This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Branch merge
- [ ] Other (about what?)

## Context
If `flex-basis` is not set, it is taken as `0px` as default (in IE), breaking the ui in IE 11.

### Bug:
screenshot 2019-01-28 at 3 07 11 pm

### Fix:
image 1

### Self Check before Merge

### DIFF
--- a/components/list/style/index.less
+++ b/components/list/style/index.less
@@ -44,7 +44,7 @@
         margin-right: @list-item-meta-avatar-margin-right;
       }
       &-content {
-        flex: 1 0;
+        flex: 1 0 auto;
       }
       &-title {
         color: @text-color;


chunlea:
Sadly, this PR has broken `ant-design-list-content` in other browsers! List content with too long content will be overflow the box. Like this:

screenshot 2019-02-10 at 10 12 15 am

but what we expect should be:

screenshot 2019-02-10 at 10 13 30 am

I think maybe the reason broken in IE10 is that the PR owner used the wrong thing about List. Like the <a/>. 

screenshot 2019-02-10 at 10 14 55 am

So I hope @Shub1427 can provide a demo about what he met the bug in IE11, and then we can test it whether it's a bug of List flex or not.

But now, after this PR, all list contents in my app are broken in Chrome : (

I had to override this to `flex: 1 0;`

Shub1427:
@chunlea You are right, I messed this css. My apologies. Will revert this PR.

P.S. - In our code-base some `inline-block` is used as a wrapper around List Meta, which messes, List in IE (when used inside a Drawer (having default `word-wrap: break-word;`)). Found this, while trying to replicate this issue in a sandbox.

yoyo837:
@chunlea How about `flex: 1 1 auto;`

chunlea:
@yoyo837 `flex: 1 1 auto;` works, but I think we should wait for @Shub1427 to provide a sandbox and to replicate this issue, and then we can figure out how to fix this but.

PR #14720, opened 2019-02-04
fix: throw error correctly by dekko

orzyyyy:
### This is a ...

- [x] Bug fix

### What's the background?
Check [there](https://circleci.com/gh/ant-design/ant-design/51990?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link), `node ./tests/dekko/lib.test.js`.

It always throws error even if there is nothing wrong. This is for the problem.


zy410419243:
I will check ci breaks later, going out now.

Issue #15297, opened 2019-03-09
card mode tabs component style is broken when browser scale is not 100%


rinick:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://ant.design/components/tabs/](https://ant.design/components/tabs/)

### Steps to reproduce
scale the browser up to something that's not 100% and check the style of tab bar

### What is expected?
the border line only shows in inactive tabs, active tab doesn't have border line

### What is actually happening?
the active tab sometimes also shows border line. when using retina display, you might see a line with half thickness.

| Environment | Info |
|---|---|
| antd | 3.15.0 |
| React | 16.7.0 |
| System | windows 10 |
| Browser | chrome 71 |

---
the problem is, the tab's height is not integer, it's 39.6px. using this to block a line at 40px does not always work

PR #15299, opened 2019-03-09
fix tabs style in card mode

rinick:
### 🤔 This is a ...
- [x] Bug fix

### 👻 What's the background?

resolves issue #15297 `card mode tabs component style is broken when browser scale is not 100%` 

### 💡 Solution

give tabs a fixed height, so it would be 40px instead of 39.6px

### 📝 Changelog description

> Describe changes from user side, and list all potential break changes or other risks.

### DIFF
--- a/components/tabs/style/card-style.less
+++ b/components/tabs/style/card-style.less
@@ -12,6 +12,7 @@
     visibility: hidden;
   }
   &&-card &-card-bar &-tab {
+    height: @tabs-card-height - 1px;
     margin: 0;
     margin-right: 2px;
     padding: 0 16px;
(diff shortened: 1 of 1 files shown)


afc163:
It will resolve #15234 too.

zombieJ:
Seems this makes < 90% style not correct...
屏幕快照 2019-03-09 下午6 29 06

rinick:
> Seems this makes < 90% style not correct...
> 屏幕快照 2019-03-09 下午6 29 06
the second tab's height is 39px, and browser shows 38.99px when I scale it to 90%.
It's really annoying that it can still block a line at 40px.
But I have another idea, let me try fix it in a different PR

PR #15457, opened 2019-03-17
Fix tab style

rinick:
### 🤔 This is a ...

- [x] Bug fix

### 👻 What's the background?

#15299
#15297
#15234

@zombieJ pointed out the previous PR doesn't fully fix the issue

### 💡 Solution

stop show hide the line based on if tab block the bottom border
new style would always block the border, and if a line is needed, show the line inside the tab


rinick:
this PR has too many local merge commits,
I will create a new PR with clean commits

close this one

PR #15760, opened 2019-03-30
Add name attribute support to Checkbox.Group

### LINKED ISSUE #15751: Support `name` attribute for Checkbox and Checkbox.Group

`Checkbox` and `Checkbox.Group` currently do not support/render a `name` property/attribute, respectively. Doing so would better facilitate universal/isomorphic applications and would be consistent with the `Radio` and `Radio.Group` components.

I propose to add a `name?: string` type prop to both the `Checkbox` and `Checkbox.Group` components. If given a value it should be rendered as a name attribute for the checkbox input(s).

bencallaway:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?

1. Describe the source of requirement, like related issue link.
This PR resolves #15751 and is related to #15753.

2. Describe the problem and the scenario.
`Checkbox.Group` does not render the `name` attribute for children inputs as `Radio.Group` does. Adding support for this property would make its API more consistent with `Radio.Group`, and therefore easier to work with both.

### 💡 Solution

1. How to fix the problem, and list final API implementation and usage sample if that is a new feature. 
This should work the same as `Radio.Group` as implemented in #7009.

2. GIF or snapshot should be provided if includes UI/interactive modification.
n/a

### 📝 Changelog description

> Describe changes from userside, and list all potential break changes or other risks.

1. English description
This only extends the API of `Checkbox.Group` with an optional prop. It should present no breaking changes.

2. Chinese description (optional)

### DIFF
--- a/components/checkbox/Checkbox.tsx
+++ b/components/checkbox/Checkbox.tsx
@@ -100,6 +100,7 @@ export default class Checkbox extends React.Component<CheckboxProps, {}> {
         }
         checkboxGroup.toggleOption({ label: children, value: props.value });
       };
+      checkboxProps.name = checkboxGroup.name;
       checkboxProps.checked = checkboxGroup.value.indexOf(props.value) !== -1;
       checkboxProps.disabled = props.disabled || checkboxGroup.disabled;
     }
--- a/components/checkbox/Group.tsx
+++ b/components/checkbox/Group.tsx
@@ -81,6 +81,7 @@ class CheckboxGroup extends React.Component<CheckboxGroupProps, CheckboxGroupSta
         toggleOption: this.toggleOption,
         value: this.state.value,
         disabled: this.props.disabled,
+        name: this.props.name,
       },
     };
   }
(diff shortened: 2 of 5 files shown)


bencallaway:
Currently `lint:ts` fails at the `HEAD` of `feature` branch. I am getting only those same failures on my PR branch.

zombieJ:
> After rebasing my branch onto the latest `feature` I'm no longer receiving the TS errors

Thanks for comment. Seems it's safe to close this.

bencallaway:
@zombieJ this was not just a comment. It was a pull request for adding a feature. Do you not think it's a good idea to support the `name` property in `Checkbox.Group` in the same way `Radio.Group` supports it? This PR adds a test and implements that support. #15753 did add the optional `name` prop to the interface, but `Checkbox.Group` still did not do anything with it. Please reconsider this PR for merge. As without this functionality I will have to either work from a fork or create a hacky HOC to accommodate this API, which I believe belongs quite well with `Checkbox.Group`.

zombieJ:
@bencallaway , Got it.

zombieJ (approved):

bencallaway:
Thanks for taking another look, @zombieJ

PR #16081, opened 2019-04-15
fix: If there is an existing debounce in progress, cancel that

raybooysen:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?

When you have a Spin component and componentDidUpdate is called, any existing debounce() call is not cancelled.

### 💡 Solution

On every time we debounce, check and cancel if appropriate.

### 📝 Changelog

- English Changelog: Spin component cancels debounce calls between update calls

### DIFF
--- a/components/spin/index.tsx
+++ b/components/spin/index.tsx
@@ -98,6 +98,10 @@ class Spin extends React.Component<SpinProps, SpinState> {
   }
 
   componentWillUnmount() {
+    this.cancelExistingSpin();
+  }
+
+  cancelExistingSpin() {
     const updateSpinning: any = this.updateSpinning;
     if (updateSpinning && updateSpinning.cancel) {
       updateSpinning.cancel();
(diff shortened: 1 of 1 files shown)


components/spin/index.tsx
```diff
@@ -116,6 +120,7 @@ class Spin extends React.Component<SpinProps, SpinState> {
   debouncifyUpdateSpinning = (props?: SpinProps) => {
     const { delay } = props || this.props;
     if (delay) {
+      this.cancelExistingSpin();
```
raybooysen: Unsure if this should be outside the if and cancelled in all cases.  Can someone advise?

raybooysen:
Fixes #15623

PR #16129, opened 2019-04-16
style: less variables for comment component font-size

SamKirkland:
### 🤔 This is a ...
- [x] Bug fix

### 👻 What's the background?
The comment component used hard-coded font-sizes in it's .less file instead of using the font-sizes defined in default.less

### 💡 Solution
I added 2 new font-size less vars in default.less (following existing patterns) and set them the the default font-size set in the theme.

### 📝 Changelog

- English Changelog: Comment component font-size less variables


SamKirkland:
This pull request is 2 months old and unreviewed, I'll be closing and creating a new one.

PR #16143, opened 2019-04-17
Button icon core is not fully handled

mikedotJS:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?

We often use Button with the icon property but we aren't able to set an icon theme directly with Button component declaration. We could have used a nested icon but what i'd say is : If the Button component has an `icon` property, it could handle it all the way.

### 💡 Solution

I added two new props to the `Button` component :

1. `iconTheme`, a `union` set to either 'filled' or 'outlined' or 'twoTone'
2. `iconTwoToneColor`, a `string` set to the primary two tone color as in the `Icon` component API

### 📝 Changelog

* User can now set `iconTheme` and `iconTwoToneColor` directly to the Button component declaration

-----
[View rendered components/button/index.en-US.md](https://github.com/mikedotJS/ant-design/blob/button-icon-is-not-fully-handled/components/button/index.en-US.md)

### DIFF
--- a/components/button/button.tsx
+++ b/components/button/button.tsx
@@ -48,10 +48,14 @@ const ButtonSizes = tuple('large', 'default', 'small');
 export type ButtonSize = (typeof ButtonSizes)[number];
 const ButtonHTMLTypes = tuple('submit', 'button', 'reset');
 export type ButtonHTMLType = (typeof ButtonHTMLTypes)[number];
+const IconThemes = tuple('filled', 'outlined', 'twoTone');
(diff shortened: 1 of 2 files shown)


zombieJ:
Since we will try to reduce the icon size of package. `icon` on Button may not future update.
ref: https://github.com/ant-design/ant-design/issues/12011

Issue #16194, opened 2019-04-19
Tooltip Type definitions do not allow 'className' prop


swillis12:
- [X] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/0pv59x7pvp)

### Steps to reproduce
Add a className prop to the Tooltip component in a .tsx file.

### What is expected?
Classname should not throw a TS error since this prop is actually valid and allowed.

### What is actually happening?
ts(2322) error is thrown since className does not exist in the Tooltip props interface.

| Environment | Info |
|---|---|
| antd | 3.16.3 |
| React | 16.8.6 |
| System | Mac |
| Browser | Chrome Latest |

PR #16195, opened 2019-04-19
Fix className prop issue when using Tooltip

swillis12:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?
1. Cannot add className prop to Tooltip when using Typescript: https://github.com/ant-design/ant-design/issues/16194.
2. TS error is thrown when adding this className prop.

### 💡 Solution
Tooltip Props now extend React.HTMLAttributes<HTMLElement>.

### 📝 Changelog

- English Changelog:

### DIFF
--- a/components/tooltip/index.tsx
+++ b/components/tooltip/index.tsx
@@ -81,8 +81,8 @@ const splitObject = (obj: any, keys: string[]) => {
   return { picked, omitted };
 };
 
-class Tooltip extends React.Component<TooltipProps, any> {
-  static defaultProps = {
+class Tooltip extends React.Component<TooltipProps & React.HTMLAttributes<HTMLElement>, any> {
+    static defaultProps = {
     placement: 'top' as TooltipPlacement,
     transitionName: 'zoom-big-fast',
     mouseEnterDelay: 0.1,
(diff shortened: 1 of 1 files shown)


swillis12:
Not sure if this is the correct fix, please advise if not.

ztplz:
What confuse me is why you want to add `classname` to `tooltip`,  I am not sure if it will work.

swillis12:
> What confuse me is why you want to add `classname` to `tooltip`, I am not sure if it will work.

It does work, it is for applying text-overflow style to the tooltip so I don't have to have an extra layer of DOM nesting.

components/tooltip/index.tsx
```diff
@@ -81,8 +81,8 @@ const splitObject = (obj: any, keys: string[]) => {
   return { picked, omitted };
 };
 
-class Tooltip extends React.Component<TooltipProps, any> {
-  static defaultProps = {
+class Tooltip extends React.Component<TooltipProps & React.HTMLAttributes<HTMLElement>, any> {
+    static defaultProps = {
```
yesmeck: Unnecessary indent.

yesmeck (changes requested):

components/tooltip/index.tsx
swillis12: Thanks, I have fixed this in https://github.com/ant-design/ant-design/pull/16195/commits/fec88af3d342d719c2c4c0f7210702e7d1d015e6

zombieJ:
merge master

PR #16289, opened 2019-04-25
feat: Button support `link` type

### LINKED ISSUE #15892: Allow Button option to make it look exactly like Hyperlink

When you want something to visually look like a Hyperlink, but it doesn't actually change your URL, you should definitely use a `<button type="button">` element.

At present, there is no way to style a `<Button />` component to look visually like a hyperlink

An example might be a "Forgot password?" button, which shouldn't appear as prominent as an actual `<button>` - being styled like an `<a>` is perfect.  But it doesn't actually change your URL, it just toggles the Login form to replace with Forgot Password form - so it should definitely be an HTML `<button>` element.

This is best practice for semantic HTML.  To have an `<a>` element do this, you either have:
 - no href (then you can't use keyboard to TAB to the `<a>` which breaks keyboard users completely)
 - a fake href (like `#` or `javascript:void(0)`) <-- screen readers can read this out, and it's bad.  Also you need to remember to `event.preventDefault` in the handler, which isn't good.

Better semantic HTML = better SEO & better accessibility, which is so important these days.

Bootstrap does this with classes `btn btn-link`

add `type="link"` or something

Only different behaviour is visual appearance is now exactly like an `<a>`

zombieJ:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?

close #15892 

### 💡 Solution

Add one more type & style

### 📝 Changelog

Button support `link` type.

-----
[View rendered components/button/demo/basic.md](https://github.com/ant-design/ant-design/blob/link-button/components/button/demo/basic.md)
[View rendered components/button/demo/block.md](https://github.com/ant-design/ant-design/blob/link-button/components/button/demo/block.md)
[View rendered components/button/demo/disabled.md](https://github.com/ant-design/ant-design/blob/link-button/components/button/demo/disabled.md)
[View rendered components/button/demo/ghost.md](https://github.com/ant-design/ant-design/blob/link-button/components/button/demo/ghost.md)
[View rendered components/button/demo/size.md](https://github.com/ant-design/ant-design/blob/link-button/components/button/demo/size.md)
[View rendered components/button/index.en-US.md](https://github.com/ant-design/ant-design/blob/link-button/components/button/index.en-US.md)
[View rendered components/button/index.zh-CN.md](https://github.com/ant-design/ant-design/blob/link-button/components/button/index.zh-CN.md)
[View rendered components/comment/demo/list.md](https://github.com/ant-design/ant-design/blob/link-button/components/comment/demo/list.md)

### DIFF
--- a/components/button/button.tsx
+++ b/components/button/button.tsx
@@ -40,7 +40,7 @@ function insertSpace(child: React.ReactChild, needInserted: boolean) {
   return child;
 }
 
-const ButtonTypes = tuple('default', 'primary', 'ghost', 'dashed', 'danger');
+const ButtonTypes = tuple('default', 'primary', 'ghost', 'dashed', 'danger', 'link');
 export type ButtonType = (typeof ButtonTypes)[number];
 const ButtonShapes = tuple('circle', 'circle-outline', 'round');
 export type ButtonShape = (typeof ButtonShapes)[number];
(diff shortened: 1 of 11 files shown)


chenshuai2144 (approved):

Stephen2:
This is great!

afc163 (approved):

components/button/button.tsx
```diff
@@ -266,20 +266,24 @@ class Button extends React.Component<ButtonProps, ButtonState> {
...
-          onClick={this.handleClick}
-          ref={this.saveButtonRef}
-        >
-          {iconNode}
-          {kids}
-        </button>
-      </Wave>
+    const buttonNode = (
+      <button
+        {...otherProps as NativeButtonProps}
```
ztplz: 这里`NativeButtonProps`不怎么准确  应该是`React.ButtonHTMLAttributes<any>`

PR #16322, opened 2019-04-25
fix: use mounted property, replace mounted state, resolves #16290

### LINKED ISSUE #16290: destroyed Menu right after it's created will cause React warning

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/6y1px2jr5n)

### Steps to reproduce
open the link and there is a console warning
```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. 
To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
```

### What is expected?
it creates a dropdown, open the menu, and then destroy the dropdown

the menu has a 10ms delay to `setState({mounted: true})` but at that time component is already unmounted

| Environment | Info |
|---|---|
| antd | 3.16.5 |
| React | 16.8.6 |
| System | windows 10 |
| Browser | chrome 71 |

---
1. `raf.cancel(this.mountRafId);` is [actually called](https://github.com/ant-design/ant-design/blob/master/components/menu/index.tsx#L173), but it doesn't seem to work
2.  using react setState is a hackish way to maintain a `mounted` state ,  and it can cause unnecessary re-render.  I think this should be done as a normal js class property, so there is no need for such delay.

rinick:
### 🤔 This is a ...

- [x] Bug fix

### 👻 What's the background?

resolves issue : destroyed Menu right after it's created will cause React warning #16290

### 💡 Solution

use mounted property instead of mounted state

### DIFF
--- a/components/menu/__tests__/index.test.js
+++ b/components/menu/__tests__/index.test.js
@@ -3,7 +3,6 @@ import { mount } from 'enzyme';
 import Menu from '..';
 import Icon from '../../icon';
 import Layout from '../../layout';
-import raf from '../../_util/raf';
 
 jest.mock('mutationobserver-shim', () => {
   global.MutationObserver = function MutationObserver() {
@@ -74,10 +73,6 @@ describe('Menu', () => {
         .at(0)
         .hasClass('ant-menu-hidden'),
     ).not.toBe(true);
-
-    const rafCount = Object.keys(raf.ids).length;
-    wrapper.unmount();
-    expect(Object.keys(raf.ids).length).toBe(rafCount - 1);
   });
 
   it('should accept defaultOpenKeys in mode vertical', () => {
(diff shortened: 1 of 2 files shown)


zombieJ:
This should fix in `raf` instead of change to `this.mounted = xxx`.

zombieJ:
close since #16370

rinick:
> This should fix in `raf` instead of change to `this.mounted = xxx`.

@zombieJ 
would this still cause an unnecessary render() ?
I think the whole point of using state is because we want to let react framework decide when to render.
if set `mounted=true` doesn't require a re-render, then there is no reason to use state

zombieJ:
Since React will release concurrent mode in ~Q2 2019, we use [`this.state` instead of `this.[variable]`](https://github.com/ant-design/ant-design/commit/89dd827b7f00aee8c1af673687d3c66269d4a928#diff-b5d7ebd6926cbb3981c237f2e491be13L91) to ensure it not breaks in concurrent mode.
I know use `this.[variable]` is still safe in some condition, but use pure state is much easier.

PR #16340, opened 2019-04-26
fix: Allow Object.entries (only for type checks, this is not a polyfill)

sbusch:
### 🤔 This is a ...

- [x] TypeScript definition update (global definition via `tsconfig.json`)

### 👻 What's the background?

Fixes the following linting error for antd authors (run via `npm run tlint` or implicitly e.g. after commits):

```shell
> tsc
components/progress/Line.tsx:21:37 - error TS2339: Property 'entries' does not exist on type 'ObjectConstructor'.

21   for (const [key, value] of Object.entries(gradients)) {
                                       ~~~~~~~
Found 1 error.
```

`Object.entries()` is used in `components/progress/Line.tsx`.

### 💡 Solution

Change in `tsconfig.json` lets TypeScript allows calls to `Object.entries`.

Note: this is only for type checks, tsc will assume that there's a polyfill in place, which is [required at least for IE11](https://caniuse.com/#feat=object-entries).

Polyfills are mentioned [in the antd Documentation](https://ant.design/docs/react/getting-started#Compatibility), but I think it is necessary to also exactly specify the required polyfills there,  `Object.entries` is one of them.

- 👨‍🚒 Doc update is needed IMO **See my note about polyfills**
- Demo is not needed
- TypeScript definition is updated (the change itself)
- Changelog not needed **I'm not aware of a changelog for antd authors, where this change belongs to**

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -9,7 +9,7 @@
     "noUnusedLocals": true,
     "noImplicitAny": true,
     "target": "es6",
-    "lib": ["dom", "es7"]
+    "lib": ["dom", "es2017"]
   },
   "exclude": ["node_modules", "lib", "es"]
 }


sbusch:
I'm wondering how this bug got past the automated linting during publish or consecutive commits. Is there a blind spot?

zombieJ:
`es7` already includes this. Please check if you meat the same problem: https://stackoverflow.com/questions/39741428/es7-object-entries-in-typescript-not-working

sbusch:
@zombieJ please reopen:

`es7` is an alias for `es2016`, which is ECMAScript 2016. See the [Wikipedia page about ECMAScript](https://en.wikipedia.org/wiki/ECMAScript#Versions) for info about confusing ECMAScript versioning scheme.  (Fortunately, they stopped using the shorter, one-off version so starting with "es2017", there never was a "es8".)

ECMAScript 2016 does **not** include `Object.entries`. [According to the Spec](http://www.ecma-international.org/ecma-262/8.0/index.html#sec-object.entries), `Object.entries()` was added in ECMAScript 2017.

TS library definitions correspond to that: [es2017.d.ts](https://github.com/Microsoft/TypeScript/blob/master/src/lib/es2017.d.ts) includes [es2017.object.d.ts](https://github.com/Microsoft/TypeScript/blob/master/src/lib/es2017.object.d.ts) (that one is reference in my PR) which adds definition for `Object.entries`.

So I think my PR is correct and should be merged

zombieJ:
@sbusch , thanks for description. Could you help change to es2017 instead of add es2017.object?

sbusch:
Thanks!

es2017: sure. But the required change in `tsconfig.json` is a one-liner, where should I help in addition to that?

zombieJ:
Update to `es2017` is help enough : )

sbusch:
@zombieJ Done

What is your opinion on documentation of required polyfills?

At least it should be noted on docs that runtime support for ES 2017 is expected, maybe here: <https://ant.design/docs/react/getting-started#Compatibility>

zombieJ:
CI failed. Need merge master.

zombieJ:
> @zombieJ Done
> 
> What is your opinion on documentation of required polyfills?
> 
> At least it should be noted on docs that runtime support for ES 2017 is expected, maybe here: https://ant.design/docs/react/getting-started#Compatibility

Since we recommend using `babel-preset-env`. It can includes what antd required by setting `targets`. May not need additional info about this.

sbusch:
@zombieJ Merged

PR #16342, opened 2019-04-26
fix: Add some missing TypeScript prop definititions and clean up date-picker definitions

sbusch:
### 🤔 This is a ...

### 👻 What's the background?

Some props were missing from TypeScript definitions.

TypeScript definitions for date-picker Components were a bit redundant.

### 💡 Solution

Props have been added, redundancies for date-picker Components have been removed.

### 📝 Changelog

No special changelog

- Doc is not needed (props are mostly standard props that are passed to the underlying HTML elements. They are most often not documented. date-picker changes doesnt need to be documented)
- Demo is not needed (same reason)
- TypeScript definition is updated (the purpose of this PR)
- Changelog is not needed

### DIFF
--- a/components/date-picker/interface.tsx
+++ b/components/date-picker/interface.tsx
@@ -31,7 +31,7 @@ export interface SinglePickerProps {
   defaultValue?: moment.Moment;
   defaultPickerValue?: moment.Moment;
   placeholder?: string;
-  renderExtraFooter?: (mode: 'date' | 'month' | 'year' | 'decade') => React.ReactNode;
+  renderExtraFooter?: (mode: DatePickerMode) => React.ReactNode;
   onChange?: (date: moment.Moment, dateString: string) => void;
 }
 
--- a/components/date-picker/interface.tsx
+++ b/components/date-picker/interface.tsx
@@ -35,7 +35,7 @@ export interface SinglePickerProps {
   onChange?: (date: moment.Moment, dateString: string) => void;
 }
 
-const DatePickerModes = tuple('time', 'date', 'month', 'year');
+const DatePickerModes = tuple('time', 'date', 'month', 'year', 'decade');
 export type DatePickerMode = (typeof DatePickerModes)[number];
 
 export interface DatePickerProps extends PickerProps, SinglePickerProps {
--- a/components/date-picker/interface.tsx
+++ b/components/date-picker/interface.tsx
@@ -24,12 +24,14 @@ export interface PickerProps {
   disabledDate?: (current: moment.Moment | undefined) => boolean;
   renderExtraFooter?: () => React.ReactNode;
   dateRender?: (current: moment.Moment, today: moment.Moment) => React.ReactNode;
+  autoFocus?: boolean;
 }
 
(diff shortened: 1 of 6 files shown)


zombieJ:
Seems conflict.

sbusch:
Conflicts resolved.

components/date-picker/interface.tsx
```diff
@@ -31,6 +31,7 @@ export interface SinglePickerProps {
   defaultValue?: moment.Moment;
   defaultPickerValue?: moment.Moment;
   placeholder?: string;
+  renderExtraFooter?: (mode: 'date' | 'month' | 'year' | 'decade') => React.ReactNode;
```
zombieJ: use `DatePickerMode` instead.

ztplz (approved):

components/date-picker/interface.tsx
sbusch: @zombieJ done.

I also added `"decade"` to `DatePickerMode`, that was missing there. See commit for link to playground.

components/date-picker/interface.tsx
```diff
@@ -23,21 +23,22 @@ export interface PickerProps {
...
+  autoFocus?: boolean;
 }
 
 export interface SinglePickerProps {
   value?: moment.Moment;
   defaultValue?: moment.Moment;
   defaultPickerValue?: moment.Moment;
-  renderExtraFooter?: (mode: 'date' | 'month' | 'year' | 'decade') => React.ReactNode;
+  placeholder?: string;
+  renderExtraFooter?: (mode: DatePickerMode) => React.ReactNode;
```
zombieJ: `DatePickerMode | 'decade'`， `decade` is not export as mode.

sbusch: I don't understand this. Do you mean that `DatePickerMode` (which is exported) should not include `'decade'`?

But `'decade'` is also a valid value for the `mode` prop ([CodeSandbox example](https://codesandbox.io/s/0q5kmwr7kl)), so it should be included in `DatePickerMode`...

zombieJ: It's not in the document since we currently not export to user:
屏幕快照 2019-05-03 下午2 14 31

sbusch: Ok, then the docs are incomplete and should be fixed:

There is definitely a "decade" mode the user can use a value for the mode prop, and `renderExtraFooter()` then gets "decade" as first argument (parameter "mode). See the CodeSandbox linked in my previous comment (<https://codesandbox.io/s/0q5kmwr7kl>), where mode prop is pinned to "decade".

You can also take a look at the more complete CodeSandbox <https://codesandbox.io/s/388xplj646>

1. activate first date picker
2. click on year in header
3. click on decade in header
4. take a look at the footer

sbusch: I fixed the docs, but accidentally pushed too much into this PR branch😠

Sorry, will fix it in the next minutes.

sbusch: Fixed

PR #16344, opened 2019-04-26
fix: TypeScript definitions for Select component's onSelect / onDeselect props, update inaccurate docs

mathieu-anderson:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [x] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?
The `SelectProps` interface defines the `value` param for `onSelect` and `onDeselect` as `SelectValue` (https://github.com/ant-design/ant-design/blob/master/components/select/index.tsx#L50)

In the context of these handlers, as I understand it, `value` should never be an array, but always a single member of the array (`string`, `number` or `LabeledValue`). However, TS seems to not be able to correctly infer this. Therefore, typing the `value` param correctly in these handlers is not currently possible.

Moreover, the documentation seems to be inaccurate for the `value`, `defaultValue` and `onDeselect` props (especially as `onDeselect` does not get the `option` param, but is documented as such).

### 💡 Solution

* Define the `value` param for `onSelect` and `onDeselect` as `string|number|LabeledValue`
* Fix inaccurate documentation (partial update for CN docs : only type values)

### 📝 Changelog

- English Changelog: 
  * `Select` component : updated types for `onSelect` and `onDeselect` props to be `string|number|LabeledValue`
  * Update docs

### DIFF
--- a/components/select/index.tsx
+++ b/components/select/index.tsx
@@ -48,6 +48,7 @@ export interface LabeledValue {
 }
 
 export type SelectValue = string | string[] | number | number[] | LabeledValue | LabeledValue[];
+export type SelectedValue = string | number | LabeledValue;
 
 export interface SelectProps<T = SelectValue> extends AbstractSelectProps {
   value?: T;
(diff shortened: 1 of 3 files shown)


components/select/index.tsx
```diff
@@ -56,8 +56,8 @@ export interface SelectProps<T = SelectValue> extends AbstractSelectProps {
   optionLabelProp?: string;
   firstActiveValue?: string | string[];
   onChange?: (value: T, option: React.ReactElement<any> | React.ReactElement<any>[]) => void;
-  onSelect?: (value: T, option: React.ReactElement<any>) => any;
-  onDeselect?: (value: T) => any;
+  onSelect?: (value: string | number | LabeledValue, option: React.ReactElement<any>) => any;
```
zombieJ: Define an type instead~

mathieu-anderson: Good point, will do !

zombieJ:
CI failed. Merge master please.

mathieu-anderson:
Ugh sorry @zombieJ, made some kind of git mess trying to fix the build :|

zombieJ:
Seems it need rebase or cherry-pick to fix it.

mathieu-anderson:
@zombieJ cleaned up the git history, but the codebeat check seems to be stuck on pending. Does it require action?

PR #17009, opened 2019-06-07
fix: remove margin for collapse arrow

mtadams007:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [x] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?

1. Using overflow auto adds scrollbar on collapse component when there is a header below.

collapseScrollBar

2. We want to have a collapse component where we can scroll if the content is bigger than the height. Adding `height: 100%` and `overflow: auto` to the Collapse component causes a scrollbar to appear because the arrow within the header has a margin attached.

### 💡 Solution

1. Remove the added margin for the header arrow.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/collapse/style/index.less
+++ b/components/collapse/style/index.less
@@ -37,7 +37,6 @@
         top: 50%;
         left: @padding-md;
         display: inline-block;
-        margin-top: 2px;
         font-size: @font-size-sm;
         line-height: 46px;
         transform: translateY(-50%);
--- a/components/collapse/style/index.less
+++ b/components/collapse/style/index.less
@@ -38,8 +38,7 @@
         left: @padding-md;
         display: inline-block;
         font-size: @font-size-sm;
-        line-height: 46px;
-        transform: translateY(-21px);
+        transform: translateY(-50%);
 
         & svg {
           transition: transform 0.24s;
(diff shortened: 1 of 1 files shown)


zombieJ:
Remove margin-top makes icon little up. Could you adjust this 2px also:
屏幕快照 2019-06-08 下午10 13 01

components/collapse/style/index.less
```diff
@@ -39,7 +39,7 @@
         display: inline-block;
         font-size: @font-size-sm;
         line-height: 46px;
-        transform: translateY(-50%);
+        transform: translateY(-21px);
```
mtadams007: `-50%` would be -23px so by making it `-21px` we can eliminate the shift seen earlier

zombieJ: Maybe adjust `line-height` instead? Use `px` instead of `%` may little hard code.

mtadams007:
I put the transform back and removed `line-height` completely. It looked identical and had the arrow centered. I don't think that will affect anything else, but let me know if that's not acceptable.

zombieJ:
CI broken, rebase please~

PR #17010, opened 2019-06-08
fixing inaccurate types

dgreene1:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?

This fixes two issues:
1. When utilizing `form.validateFields`, the TypeScript definitions imply that the `values` parameter from the callback will be truthy. However, at runtime the `values` parameter of the callback is partially-hydrated (i.e. each of the properties on the object have the potential to be undefined). The goal of this change is to align the types with what is actually happening at runtime
2. I believe there was a typo with `getFieldDecorator` where it wasn't automatically inferring the type from `Form.create`. When the person submitted [the original PR](https://github.com/ant-design/ant-design/pull/11799), I believe they meant `V` but typed `T` instead.

### 💡 Solution

For problem 1, I changed the generic to be a partial generic
For problem 2, I corrected the generic so it can be inferred from the class generation

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/form/Form.tsx
+++ b/components/form/Form.tsx
@@ -76,7 +76,7 @@ export type ValidationRule = {
   validator?: (rule: any, value: any, callback: any, source?: any, options?: any) => any;
 };
 
-export type ValidateCallback<V> = (errors: any, values: V) => void;
+export type ValidateCallback<V> = (errors: any, values: Partial<V>) => void;
 
 export type GetFieldDecoratorOptions = {
   /** 子节点的值的属性，如 Checkbox 的是 'checked' */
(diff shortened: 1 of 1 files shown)


dgreene1:
What are your thoughts @afc163? I’ve seen that you have merged TypeScript changes in the past. Does this look good to you?

components/form/Form.tsx
```diff
@@ -180,8 +180,8 @@ export type WrappedFormUtils<V = any> = {
   /** 重置一组输入控件的值与状态，如不传入参数，则重置所有组件 */
   resetFields(names?: Array<string>): void;
   // tslint:disable-next-line:max-line-length
-  getFieldDecorator<T extends Object = {}>(
-    id: keyof T,
+  getFieldDecorator(
+    id: keyof V,
```
yesmeck: Don't remember why we use a generic type `T` here, but we can't use `keyof V`, because of a field key can be a nested key path.

dgreene1: @yesmeck can you describe what you mean by nested key path? Better yet, can you provide a snippet of what that object might look like?

Because I might be able to produce an even smarter type.

yesmeck: We can write `getFieldDecorator('foo.bar')` which produces a `{ foo: { bar: '' } }` value.

PR #16535, opened 2019-05-12
feat: New calendar apis, headerRender method

abgaryanharutyun:
Added new calendar APIs`renderHeader`
 - With `renderHeader` API now we can render custom header with all functionalities.

Thank you guys for creating an awesome UI for React. I am using ant for many years and I will be very happy to join to Ant team.

I will wait for feedbacks.

Thanks.

### 🤔 This is a ...

- [X] New feature
- [ ] Bug fix
- [X] Site / document update
- [ ] Component style update
- [ ] TypeScript definition update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Solution

Screen Shot 2019-05-12 at 5 38 32 PM
### 📝 Changelog

- English Changelog:
 Not added yet
- Chinese Changelog (optional):

### ☑️ Self Check before Merge

- [X] Doc is updated/provided or not needed
- [X] Demo is updated/provided or not needed
- [ ] TypeScript definition is updated/provided or not needed
- [ ] Changelog is provided or not needed

-----
[View rendered components/calendar/demo/card-with-custom-header-api.md](https://github.com/abgaryanharutyun/ant-design/blob/calendar-custome-header-and-small-size/components/calendar/demo/card-with-custom-header-api.md)
[View rendered components/calendar/index.en-US.md](https://github.com/abgaryanharutyun/ant-design/blob/calendar-custome-header-and-small-size/components/calendar/index.en-US.md)

PR #17063, opened 2019-06-12
fix: prop headerRender should be optional for component Calendar

wonderjar:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?

In Release 13.19.0, new prop headerRender is added for component Calendar(#16535 ), but the ts definition define it mandatory. It's designed as optional in fact.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/calendar/Header.tsx
+++ b/components/calendar/Header.tsx
@@ -12,6 +12,8 @@ export interface RenderHeader {
   onTypeChange: (type: string) => void;
 }
 
+export type HeaderRender = (headerRender: RenderHeader) => React.ReactNode;
+
 export interface HeaderProps {
   prefixCls?: string;
   locale?: any;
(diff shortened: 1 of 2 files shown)


components/calendar/Header.tsx
```diff
@@ -178,8 +180,8 @@ export default class Header extends React.Component<HeaderProps, any> {
     );
   };
 
-  headerRenderCustom = (): React.ReactNode => {
-    const { headerRender, type, onValueChange, value } = this.props;
```
yesmeck: I think this change is unnecessary.

wonderjar: I change here because
`
const { headerRender, type, onValueChange, value } = this.props;
`
Type of headerRender is    HeaderRender **| undefined**

If I pass headerRender as function argument, because the func is called by
`
headerRender ? this.headerRenderCustom(headerRender) : (...)
`
the type of headerRender is HeaderRender

I just want to avoid the repeated undefined check in TS

PR #17557, opened 2019-07-09
fix: Collapse Panel key type definition

thylsky:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?
1. Describe the source of requirement, like related issue link.
    Currently the allowed key for the `Collapse.Panel` is just string.

2. Describe the problem and the scenario.
    When you are rendering `Collapse.Panel` with map function, you can't simply provide just index of the map function. Also React's definition for Key type is `number | string`. 

### 💡 Solution

1. How to fix the problem, and list final API implementation and usage sample if that is an new feature.
    Fix is prepared in this PR: Changing type definition of the `key` props.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |          |
| 🇨🇳 Chinese |           |

-----
[View rendered components/collapse/index.en-US.md](https://github.com/thylsky/ant-design/blob/patch-1/components/collapse/index.en-US.md)
[View rendered components/collapse/index.zh-CN.md](https://github.com/thylsky/ant-design/blob/patch-1/components/collapse/index.zh-CN.md)

### DIFF
--- a/components/collapse/Collapse.tsx
+++ b/components/collapse/Collapse.tsx
@@ -9,8 +9,8 @@ import animation from '../_util/openAnimation';
 export type ExpandIconPosition = 'left' | 'right';
 
 export interface CollapseProps {
-  activeKey?: Array<string> | string;
-  defaultActiveKey?: Array<string>;
+  activeKey?: Array<string | number> | string | number;
+  defaultActiveKey?: Array<string | number> | string | number;
   /** 手风琴效果 */
   accordion?: boolean;
   destroyInactivePanel?: boolean;
--- a/components/collapse/CollapsePanel.tsx
+++ b/components/collapse/CollapsePanel.tsx
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import { ConfigConsumer, ConfigConsumerProps } from '../config-provider';
 
 export interface CollapsePanelProps {
-  key: string;
+  key: string | number;
   header: React.ReactNode;
   disabled?: boolean;
   className?: string;
--- a/components/collapse/index.en-US.md
+++ b/components/collapse/index.en-US.md
@@ -18,8 +18,8 @@ A content area which can be collapsed and expanded.
 
 | Property | Description | Type | Default | Version |
 | --- | --- | --- | --- | --- |
-| activeKey | Key of the active panel | string\[]\|string | No default value. In `accordion` mode, it's the key of the first panel. |  |
-| defaultActiveKey | Key of the initial active panel | string | - |  |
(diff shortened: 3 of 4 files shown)


dengfuping:
@thylsky Remember to rebase it.

thylsky:
> @thylsky Remember to rebase it.

Fixed. 👍

dengfuping:
@thylsky `activeKey` and `defaultActiveKey` should be same with `key`.
image

thylsky:
> @thylsky `activeKey` and `defaultActiveKey` should be same with `key`.
> image

Good catch, thanks! Resolved.

components/collapse/Collapse.tsx
```diff
@@ -9,8 +9,8 @@ import animation from '../_util/openAnimation';
 export type ExpandIconPosition = 'left' | 'right';
 
 export interface CollapseProps {
-  activeKey?: Array<string> | string;
-  defaultActiveKey?: Array<string>;
+  activeKey?: Array<string> | string | number;
+  defaultActiveKey?: Array<string | number>;
```
dengfuping: @thylsky Should be:
```
activeKey?: Array<string | number> | string | number;
defaultActiveKey?: Array<string | number> | string | number;
```

thylsky: Yes, it was kinda confusing to me, why the defaultActiveKey expects only array.

thylsky: But the documentation says that it doesn't expects an array...

dengfuping: @thylsky activeKey supports single value and array value both and `defaultActiveKey`'s type is as same as `activeKey`.

thylsky: @dengfuping Fixed. 👍I've also updated the documentation related to this.

components/collapse/CollapsePanel.tsx
```diff
@@ -4,7 +4,7 @@ import classNames from 'classnames';
 import { ConfigConsumer, ConfigConsumerProps } from '../config-provider';
 
 export interface CollapsePanelProps {
-  key: string;
+  key: string | number;
```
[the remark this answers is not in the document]
thylsky: I don't think it's necessary to add the activeKey and the defaultActiveKey here. It should be defined in the Collapse parent component.

dengfuping (approved):

PR #17614, opened 2019-07-12
feat: Add visible to filterDropdown function

sedx:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [x] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?

In some cases in my project I need to get actual visibility status of table's filterDropdown component.

There are now easy way to get it, because it's render when table is mount and not only when column's filter shown.

So I need to use something to like this (ugly and markup dependent  🤦‍♂️):
```
    const { current: someElement } = elementRef;
    if (someElement.parentNode.parentNode.parentNode.offsetHeight === 0) {
      return;
    }
    someElement.focus();
```
To trigger focus when filterDropdown shown.

### 💡 Solution

Currently `FilterMenu` contains current visibility status, just pass it to filterDropdown.

### 📝 Changelog

Add yet another attribute to call of filterDropdown, which contain result of `FilterMenu#getDropdownVisible()` call

Also update doc of filterDropdown prop. Now it contains info that filterDropdown can accept function (in previous doc revision was only ReactNode) and add link to ts defenition of arguments of this function.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Add visible attrbiute to argument of filterDropdown    |
| 🇨🇳 Chinese |           |

-----
[View rendered components/table/index.en-US.md](https://github.com/sedx/ant-design/blob/filterDropdown_visible_prop/components/table/index.en-US.md)
[View rendered components/table/index.zh-CN.md](https://github.com/sedx/ant-design/blob/filterDropdown_visible_prop/components/table/index.zh-CN.md)


afc163 (approved):

components/table/filterDropdown.tsx
```diff
@@ -266,6 +266,8 @@ class FilterMenu<T> extends React.Component<FilterMenuProps<T>, FilterMenuState<
         confirm: this.handleConfirm,
         clearFilters: this.handleClearFilters,
         filters: column.filters,
+        visible: this.getDropdownVisible(),
+        getPopupContainer: (triggerNode: HTMLElement) => triggerNode.parentNode as HTMLElement,
```
sakalauskas: This line seems to cause an error in some highly customized dropdown filters as `triggerNode.parentNode` is being `null` for some reason on initial component mount.

PR #17826, opened 2019-07-23
[Types] [Form] improve def setFieldsValue

pierre-H:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Improve setFieldsValue def |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/form/Form.tsx
+++ b/components/form/Form.tsx
@@ -141,7 +141,7 @@ export type WrappedFormUtils<V = any> = {
   /** 获取一个输入控件的值 */
   getFieldValue(fieldName: string): any;
   /** 设置一组输入控件的值 */
-  setFieldsValue(obj: Partial<V>, callback?: Function): void;
+  setFieldsValue(obj: Partial<V> | any, callback?: Function): void;
   /** 设置一组输入控件的值 */
   setFields(obj: Object): void;
   /** 校验并获取一组输入域的值与 Error */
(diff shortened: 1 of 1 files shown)


zombieJ:
This will break nest path.

pierre-H:
Now, nest path is not broken and we still have the autocompletion when using the function.
Thank you @zombieJ

components/form/Form.tsx
```diff
@@ -141,7 +141,7 @@ export type WrappedFormUtils<V = any> = {
   /** 获取一个输入控件的值 */
   getFieldValue(fieldName: string): any;
   /** 设置一组输入控件的值 */
-  setFieldsValue(obj: Object, callback?: Function): void;
+  setFieldsValue(obj: Partial<V> | any, callback?: Function): void;
```
zombieJ: `Partial<V> | any` is same as `any`...

pierre-H: Technically yes, but we win the auto completion in the code editor.

zombieJ: @ztplz , how do you think about this?

ztplz: A little strange, just for auto completion?

pierre-H: Because of the nest path, yes. Better autocompletion is one of the interest of TypeScript no ?

ztplz: @pierre-H  
image
`any` here doesn't seem to be necessary because it has a defaulted `any`.
image

PR #17910, opened 2019-07-26
Tabsbugfix

oldturkey:
if tabs type ="editable-card",if children has falsy value ,we will get error message,so we should fix it in tabs.
https://codesandbox.io/s/lively-glade-jmg5s here is error info

### DIFF
--- a/components/tabs/index.tsx
+++ b/components/tabs/index.tsx
@@ -132,6 +132,7 @@ export default class Tabs extends React.Component<TabsProps, any> {
       React.Children.forEach(
         children as React.ReactNode,
         (child: React.ReactElement<any>, index) => {
+        if(child){
           let closable = child.props.closable;
           closable = typeof closable === 'undefined' ? true : closable;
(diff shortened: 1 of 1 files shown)


oldturkey:
@afc163 上次说的pr~

zombieJ (commented):
Use isValidate instead.  And test case.

oldturkey:
> Use isValidate instead. And test case.

改动没有效果么？

oldturkey:
> Use isValidate instead. And test case.

inValidate？

oldturkey:
> Use isValidate instead. And test case.

i test it in this project demo，
That's all right.

oldturkey:
i test it in this project demo，
That's all right. @zombieJ

zombieJ:
Use `React.isValidElement(child)` instead of `if (child)` directly.

oldturkey:
> Use `React.isValidElement(child)` instead of `if (child)` directly.

ok，thank you，get new knowledge

PR #18345, opened 2019-08-19
fix select-tree component in Modal style。Description: When the select…

xyj404:
…-tree is in the modal component and there are too many drop-down options, when you pull down a distance and show the Modal,  the options is incomplete.

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [x] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
Adjust TreeSelect component style to make it work well when it in modal and pull down.

### 📝 Changelog
TreeSelect style adjust.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/tree-select/style/index.less
+++ b/components/tree-select/style/index.less
@@ -104,8 +104,9 @@
         &.@{select-tree-prefix-cls}-switcher_open,
         &.@{select-tree-prefix-cls}-switcher_close {
           .@{select-prefix-cls}-switcher-loading-icon {
-            position: absolute;
+            position: fixed;
             left: 0;
+            z-index: 1000;
             display: inline-block;
             width: 24px;
             height: 24px;


jianhunxia:
image
`:global{
  .ant-tree li.ant-tree-treenode-loading span.ant-tree-switcher.ant-tree-switcher_open .ant-tree-switcher-loading-icon,
  .ant-tree li.ant-tree-treenode-loading span.ant-tree-switcher.ant-tree-switcher_close .ant-tree-switcher-loading-icon {
    position: fixed!important;
    left: 0;
    z-index:1000!important;
    display: inline-block;
    width: 24px;
    height: 24px;
    color: #1890ff;
    font-size: 14px;
    -webkit-transform: none;
        -ms-transform: none;
            transform: none;
  }
}`
image

According to the above friend's code, the overwrite style is added, but it has no effect, I don't know why. I have repeated the bug, which seems to have nothing to do with this, but has to do with the fact that the top attribute of the package element in the drop-down menu has changed to 0px.

PR #18397, opened 2019-08-21
chore: Move openAnimation out of defaultProps

lukevella:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

No issue reported.

### 💡 Background and solution

The `defaultProps` of the `<Collapse />` included the prop `openAnimation` which is not defined in the prop types `CollapseProps`.  This can lead to ts errors for anyone wanting to extend this component since `<Collapse />` does not satisfy the type definition of `React.ComponentType<CollapseProps>` due to `defaultProps` not satisfying `Partial<CollapseProps>`.

To fix this, I moved  `openAnimation` outside of `defaultProps` and applied it directly to `<RcCollapse />` in the render function.

### 📝 Changelog

No breaking changes.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/collapse/Collapse.tsx
+++ b/components/collapse/Collapse.tsx
@@ -73,8 +73,8 @@ export default class Collapse extends React.Component<CollapseProps, any> {
     );
     return (
       <RcCollapse
-        {...this.props}
         openAnimation={{ ...animation, appear() {} }}
+        {...this.props}
         expandIcon={(panelProps: PanelProps) => this.renderExpandIcon(panelProps, prefixCls)}
         prefixCls={prefixCls}
         className={collapseClassName}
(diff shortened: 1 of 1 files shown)


components/collapse/Collapse.tsx
```diff
@@ -75,6 +74,7 @@ export default class Collapse extends React.Component<CollapseProps, any> {
     return (
       <RcCollapse
         {...this.props}
+        openAnimation={{ ...animation, appear() {} }}
```
yoyo837: It should be allowed to be override.

PR #18515, opened 2019-08-28
feat: improve prop def for tooltip

laysent:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

(none)

### 💡 Background and solution

Tooltip component does not require to pass either `title` or `overlay`. However, when not passing the any of them, the component itself does not work (not rendering tooltip).

The solution is to improve the type definition, so that at least one of `title` and `overlay` is provided.

Please notice that `React.ReactNode` includes falsy values such as `undefined`, `null` and `false`, thus following works:

```jsx
<Tooltip title={undefined} />
```

However, following should consider type error and TypeScript should warn about it:

```jsx
<Tooltip />
```

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  fix: ensure `title` or `overlay` is required in Tooltip props definition         |
| 🇨🇳 Chinese |  fix: 确保 Tooltip props 定义中，`title` 或 `overlay` 至少有一个是必填的         |

### DIFF
--- a/components/menu/MenuItem.tsx
+++ b/components/menu/MenuItem.tsx
@@ -42,11 +42,12 @@ export default class MenuItem extends React.Component<MenuItemProps> {
     return (
       <MenuContext.Consumer>
         {({ inlineCollapsed }: MenuContextProps) => {
-          const tooltipProps: TooltipProps = {};
+          const tooltipProps: TooltipProps = {
+            title: title || (level === 1 ? children : ''),
+          };
 
-          let titleNode = title || (level === 1 ? children : '');
           if (!siderCollapsed && !inlineCollapsed) {
-            titleNode = null;
+            tooltipProps.title = null;
             // Reset `visible` to fix control mode tooltip display not correct
             // ref: https://github.com/ant-design/ant-design/issues/16742
             tooltipProps.visible = false;
@@ -55,7 +56,6 @@ export default class MenuItem extends React.Component<MenuItemProps> {
           return (
             <Tooltip
               {...tooltipProps}
-              title={titleNode}
               placement="right"
               overlayClassName={`${rootPrefixCls}-inline-collapsed-tooltip`}
             >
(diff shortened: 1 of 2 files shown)


afc163:
The optional `title` prop could be useful, image this situation: You have a tooltip need to display depends on another state, with optional `title` we don't need to change JSX structure.

Instead of

```jsx
return this.state.showTooltip ? (
  <Tooltip tooltip="tooltip">
    <span>text</span>
  </Tooltip>
) : <span>text</span>;
```

We can do this way:

```jsx
return (
  <Tooltip tooltip={this.state.showTooltip ? 'tooltip' : undefined}>
    <span>text</span>
  </Tooltip>
);
```

laysent:
> The optional `title` prop could be useful, image this situation: You have a tooltip need to display depends on another state, with optional `title` we don't need to change JSX structure.
> 
> Instead of duplicating the trigger node:
> 
> ```js-jsx
> return this.state.showTooltip ? (
>   <Tooltip tooltip="tooltip">
>     <span>trigger</span>
>   </Tooltip>
> ) : <span>trigger</span>;
> ```
> 
> We can write in a simpler way:
> 
> ```js-jsx
> return (
>   <Tooltip tooltip={this.state.showTooltip ? 'tooltip' : undefined}>
>     <span>trigger</span>
>   </Tooltip>
> );
> ```

Hi, I believe this is a real-world case, and the modified version can support it, because `title` is having type `React.ReactNode`. According to the definition, `React.ReactNode` is defined as `ReactChild | ReactFragment | ReactPortal | boolean | null | undefined`.

Thus, the above example is allowed in new version, as `title` can be either `string` or `undefined`.

This modified version is only to prevent the case where user forget to set `title` at all (including not setting it as `undefined`), i.e. `<Tooltip />` is not allowed.

afc163:
Tested in my machine, it works like a charm!

image

Issue #5489, opened 2017-03-24
feat: add `innerRef`


benjycui:

benjycui:
Don't use ref with antd's component, it's unstable.

benjycui:
`innerRef` may have conflict with styled-component, we can use other name, such as: https://github.com/react-component/form/pull/87

benjycui:
See: https://reacttraining.com/react-router/web/api/withRouter/wrappedComponentRef-func

We even use the same prop name....

tkrotoff:
FYI
- material-ui has `inputRef` on all its components, see https://github.com/callemall/material-ui/blob/v1.0.0-beta.12/src/Input/Input.js#L255 and https://github.com/callemall/material-ui/search?utf8=%E2%9C%93&q=inputRef&type=
- reactstrap defines `getRef` (will be renamed `innerRef` in the future?), see https://github.com/reactstrap/reactstrap/issues/583
- React-Bootstrap defines `inputRef`, see https://github.com/react-bootstrap/react-bootstrap/blob/v0.31.3/src/Checkbox.js#L26

NugzarGaguliya:
What about inerRef? Is it in progress?

yesmeck:
Closing since it's inactive.

yesmeck:
https://github.com/reactjs/rfcs/pull/30

valerii-cognite:
What is the status for the issue?

Tymek:
@yesmeck could you reopen it? I'd like to contribute if it's possible to implement this.

yesmeck:
This issue doesn't have a clear goal, I see you've proposed a PR, we can discuss on the PR.

PR #18555, opened 2019-08-29
feat(Input): add input ref prop

Tymek:
#5489

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/ant-design/ant-design/issues/5489

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Add `inputRef` prop on `Input` component |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/input/Input.tsx
+++ b/components/input/Input.tsx
@@ -32,6 +32,7 @@ export interface InputProps
   onPressEnter?: React.KeyboardEventHandler<HTMLInputElement>;
   addonBefore?: React.ReactNode;
   addonAfter?: React.ReactNode;
+  inputRef?: (node: HTMLInputElement) => void;
   prefix?: React.ReactNode;
   suffix?: React.ReactNode;
   allowClear?: boolean;
@@ -67,6 +68,7 @@ class Input extends React.Component<InputProps, any> {
     onKeyUp: PropTypes.func,
     onFocus: PropTypes.func,
     onBlur: PropTypes.func,
+    inputRef: PropTypes.func,
     prefix: PropTypes.node,
     suffix: PropTypes.node,
     allowClear: PropTypes.bool,
@@ -144,6 +146,9 @@ class Input extends React.Component<InputProps, any> {
   }
 
   saveInput = (node: HTMLInputElement) => {
+    if (this.props.inputRef) {
+      this.props.inputRef(node);
+    }
     this.input = node;
   };
 


zombieJ:
Why need `inputRef`? Could you provides the usage case?

Tymek:
In an library that prefers uncontrolled inputs for performance reason. Use case, with an example here: https://github.com/react-hook-form/react-hook-form/issues/271#issuecomment-525804589

zombieJ:
antd v4 will provide high performance form management. Use ref to control input with perf consideration should be avoid since it will out of React control.

Tymek:
I agree it should be avoided, but not forbidden. Library I referenced earlier is useful. Using `className` to import input from framework is even less "React way". Opinionated nature of antd forms will force that.

zombieJ:
For Component level, export internal node should be avoid since it makes Component coupling and hard to refactor. In some component we provides api that can access node as event callback param or some subComponent className define to handle some edge case but its value is less than harm. For Component that not provides the api yet need to be careful. At least avoid to do that on current version since code may be refactor and internal node may change also.
In current form case, I don't think should provides `inputRef` since it already not controlled in React way, user can use `findDOMNode` to get input element.

Tymek:
`findDOMNode` is deprecated. Should I work on introducing this feature to new version, or is it going to be rejected?

zombieJ:
There have a [proposal](https://github.com/reactjs/rfcs/pull/97)  to replace the `findDOMNode`. Before it finalize, you can safe to use `findDOMNode`.

For this `inputRef` requirement, currently closed til if some other use case force need that.

ManAnRuck:
@Tymek @zombieJ i am also very interested to have some kind of ref to form elements (also for TextArea) 

i want to control my nice looking antd form also with react-hook-form

mgcrea:
I'm in the same boat, ref are an important part of the react API and I don't think a general-purpose UI library should obfuscate it. I don't see any downsides in exposing an `inputRef` prop, most other UI-library do this as well (eg. [material-ui](https://material-ui.com/api/input/)).

@zombieJ many advanced devs can't wait for v4/field-form to be production ready and want to use hooks today, not sure you should turn them away by blocking core react API usage of the lib.

WPaczula:
I'd love to use `inputRef`/`innerRef` to focus my input as a side effect. I really don't want to add technical debt by using `findDOMnode`.

PR #18578, opened 2019-08-30
chore: Col support key if user providered

### LINKED ISSUE #18408: Allow custom keys for Descriptions.Item instead of just indexing an array of children

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/antd-reproduction-template-7mtno)

### Steps to reproduce
Look at react component tree (in provided example but not in sandbox due to some sandbox limitations).
Escpecially look at Descriptions.Item keys in comparsion with List.Item keys

### What is expected?
Descriptions should use user-provided keys for Item-s

Descriptions enumerates children components from zero to length-1 and use that values as keys.

| Environment | Info |
|---|---|
| antd | 3.22.0 |
| React | react 16.8.6 |
| System | any |
| Browser | any |

---
Provided example is search results for multiple categories.
Each user query will return different number of categories (blog posts, news, events, regular website pages...)

While Descriptions component does not use provided keys for Items, all children will be re-rendered after each query/response change.

zombieJ:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [x] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

close #18408

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    Description.Item will reuse `key` in render if user provided.        |
| 🇨🇳 Chinese |    Description.Item 在渲染时会复用用户提供的 `key`。      |

### DIFF
--- a/components/descriptions/__tests__/__snapshots__/index.test.js.snap
+++ b/components/descriptions/__tests__/__snapshots__/index.test.js.snap
@@ -39,7 +39,7 @@ exports[`Descriptions Descriptions support colon 1`] = `
                 </DescriptionsItem>
               }
               colon={false}
-              key="label-0"
+              key="0"
               layout="horizontal"
               type="label"
             >
@@ -116,7 +116,7 @@ exports[`Descriptions Descriptions support style 1`] = `
                 </DescriptionsItem>
               }
               colon={true}
-              key="label-0"
+              key="0"
               layout="horizontal"
               type="label"
             >
@@ -183,7 +183,7 @@ exports[`Descriptions Descriptions.Item support className 1`] = `
                 </DescriptionsItem>
               }
               colon={true}
-              key="label-0"
+              key="0"
               layout="horizontal"
               type="label"
             >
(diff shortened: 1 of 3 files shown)


travikk:
This introduced a bug now. React keeps complaining about having children with the same key. Reason is, the same key is applied to label and content:

```key={colItem.key || idx}```

key used to be prefixed with the type of the cell, like so:

```key={`${type}-${idx}`}```

Can we change it to something like:

```
const cellKey = colItem.key || idx

return (
   <Col
      key={`${type}_${cellKey}`}
      ...
    />
);
```

zombieJ:
hi @travikk, could you help to create an issue for this?

Swind:
Hi @zombieJ,

I create a ticket #18633 about this issue.

Thanks

PR #18851, opened 2019-09-17
fix: correct registry-url

orzyyyy:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    fix action `publish` crashed caused by wrong `registry-url`      |
| 🇨🇳 Chinese |    修复 publish 时由于 `registry-url` 错误而无法发布的问题       |

### DIFF
--- a/.github/workflows/publish-to-github-package.yml
+++ b/.github/workflows/publish-to-github-package.yml
@@ -2,6 +2,8 @@ name: Publish to github-package
 on:
   release:
     types: [push]
+    branches:
+      - master
 
 jobs:
   publish:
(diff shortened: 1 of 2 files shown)


zombieJ:
This script is aim to publish to github package management system. Npm package still using manually publish since it has check file diff script which will human confirm if conflict happened.

orzyyyy:
I'll revert version check of npm. I'm going to try in another way.

orzyyyy:
I think it will works in next release.
Checks of npm would run behind of this action, so the version won't conflict with the one of github package, which is checked by `scripts/check-commit.js`.

PR #19199, opened 2019-10-13
feat: Add props to pass css classes to head and body of Card component

BenjaVR:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

Until now we could pass custom styles to the head and body of the `Card` component using the `headStyle` and `bodyStyle` props. However, when using e.g. css modules where we import css class names instead of inline css, we cannot easily style the body or header of this component.
I would like to add 2 new props (`headClassName` and `bodyClassName`) to be able to pass css classes to the head and the body of this component.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | You can now pass custom css classes to the body and head of the Card component. |
| 🇨🇳 Chinese |           |


BenjaVR:
@afc163 any more changes that need to be done?
Also, I'm not sure about the failing linting/tests, they seem to fail on parts I did not touch.

PR #19253, opened 2019-10-16
fix: Allow to use TableProps without Store

pierre-H:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

A little bit with #19249 but not exactly

### 💡 Background and solution

When you have a Custom Component which use the Table component and you want to pass all prop of the Custom Component to Table. The definition of the Custom Component must be without the store.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Add TablePropsWithoutStore interface |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/table/interface.tsx
+++ b/components/table/interface.tsx
@@ -161,7 +161,7 @@ export interface WithStore {
   setCheckboxPropsCache: (cache: CheckboxPropsCache) => void;
 }
 
-export interface TableProps<T> extends WithStore {
+export interface TablePropsWithoutStore<T> {
   prefixCls?: string;
   dropdownPrefixCls?: string;
   rowSelection?: TableRowSelection<T>;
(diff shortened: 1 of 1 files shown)


components/table/interface.tsx
```diff
@@ -219,6 +219,8 @@ export interface TableProps<T> extends WithStore {
   getPopupContainer?: (triggerNode: HTMLElement) => HTMLElement;
 }
 
+export type TableProps<T> = WithStore & TablePropsWithoutStore<T>;
```
JounQin: It will breaks `interface A<T> extends TableProps<T> {}`

`WithStore` should just be `Partial`

JounQin (changes requested):

PR #19409, opened 2019-10-24
fix: RangePicker added month

zotho:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/ant-design/ant-design/issues/15353

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/date-picker/RangePicker.tsx
+++ b/components/date-picker/RangePicker.tsx
@@ -31,8 +31,7 @@ function getShowDateFromValue(value: RangePickerValue, mode?: string | string[])
   if (mode && mode[0] === 'month') {
     return [start, end] as RangePickerValue;
   }
-  const newEnd = end && end.isSame(start, 'month') ? end.clone().add(1, 'month') : end;
-  return [start, newEnd] as RangePickerValue;
+  return [start, end] as RangePickerValue;
 }
 
 function pickerValueAdapter(


yoyo837:
我很奇怪，两年半以前就end月加1了，怎么都没问题吗！ #5663

yoyo837:
The `newEnd` code added 2.5 years ago. Hasn't anyone found something wrong for a long time?

zotho:
@yoyo837 This is not obvious behavior. Why add one month at all? About it in https://github.com/ant-design/ant-design/issues/15353

orzyyyy:
There is no problem for the unexpected behavior in #15353, the changes does fix this bug, but I'm affaid that you are working on the surface instead of getting to the point. It's hard to say whether the changes in this PR is correct, and it seemed like a legacy issue. 

Let's trace on the timeline:
- This is the [first issue](https://github.com/ant-design/ant-design/issues/4985)(Don't be afraid of Chinese, I will translate it roughly). 
  - It is very similar to this issue. The report guy found that, the date that has been selected before will automatically increase one. For example, start is 2015-01-01, end is 2016-02-02. Then pick year 2017 as an end time, the start time will be changed to 2017-01-01.
  - The maintainers at the time thought this was a problem in interactive design, related to [this](https://github.com/ant-design/ant-design/issues/4985#issuecomment-281550399).
  - Then [this PR](https://github.com/ant-design/ant-design/pull/5663) is the starting point of the problem.
- There is a [patch](https://github.com/ant-design/ant-design/commit/e3b252b5277fa37c9ce46523d5e78e5b98753970) based on previous commit.

What I think the way to completely solve this problem is to redesign the component, maybe 4.0 is working in process :)

orzyyyy (approved):
For 3.x, I prefer to "fix" as a patch.

I'm sorry that I can only point out the pain point but can not give a solution.

yoyo837:
Can we merge after your approved? @orzyyyy

orzyyyy:
I checked it again and find that these changes may not work.
test

zotho:
It is very strange why the changes caused such an effect.

orzyyyy:
Closing this for the changes not working as expect. Thanks for your effort, maybe you can have a try in  [calendar](https://github.com/react-component/calendar).

PR #19539, opened 2019-11-03
feat: Gets the XMLHttpRequest Object in Onchange function

hahmadia:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Possibly fix the following issue (but not 100%)
https://github.com/ant-design/ant-design/issues/19386

### 💡 Background and solution
I noticed that in https://github.com/react-component/upload/blob/master/src/AjaxUploader.jsx#L177 , we are passing the XMLHttpRequest object to the OnSuccess function but we are not retrieiving it and exposing it to the user (via the onChange function). This PR aims to fix that but allowing the user to have access to it through the file parameter on the `onChange` function.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

-----
[View rendered components/upload/index.en-US.md](https://github.com/hahmadia/ant-design/blob/feature/components/upload/index.en-US.md)

### DIFF
--- a/components/upload/Upload.tsx
+++ b/components/upload/Upload.tsx
@@ -103,7 +103,7 @@ class Upload extends React.Component<UploadProps, UploadState> {
     }
   };
 
-  onSuccess = (response: any, file: UploadFile) => {
+  onSuccess = (response: any, file: UploadFile, xhr: any) => {
     this.clearProgressTimer();
     try {
       if (typeof response === 'string') {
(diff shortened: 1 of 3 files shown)


yoyo837:
Does it works well when using window.Fetch?

afc163 (approved):

PR #19380, opened 2019-10-23
feat: added rtl direction to all of ant-design components

saeedrahimi:
### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Site / document update
- [x] Component style update
- [ ] TypeScript definition update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
close #4051

Previous pull request was on `feature` branch: #18006 
### 💡 Background and solution

As requirement of this [issue](https://github.com/ant-design/ant-design/issues/4051) components of ant-design doesn't support right to left direction, and those components are not useful in some languages like Persian(Farsi) or Arabic.

Solution: 
I started to develop Rtl style on ant-design and as @zombieJ  mentioned in issue i used ConfigProvider to set direction of component.

direction property added to ConfigProvider.
```
<ConfigProvider direction="rtl">
      <Cascader  options={options} placeholder="select an item"/>
</ConfigProvider>
```
in style files added prefixed rtl styles to change direction of displayed items
```
  .@{inputClass}-suffix {
    right: @input-padding-horizontal-base + 1px;

    .@{inputClass}-rtl & {
      right: auto;
      left: @input-padding-horizontal-base + 1px;
    }
  }
```
so when RTL'ed components wrapped by ConfigProvider with `direction='rtl'` property, all child components (also grid-system) will shown in right-to-left direction.
ant-rtl

## RTL Progress

- [x] Button
- [x] Icon
- [x] Typography
- [x] Grid
- [x] Layout
- [x] Affix
- [x] Breadcrumb
- [x] Dropdown
- [x] Menu
- [x] Pagination
- [x] PageHeader
- [x] Steps
- [x] AutoComplete
- [x] CheckBox
- [x] Cascader
- [x] DatePicker
- [x] Form
- [x] InputNumber
- [x] Input
- [x] Mention
- [x] Rate
- [x] Radio
- [x] Switch
- [x] Slider
- [x] Select
- [x] TreeSelect
- [x] Transfer
- [x] TimePicker
- [x] Upload
- [x] Avatar
- [x] Badge
- [x] Comment
- [x] Collapse
- [x] Carousel
- [x] Cad
- [x] Calendar
- [x] Description List
- [x] Empty
- [x] List
- [x] Popover
- [x] Statistic
- [x] Tree
- [x] Tooltip
- [x] Timeline
- [x] Tag
[...]

PR #19685, opened 2019-11-12
Adding both RTL and LTR support

sepiddd:
### 🤔 This is a 

- [x] New feature
- [ ] Site / document update
- [ ] Bug fix
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

1. We had an issue with RTL direction for ant design. After that, we decided to change the source code in order to have this feature.
2. I've created a mixins file(direction.less) that contains lots of helpful mixins which handle the directions with variables. Also, I've added 3 variables to enhance the workflow to default.less file.

`@dir: 'rtl';`
`@main-website-align: 'right';`
`@main-website-idir-align: 'left';`

If you change the values of the variables, you will have your indeed direction.

### 📝 Changelog

There is no need to change anything on the user side.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

- [x] Doc not needed
- [x] Demo is updated
- [x] TypeScript definition not needed
- [x] Changelog not needed


yoyo837:
Please check #19380, is it completely duplicated?

sepiddd:
Hi,

They are almost the same.

sepiddd:
> Please check #19380, is it completely duplicated?

Maybe they are almost the same, but I've handled it by `less` mixins, not by `ConfigProvider`.

afc163:
Duplicate of #19380

yoyo837:
Thx for your PR; it should be noted that the guy has been working hard for a long time with this in #19380.

sepiddd:
> Thx for your PR. it should be noted that the guy has been working hard for a long time with this in #19380.

I understand it and I didn't know that anyone was working on this issue. It was my challenge for about 3 years and I have fixed it temporarily by building and replaced `dist, lib` and `es` in `node_modules` folders every time to use them fine. Now, what should we do?

yoyo837:
You should let us know as soon as possible that you are working on it. Now, unless the Owner supports your `css` solution and abandons the `ConfigProvider` solution, we can only turn off your PR and sincerely thank you. Or the `ConfigProvider` solution fails and eventually returns to your solution in the end.

yoyo837:
Whether this PR is accepted or not, it should be change target branch to 4.0-prepare, just like #19380.

PR #19846, opened 2019-11-20
fix: Export tooltip helper interfaces

kachkaev:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

We have a library that re-exports a few ant-design components. Following an upgrade from antd 3.20.0 to 3.25.0, we noticed a typescript error in this code:

```ts
// CustomTooltip.tsx
import "antd/lib/tooltip/style"
import "./CustomTooltip.less"

import AntdTooltip, { TooltipPlacement, TooltipProps } from "antd/lib/tooltip"
import classnames from "classnames"
import React from "react"

const CustomTooltip: React.RefForwardingComponent<AntdTooltip, TooltipProps> = (
  { overlayClassName, ...rest },
  ref,
) => {
  return (
    <AntdTooltip
      ref={ref}
      overlayClassName={classnames("custom-tooltip-overlay", overlayClassName)}
      {...rest}
    />
  )
}
export default React.forwardRef(CustomTooltip)
// ⬆
// ❌ Default export of the module has or is using private name 'TooltipPropsWithTitle'.ts(4082)
// ❌ Default export of the module has or is using private name 'TooltipPropsWithOverlay'.ts(4082)
export { TooltipProps, TooltipPlacement }
``` 
This seems to happen only when we have this option in our `tsconfig.json`:

```json
{
  "compilerOptions": {
    "declaration": true,
  },
}
``` 

After adding `export` in front of `TooltipPropsWithTitle` and `TooltipPropsWithOverlay` things work perfectly!

Seems like the problem was introduced in https://github.com/ant-design/ant-design/pull/18515 /cc @laysent

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Export tooltop helper interfaces to fix an issue with generating TypeScript declarations         |
| 🇨🇳 Chinese |           |

       _need help with Chinese_

### DIFF
--- a/components/tooltip/index.tsx
+++ b/components/tooltip/index.tsx
@@ -65,12 +65,12 @@ export interface AbstractTooltipProps {
 
 export type RenderFunction = () => React.ReactNode;
 
-interface TooltipPropsWithOverlay extends AbstractTooltipProps {
+export interface TooltipPropsWithOverlay extends AbstractTooltipProps {
   title?: React.ReactNode | RenderFunction;
   overlay: React.ReactNode | RenderFunction;
 }
 
-interface TooltipPropsWithTitle extends AbstractTooltipProps {
+export interface TooltipPropsWithTitle extends AbstractTooltipProps {
   title: React.ReactNode | RenderFunction;
   overlay?: React.ReactNode | RenderFunction;
 }


yoyo837:
Can you solve your problem with type `TooltipProps`?

> `export declare type TooltipProps = TooltipPropsWithTitle | TooltipPropsWithOverlay;`

and /cc @laysent

laysent:
lgtm. All used interfaces should be exported, otherwise `--declaration` won't work. my bad.
(explanation of this behavior can be found here: https://github.com/microsoft/TypeScript/issues/6307#issuecomment-168321011)

PR #20517, opened 2019-12-29
fix(style): form  field error border style for select

jyash97:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [x] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#20482 
### 💡 Background and solution

Another border was getting added due to some internal style of select. I had to override it to fix the issue.

image

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/form/style/index.less
+++ b/components/form/style/index.less
@@ -295,6 +295,14 @@
       }
     }
 
+    // fixes https://github.com/ant-design/ant-design/issues/20482
+    .@{ant-prefix}-select {
+      &.@{ant-prefix}-select-single:not(.@{ant-prefix}-select-customize-input)
+        .@{ant-prefix}-select-selector {
+        border: 0;
+      }
+    }
+
     .@{ant-prefix}-select.@{ant-prefix}-select-auto-complete {
       .@{ant-prefix}-input:focus {
         border-color: @error-color;
--- a/components/form/style/index.less
+++ b/components/form/style/index.less
@@ -296,7 +296,7 @@
     }
 
     // fixes https://github.com/ant-design/ant-design/issues/20482
-    .@{ant-prefix}-select {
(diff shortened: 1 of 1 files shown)


components/form/style/index.less
```diff
@@ -295,6 +295,14 @@
       }
     }
 
+    // fixes https://github.com/ant-design/ant-design/issues/20482
+    .@{ant-prefix}-input-group-addon .@{ant-prefix}-select {
+      &.@{ant-prefix}-select-single:not(.@{ant-prefix}-select-customize-input)
+        .@{ant-prefix}-select-selector {
+        border: 0;
+      }
+    }
```
zombieJ: Remove this line: https://github.com/ant-design/ant-design/blob/7998a3312e780cec232f6b1d5ad389c6f9a63288/components/form/style/index.less#L305

jyash97: Hey @zombieJ , sorry I didn't understand.  Did you mean to remove the `&-selector` on `L305` and add the class name for `selector` like below ?
```
.@{ant-prefix}-input-group-addon .@{ant-prefix}-select .@{ant-prefix}-select-selector {
        border-color: transparent;
        box-shadow: none;
  }
```

zombieJ: I mean `L305` is a fix for this in v3 but not work in v4. If you add new patch to fix this, old one should be removed.

jyash97: Cool thanks 👍 Have pushed the fix

jyash97:
Thanks, @zombieJ for the quick reviews on my PR.

Issue #20999, opened 2020-01-19
Should hide dropdown list when disable Select component


Rey-Wang:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/practical-lake-1orh1?fontsize=14&hidenavigation=1&theme=dark)

### Steps to reproduce
1. focus on input
2. click Select component directly

### What is expected?
hide dropdown list when diable Select

### What is actually happening?
Observed that we can still see the dropdown list when diable Select

| Environment | Info |
|---|---|
| antd | 3.26.7 |
| React | 16.12 |
| System | codesandbox |
| Browser | chrome 79 |

Rey-Wang:
Or I think we should provide some interface can hide dropdown list directly

zombieJ:
It's a bug, `disabled` should close popup.

abdih:
I'm working on this.

morenyang:
Try to fix in this pr: https://github.com/react-component/select/pull/447

morenyang:
Seems the same issue in antd 4.x, I will take a look later

zombieJ:
fixed in v4. Reopen for v3.

PR #21016, opened 2020-01-19
fix: Ensure Select's Dropdown Closes When Disabled

abdih:
### 🤔 This is a ...

- [x] Bug fix
### 🔗 Related issue link
[#20999](https://github.com/abdih/ant-design/pull/new/20999-ha-hide-select-dropdown-when-disabled)
### 💡 Background and solution

Want Select's drop-down to close once its 'disabled' property is updated to true.
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Validate 'open' attribute's value between ant-design's Select and that of rc-select to ensure its closed if 'disabled' is true |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/select/index.tsx
+++ b/components/select/index.tsx
@@ -89,6 +89,8 @@ class Select<ValueType extends SelectValue = SelectValue> extends React.Componen
       listItemHeight = 32,
       getPopupContainer,
       dropdownClassName,
+      disabled,
+      open,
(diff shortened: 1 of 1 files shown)


zombieJ:
hi @abdih,
Should fix this logic in `rc-select` instead since open logic control is not on the antd level.

here is the `rc-select` repo: https://github.com/react-component/select

PR #21132, opened 2020-01-29
fix: Form.List resulting values

wanjas:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

`Form.List` provides an array of fields where the first element has `name` as numeric 0. If passed as it is to a `Form.Item` it leads to incorrect handling of the form values. 
You can reproduce it even with documentation example [Dynamic form item](https://next.ant.design/components/form/#components-form-demo-dynamic-form-item) (see values in the console).
After submitting anything first value will be `undefined` or with more add/remove manipulations it will end up being one of the deleted values.
 
I think that Form.List should avoid returning numbers as names and return strings instead (or if at some point for whatever reason `rc-field-form` will return array as a name pass it through).

### Alternatives
Maybe fix should be in `Form.Item` that threats 0 as being false. But I can't predict if a change from `isTrue` to `isDefined` will cause side-effects and I think name shouldn't be a number type anyway.

Another option is to change `rc-field-form`

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/form/FormList.tsx
+++ b/components/form/FormList.tsx
@@ -3,7 +3,7 @@ import { List } from 'rc-field-form';
 import warning from '../_util/warning';
 
 interface FieldData {
-  name: number;
+  name: string;
   key: number;
   fieldKey: number;
 }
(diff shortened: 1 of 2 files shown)


zombieJ:
`name` should support number type. It's a bug and should not cast since number means array in field-form.

PR #21250, opened 2020-02-05
fix: Minor fix on ValidationRule to match AsyncValidator

hansololai:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
close https://github.com/ant-design/ant-design/issues/21143

### 💡 Background and solution
In a scenario where one could store the `ValidationRule` and use in both in front end with Antd Form, and also could be used in the back end with `async-validator` alone. The rules could not be shared because the newest version `async-validator` v3's type definition`RuleItem`[link here](https://github.com/yiminghe/async-validator/blob/master/src/index.d.ts#L41) does not match the ValidationRule defined in Antd v3. 

After some investigation I found out  `rc-form` still depends on `async-validator`, which did not have the Typescript definitions (async validator added Typescript support later). So Antd could not directly reexport the `RuleItem`, but have to manually define `ValidationRule`. But it is not the same. 

Without introducing new features, I just corrected some fields on `ValidationRule` to be more consistent with the `RuleItem`. Even though the `RuleItem` is for `async-validator`v3, it was not changed very much after it was introduced, indicating the RuleItem still works for v1.11. 

To have `RuleItem` and `ValidationRule` to be fully assignable, the `RuleItem` may also need to be changed. 
```
import { ValidationRule } from 'antd/lib/form';
import { RuleItem, RuleType } from 'async-validator';
import { ReactNode } from 'react';


// ValidationRule.type is string, which is not assignable to RuleType
// Validation.enum is string|string[], not assignable to RuleItem
// These two are included in this PR
interface VR  extends Omit<ValidationRule,'type'|'enum'>{
  type: RuleType;
  enum?: (string|number|boolean)[];
}
// ValidationRule.message is ReactNode, RuleItem.message is string
// ValidationRule.pattern is RegExp, RuleItem.pattern is string. (actually async-validator handles pattern and string. 
// To fix these, async-validator needs a PR. 
interface RI extends Omit<RuleItem, 'pattern'|'message'>{
  pattern?: RegExp ;
  message?: ReactNode;
}

const a = {} as VR;
const b: RI = a;
```

### 📝 Changelog

Changed the ValidationRule `type` and `enum` field. 
1. The type is not a string enum. This should not affect you if you did not use types that is not handlable by async-validator.
2. BREAKING CHANGE The enum field now only support array of (string, number, boolean), it no longer support simple string. Which did not make sense anyways. 

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     ☑️      |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/form/Form.tsx
+++ b/components/form/Form.tsx
@@ -29,7 +29,7 @@ export interface FormCreateOption<T> {
 }
 
 const FormLayouts = tuple('horizontal', 'inline', 'vertical');
-export type FormLayout = (typeof FormLayouts)[number];
+export type FormLayout = typeof FormLayouts[number];
 
 export interface FormProps extends React.FormHTMLAttributes<HTMLFormElement> {
   layout?: FormLayout;
@@ -51,11 +51,28 @@ export interface FormProps extends React.FormHTMLAttributes<HTMLFormElement> {
   labelAlign?: FormLabelAlign;
 }
 
+export type ValidationRuleType =
+  | 'string'
+  | 'number'
+  | 'boolean'
+  | 'method'
+  | 'regexp'
+  | 'integer'
+  | 'float'
+  | 'array'
+  | 'object'
+  | 'enum'
+  | 'date'
+  | 'url'
+  | 'hex'
+  | 'email'
+  | 'any';
+
 export type ValidationRule = {
   /** validation error message */
   message?: React.ReactNode;
   /** built-in validation type, available options: https://github.com/yiminghe/async-validator#type */
-  type?: string;
+  type?: ValidationRuleType;
   /** indicates whether field is required */
   required?: boolean;
   /** treat required fields that only contain whitespace as errors */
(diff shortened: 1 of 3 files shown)


Kamahl19:
@afc163 @hansololai This change is actually not backward compatible. After upgrading to `3.26.9` my app wont build.

```
Type '({ required: boolean; message: string; } | { type: string; message: string; })[]' is not assignable to type 'ValidationRule[]'.
  Type '{ required: boolean; message: string; } | { type: string; message: string; }' is not assignable to type 'ValidationRule'.
    Type '{ type: string; message: string; }' is not assignable to type 'ValidationRule'.
      Types of property 'type' are incompatible.
        Type 'string' is not assignable to type '"string" | "number" | "boolean" | "object" | "url" | "email" | "date" | "method" | "regexp" | "integer" | "float" | "array" | "enum" | "hex" | "any" | undefined'.  TS2322
```

hansololai:
@Kamahl19 You are right. The type has been restricted to not just any `string`, but have to be one of the options in the union type. I did it to match `async-validator`. What do you think is the best way to handle this? @afc163 . If we want to only do any breaking changes on major release, like 4.0, then we can revert this PR. because 4.0 already handled this issue. 

I feel like it is so hard to update any Typescript definitions because any update on Types can be backward incompatible.

Issue #20301, opened 2019-12-17
Use classcat instead of classnames


Ambyjkl:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
Classcat does the same thing as classnames, but is about 5 times faster and also smaller
https://github.com/jorgebucaran/classcat

### What does the proposed API look like?
There is change, since this is just an internal tweak.

Ambyjkl:
I could make a quick PR switching everything over.

zombieJ:
only change `classnames` to other class concat library will not reduce the size since it deps component lib is still use `classnames` and make bundle have both `classnames` & other cls concat lib.

PR #21440, opened 2020-02-18
[WIP]feat: Replace classnames with cnbuilder

xobotyi:
### 🤔 This is a ...

- [x] Performance improvement

Replace `classnames` package with `cnbuilder`, it is faster at least 3-5 times (benchmarks can be found [here](https://github.com/xobotyi/cnbuilder/tree/master/benchmark)).

### 💡 Background and solution

Ofcourse it won't drastically improve the performance, but it is worth of changing because used in every render.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     Render performance improvement      |
| 🇨🇳 Chinese |           |


yoyo837:
https://github.com/ant-design/ant-design/issues/20301#issuecomment-569003081

xobotyi:
@yoyo837 the idea is to have more performant code.
I have opened performance [PR](https://github.com/JedWatson/classnames/pull/197) to classnames, but they're hang there for more than half a year and looks like wont be merged.

It's not a problem to update deps if this PR will be merged.

zombieJ:
Thanks for PR. Currently we do not have plan for replace `classnames`.
But I see your [PR](https://github.com/JedWatson/classnames/pull/197) for `classnames` which is a better way. Let's ping the maintainer again : )

hc-codersatlas:
If there's going to be a replacement, it would be good to compare it to other solutions e.g. [cslx](https://github.com/lukeed/clsx) too.

xobotyi:
@hc-codersatlas it is in benchmarks. cnbuilder is faster.

hc-codersatlas:
@xobotyi which benchmarks are you referring to? The 1 that was run with nodejs 13 as linkedin in cnbuilder? Please run an actual frontend benchmark that runs in a browser. There could be differences.

PR #21446, opened 2020-02-18
chore: bump rc-select

### LINKED ISSUE #21210: Select shouldn't always read container's offsetWidth in useLayoutEffect

This issue is similar to https://github.com/ant-design/ant-design/issues/21160

https://github.com/react-component/select/blob/master/src/generate.tsx#L898
this above line makes component render extremely slow when component is updated asynchronously 

### benchmark
example, super slow: https://rinick.github.io/antd-performance/select-performance.html
fixed version: https://rinick.github.io/antd-performance/select-measure-only-visible.html

source code of the benchmark: https://github.com/rinick/antd-performance/blob/master/select.jsx

You wont see this issue if everything is done in initial render, it only happens when children components are updated asynchronously, this usecase is common when page is driven by server data.

I will submit a PR shortly

rinick:
### 🤔 This is a ...

- [x] Others

### 🔗 Related issue link

https://github.com/ant-design/ant-design/issues/21210

### DIFF
--- a/package.json
+++ b/package.json
@@ -119,7 +119,7 @@
     "rc-progress": "~2.5.0",
     "rc-rate": "~2.5.1",
     "rc-resize-observer": "^0.1.0",
-    "rc-select": "~10.0.0-alpha.36",
+    "rc-select": "~10.0.0-rc.6",
     "rc-slider": "~9.2.1",
     "rc-steps": "~3.5.0",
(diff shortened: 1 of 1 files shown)


zombieJ:
`~10.0.0-alpha.36` do match rc version:
截屏2020-02-19上午9 14 58

This PR is OK but meanless.

Issue #21282, opened 2020-02-08
Multipe rows with gutter containing responsive width columns do not are not spaced properly


simonsankar:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/condescending-grass-rd5lw?fontsize=14&hidenavigation=1&theme=dark)

### Steps to reproduce
1. Create a row and set the gutter to an array - gutter={[16,16]}
2. Add columns with responsive widths - xxl={3} lg={4} md={6} sm={8} xs={8}
3. Repeat the Row containing the columns multiple times
4. Resize the browser to see how the columns of the second row begin after the last column of the first row

### What is expected?
It is expected that each row's columns start at the leftmost end

### What is actually happening?
After the first row's columns end, the second row's columns start below as expected on the Y-axis but are not to the leftmost end on the X-axis but after the first row's last column

| Environment | Info |
|---|---|
| antd | 3.26.7 |
| React | reactjs |
| System | Windows10 |
| Browser | chrome |

zombieJ:
Revert patch. Seems not find better solution.

zombieJ:
Seems user can rewrite Row display style. Patch fix makes more [style issue](#21403). You can use `type=flex` for fixing this.

And in v4, this is not a issue more since it default use flex layout.

https://codesandbox.io/s/intelligent-kapitsa-31ybi

felipeptcho:
The clearfix is not working because we have negative margin-top and margin-bottom on the grid rows when using vertical gutter. Instead of setting `clear: both` just for the `::after` pseudo-element, we should set it also for the `::before` one.

```diff
// mixins for clearfix
// ------------------------
.clearfix() {
  zoom: 1;
  &::before,
  &::after {
    display: table;
    content: '';
+   clear: both;
  }
-  &::after {
-    clear: both;
-  }
}
```

I tested this and it will not impact the other related ticket.

PR #21518, opened 2020-02-21
fix: Grid rows overlapping when using vertical gutter and having columns of different heights

felipeptcho:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#21282

### 💡 Background and solution

The clearfix was not enough when having columns of different heights because we have negative margin-top and margin-bottom on the grid rows when using vertical gutter. Instead of setting `clear: both` just for the `::after` pseudo-element, we should also set it for the `::before` one when we are dealing with an ant-row that is immediately preceded by another ant-row.

[Edit sharp-surf-lcsio](https://codesandbox.io/s/sharp-surf-lcsio?fontsize=14&hidenavigation=1&module=%2Findex.css&theme=dark)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix grid rows overlapping when using vertical gutter |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/grid/style/index.less
+++ b/components/grid/style/index.less
@@ -10,6 +10,10 @@
   box-sizing: border-box;
 }
 
+.@{ant-prefix}-row + .@{ant-prefix}-row::before {
+  clear: both;
+}
+
 .@{ant-prefix}-row-flex {
   display: flex;
   flex-flow: row wrap;


afc163:
This will break layout component, please check it: http://127.0.0.1:8001:components/layout/

zombieJ:
Patch snapshot:
截屏2020-02-22下午8 20 57

felipeptcho:
Thanks for the feedback. I realized that the clearfix is indeed correct. What we need to do instead is to apply this CSS rule specifically to the ant-row element only when it's preceded by another ant-row element.

```
.ant-row + .ant-row::before {
  clear: both;
}
```

Please, check again.

zombieJ (approved):

zombieJ:
Seems work as expect. Pls help to double confirm if I miss something. @afc163 @shaodahong @yoyo837

PR #21792, opened 2020-03-02
chore: removes erroniously present comment

dimitropoulos:
This comment was an accidental copy/pasta from https://reactjs.org/docs/error-boundaries.html.  Also, it is inaccurate since the user (i.e. AntD user) cannot actually render any fallback component.

### 🤔 This is a ...

- [x] Bug fix

### 🔗 Related issue link

https://reactjs.org/docs/error-boundaries.html

### 💡 Background and solution

removal of the comment


yoyo837:
ref: #19923

PR #22916, opened 2020-04-03
fix: Extract type of the element of a tuple without indexing.

nicu-chiciuc:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/microsoft/TypeScript/issues/13778

### 💡 Background and solution
This is a long running issue regarding array indexing in TS.
Basically, when performing indexing on an array TS will not show an error if the element is not present:
```TS
const arr: number[] = [];
// No errors
const num: number = arr[0];
```

One of the approaches in the thread linked above is to amend the definition of indexing of an element from:
```
  [n: number]: T;
```
to
```
  [n: number]: T | undefined;
```

The only downside is that this also restricts the indexing of array Types.
The solution is pretty straightforward, instead of:
```
export type OmitProp = typeof OmitProps[number];
```
we can use: 
```
export type OmitProp = ElementOf<typeof OmitProps>;
```

This will not cause any changes to existing users but will greatly improve the experience of those who use a stricter version of TS.

### 📝 Changelog

Users will not see any change but those who use the stricter versions of TS will be able to not use @ts-ignore or something similar

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     x      |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/_util/type.ts
+++ b/components/_util/type.ts
@@ -5,6 +5,7 @@ export const tuple = <T extends string[]>(...args: T) => args;
 export const tupleNum = <T extends number[]>(...args: T) => args;
 
 /**
+ * https://stackoverflow.com/a/59187769
  * Extract the type of an element of an array/tuple without performing indexing
  */
-export type ElementOf<T extends unknown[]> = T extends (infer R)[] ? R : never;
+export type ElementOf<T> = T extends (infer E)[] ? E : T extends readonly (infer E)[] ? E : never;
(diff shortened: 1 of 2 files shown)


nicu-chiciuc:
The last commit adds a more common version of `ElementOf<T>` that also supports `readonly` arrays just in case.
https://stackoverflow.com/a/59187769

nicu-chiciuc:
Screen Shot 2020-04-03 at 23 37 04
We're currently updating to version 4 and just this one error.

PR #22947, opened 2020-04-05
chore: Introduce customer jest match to support matchRenderedSnapshot

fa93hws:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
`render` and `matchSnapshot` are repeated too many times..
Maybe it can be simplified with a customer jest matcher so that
```
const wrapper = render(<Button>按钮</Button>);
```
becomes
```
expect(<Button>按钮</Button>).matchRenderedSnapshot();
```

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


package.json
```diff
@@ -65,7 +65,7 @@
     "lint:md": "remark components/",
     "lint:script": "eslint . --ext '.js,.jsx,.ts,.tsx'",
     "lint:style": "stylelint '{site,components}/**/*.less' --syntax less",
-    "lint:tsc": "npm run tsc",
+    "lint:tsc": "npm run tsc -- --noEmit",
```
fa93hws, line 68: Should typecheck emit the output `js{,x}` files?

PR #23452, opened 2020-04-21
fix: include tests in type check

fa93hws:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
Tests should be included in typecheck

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,7 +4,6 @@ components/**/*.jsx
 !components/*/demo/*
 !.*.js
 # Docs templates
-*.md
 site/theme/template/Color/ColorPicker.jsx
 site/theme/template/IconDisplay/*.js
 site/theme/template/IconDisplay/*.jsx
--- a/.eslintignore
+++ b/.eslintignore
@@ -4,6 +4,7 @@ components/**/*.jsx
 !components/*/demo/*
 !.*.js
 # Docs templates
+*.md
 site/theme/template/Color/ColorPicker.jsx
 site/theme/template/IconDisplay/*.js
 site/theme/template/IconDisplay/*.jsx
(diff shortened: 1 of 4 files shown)


.eslintignore
```diff
@@ -4,6 +4,7 @@ components/**/*.jsx
 !components/*/demo/*
 !.*.js
 # Docs templates
+*.md
```
afc163, line 7: Why ignore md from eslint?

fa93hws: Are we linting markdown file with eslint?
`components/table/index.en-US.md` are full of red lines..

If it's expected to lint markdown file with eslint, I will try to find some other workaround in another PR

afc163: `npm run lint:demo`

fa93hws: Cool, I will remove this ignore here.
And I do have some red lines in `components/table/demo/ajax.md` as well...
Will take a look when I have time.

.eslintrc.js
```diff
@@ -1,7 +1,3 @@
-const commonGlobals = {
-  gtag: true,
-};
-
```
fa93hws, line 4: Actually the `overrides` will handle it for usd

afc163:
image

fa93hws:
> image

Hmmm interesting.. I did add few lines so that the typescript is happy about the types and there was no test for this file before..
But after having a comprehensive read on `domHook.ts`, I have a feeling that it's not necessary at all... Let me have  a try on removing it..

PR #23543, opened 2020-04-24
chore: Deprecate AnchorDefaultProps

fa93hws:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
The type of `defaultProps` will be resolved automatically by `typescript`. There is no need to define our own type.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/anchor/Anchor.tsx
+++ b/components/anchor/Anchor.tsx
@@ -70,7 +70,7 @@ export interface AnchorState {
   activeLink: null | string;
 }
 
-// Deprecate
+/** @deprecated Use AnchorProps */
 export type AnchorDefaultProps = AnchorProps;
 
 export interface AntAnchor {
--- a/components/anchor/Anchor.tsx
+++ b/components/anchor/Anchor.tsx
@@ -51,9 +51,9 @@ export interface AnchorProps {
   children?: React.ReactNode;
   offsetTop?: number;
   bounds?: number;
-  affix?: boolean;
-  showInkInFixed?: boolean;
-  getContainer?: () => AnchorContainer;
+  affix: boolean;
+  showInkInFixed: boolean;
+  getContainer: () => AnchorContainer;
   /** Return customize highlight anchor */
   getCurrentAnchor?: () => string;
   onClick?: (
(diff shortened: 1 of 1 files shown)


components/anchor/Anchor.tsx
```diff
@@ -70,12 +70,8 @@ export interface AnchorState {
...
 }
 
-export interface AnchorDefaultProps extends AnchorProps {
-  prefixCls: string;
-  affix: boolean;
-  showInkInFixed: boolean;
-  getContainer: () => AnchorContainer;
-}
+// Deprecate
+export type AnchorDefaultProps = AnchorProps;
```
fa93hws, line 74: How to deprecate a type?

components/anchor/Anchor.tsx
```diff
@@ -51,9 +51,9 @@ export interface AnchorProps {
   children?: React.ReactNode;
   offsetTop?: number;
   bounds?: number;
-  affix?: boolean;
-  showInkInFixed?: boolean;
-  getContainer?: () => AnchorContainer;
+  affix: boolean;
+  showInkInFixed: boolean;
+  getContainer: () => AnchorContainer;
```
afc163, line 56: Don't remove `?` that make props required.

fa93hws: If you've defined the `defaultProps` in the class, typescript is clever enough now to mark them as optional when the component is used. 
image

components/anchor/Anchor.tsx
zombieJ: use jsdoc:
```tsx
/** @deprecated your deprecated reason here ... */
```

fa93hws: Updated

components/anchor/Anchor.tsx
zombieJ: We can remove `defaultProps` of Anchor. `?` is correct and should not be removed.

yoyo837:
Conflicted

components/anchor/Anchor.tsx
fa93hws: @zombieJ Do you mean that the exported type need to preserve the optional parameter for consumers?

afc163:
Close for no response, feel free to send us a new PR.

PR #24932, opened 2020-06-11
fix: Remove autofocus property

### LINKED ISSUE #24911: Icon 图表-代码演示-点击"展开全部代码"滚动条会回到顶部

### Reproduction link
[https://ant.design/components/icon-cn/](https://ant.design/components/icon-cn/)

### Steps to reproduce
ant.design官网 - 组件 - 图表 - 代码演示 - 点击"展开全部代码"

### What is expected?
点击"展开全部代码"滚动条不回到顶部

点击"展开全部代码"滚动条回到了顶部

| Environment | Info |
|---|---|
| antd | 4.3.3 |
| React | 16.9.0 |
| System | mac |
| Browser | 83.0.4103.97 |

---
搜索图标功能,IconDisplay/index.tsx 中  117行的autoFocus 属性导致的

nitinknolder:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |


xrkffgg:
has error

image

I try put `this.setState()` in `componentDidMount`, but the effect is not what I think, maybe my idea is wrong.

nitinknolder:
> has error
> 
> image
> 
> I try put `this.setState()` in `componentDidMount`, but the effect is not what I think, maybe my idea is wrong.

So what should i do should i remove this or directly assign value to it like **autofocus={false}**

xrkffgg:
Let's look at other people's opinions.

nitinknolder:
> Let's look at other people's opinions.

Ok

nitinknolder:
> > Let's look at other people's opinions.
> 
> Ok

I think my latest push will work please have a look

zombieJ:
You can call `focus` on `compomentDidMount`. Seems no need to set autoFocus.

nitinknolder:
> You can call `focus` on `compomentDidMount`. Seems no need to set autoFocus.

So like what exactly you want me to do here?

components/input/Input.tsx
```diff
@@ -144,6 +144,9 @@ class Input extends React.Component<InputProps, InputState> {
   }
 
   componentDidMount() {
+    this.setState({
+      focused: true,
```
zombieJ, line 148: `focused` is controlled by `onFocus` & `onBlur` event. Should not modify here.

nitinknolder: I mean I have to make a change only on the index.TSX file?

zombieJ: Yes. Just fix with the page code.

nitinknolder: I think the issue is somehow different basically it will initialize the page again onClick of "Expand All Code" This is the reason why it scrolls up the page.
So, What I suggest is to take one boolean flag and initialize it to true at the time of page initialization and set as false below. or you could have suggest a better solution.

zombieJ: The Input component works as expect. Should check the reason why page re-render when expanded.

zombieJ: icon index.md file use inject react which makes IconDisplay re-render.
`````
```__react
import IconDisplay from 'site/theme/template/IconDisplay';
ReactDOM.render(<IconDisplay />, mountNode);
```
`````

We may use sessionStorage to save the state that already rendered once.

nitinknolder: Not working with session storage I think we need to restrict the call on click of expand all button.

PR #24958, opened 2020-06-12
feat: Implement centered prop in Tabs

ashkan-pm:

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#24941

### 💡 Background and solution

According to #24941 feature request, Tabs component should accept a `centered` prop which, if true, will align the tabs to the center.
With this PR, the Tabs component will accept a boolean prop `centered`. A conditional class name is then added which will set `justifyContent` to `center` on the appropriate element.
Note that this will not happen if the element has the `ant-tabs-nav-wrap-ping-*` class. In this case the tabs are slidable due to overflow and adding `justifyContent` is unnecessary and breaks the UI and therefore it is avoided.
So <Tabs centered> will produce a result like this:
Screen Shot 2020-06-12 at 3 49 38 PM

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |Add centered prop to Tabs component|
| 🇨🇳 Chinese |           |

-----
[View rendered components/tabs/demo/centered.md](https://github.com/ashkan-pm/ant-design/blob/tabs-centered/components/tabs/demo/centered.md)
[View rendered components/tabs/index.en-US.md](https://github.com/ashkan-pm/ant-design/blob/tabs-centered/components/tabs/index.en-US.md)


afc163:
image

Just `New feature` would be OK.

yoyo837:
Does it support using with `tabPosition`?

components/tabs/__tests__/index.test.js
```diff
@@ -77,6 +77,19 @@ describe('Tabs', () => {
...
+  describe('centered', () => {
+    it('center tabs', () => {
+      const wrapper = render(
+        <Tabs centered>
+          <TabPane tab="foo" key="1">
+            foo
+          </TabPane>
+        </Tabs>,
+      );
+      expect(wrapper).toMatchSnapshot();
```
zombieJ, line 89: No need to add this test case since demo will auto compare snapshot.

components/tabs/demo/centered.md
```diff
@@ -0,0 +1,40 @@
+---
+order: 1
+title:
+  zh-CN: Centered
```
zombieJ, line 4: `Centered` -> `居中`

`order` move to after `disabled`

components/tabs/demo/centered.md
```diff
@@ -0,0 +1,40 @@
...
+  en-US: Centered
+---
+
+## zh-CN
+
+Centered tabs.
```
zombieJ, line 10: > 标签居中展示。

components/tabs/demo/centered.md
```diff
@@ -0,0 +1,40 @@
...
+## en-US
+
+Centered tabs.
+
+```jsx
+import { Tabs } from 'antd';
+
+const { TabPane } = Tabs;
+
+function callback(key) {
```
zombieJ, line 21: No need `callback` in this demo

components/tabs/index.en-US.md
```diff
@@ -28,6 +28,7 @@ Ant Design has 3 types of Tabs for different situations.
 | defaultActiveKey | Initial active TabPane's key, if `activeKey` is not set. | string | - |  |
 | hideAdd | Hide plus icon or not. Only works while `type="editable-card"` | boolean | `false` |  |
 | size | preset tab bar size | `large` \| `default` \| `small` | `default` |  |
+| centered | Centers tabs | boolean | false |  |
```
zombieJ, line 31: Should add version info of `4.4.0`

zombieJ: Also need add with Chinese document:
`Centers tabs` : `标签居中展示`

ashkan-pm:
> Does it support using with `tabPosition`?

Yeah as far as I have checked it works with all tab positions

yoyo837 (approved):

PR #24983, opened 2020-06-14
fix: allow selecting only value on double click on DescriptionsItem

### LINKED ISSUE #24977: Double-clicking the value of `Descriptions.Item` selects both label and value
Jun-14-2020 02-34-35

### Reproduction link
[https://ant.design/components/descriptions/](https://ant.design/components/descriptions/)

### Steps to reproduce
Double-Click the value of `Descriptions.Item`

### What is expected?
Only the value should be selected.

Both label and value are selected.

| Environment | Info |
|---|---|
| antd | 4.3.3 |
| React | - |
| System | macOS Mojave |
| Browser | Chrome 83.0.4103.61 (Official Build) (64-bit) |

This is not harmful at all, but a user normally would expect that only the value is selected.

harupy:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Fix #24977

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/descriptions/Cell.tsx
+++ b/components/descriptions/Cell.tsx
@@ -63,11 +63,7 @@ const Cell: React.FC<CellProps> = ({
           {label}
         </span>
       )}
-      {content && (
-        <span className={classNames(`${itemPrefixCls}-item-content`)}>
-          <div>{content}</div>
-        </span>
-      )}
+      {content && <span className={classNames(`${itemPrefixCls}-item-content`)}>{content}</span>}
     </Component>
   );
 };
--- a/components/descriptions/Cell.tsx
+++ b/components/descriptions/Cell.tsx
@@ -63,7 +63,12 @@ const Cell: React.FC<CellProps> = ({
           {label}
         </span>
       )}
-      {content && <span className={classNames(`${itemPrefixCls}-item-content`)}>{content}</span>}
+      {content && (
+        <span className={classNames(`${itemPrefixCls}-item-content`)}>
+          <i style={{ fontSize: 0 }}> </i>
+          {content}
+        </span>
+      )}
     </Component>
   );
 };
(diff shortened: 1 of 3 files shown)


Rustin-Liu:
The CI broken.

components/descriptions/Cell.tsx
```diff
@@ -63,7 +63,11 @@ const Cell: React.FC<CellProps> = ({
           {label}
         </span>
       )}
-      {content && <span className={classNames(`${itemPrefixCls}-item-content`)}>{content}</span>}
+      {content && (
+        <span className={classNames(`${itemPrefixCls}-item-content`)}>
+          <div>{content}</div>
```
Rustin-Liu, line 68: I don't think put a div into span is a good idea.

Rustin-Liu: Maybe we should consider other element?

harupy: @Rustin-Liu Thanks for the comment.

I agree. I found a related issue on SO:

https://stackoverflow.com/questions/43690276/text-selection-inside-inline-blocks-without-space-between

harupy: @Rustin-Liu What do you think?

Rustin-Liu: It's OK for me. Thanks!

afc163:
Could we try this?

```diff
.ant-descriptions-item>span {
- display: inline-block;
+ display: inline-flex;
+ align-items: baseline;
}
```

harupy:
@afc163 Got it.

harupy:
@afc163 Verified it works (tested on my local).

Jun-16-2020 14-43-53

harupy:
@Rustin-Liu @afc163 Thanks for the support!

PR #24222, opened 2020-05-17
fix: remvoe wave findDOMNode api

hengkx:
[[中文版模板 / Chinese template](https://github.com/ant-design/ant-design/blob/master/.github/PULL_REQUEST_TEMPLATE/pr_cn.md)]

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] TypeScript definition update
- [ ] Bundle size optimization
- [ ] Perfermance optimization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#22493

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    remvoe button,switch,tag deprecated api       |
| 🇨🇳 Chinese | 移除button,switch,tag中废弃的api          |

### ☑️ Self Check before Merge

⚠️ Please check all items below before review. ⚠️

- [x] Doc is updated/provided or not needed
- [x] Demo is updated/provided or not needed
- [x] TypeScript definition is updated/provided or not needed
- [x] Changelog is provided or not needed

PR #25184, opened 2020-06-23
fix: findDOMNode is deprecated in StrictMode warning when using Button component

MostafaBalata:
**💡 Background and solution

**What is actually happening?**
When using <Button> component a warning message will appear in the browser console "Warning: findDOMNode is deprecated in StrictMode. findDOMNode was passed an instance of Wave which is inside StrictMode. Instead, add a ref directly to the element you want to reference..."

The problem occurs because of Wave component is using a deprecated function `finDOMNode` if you are in StrictMode and using react `^16.13.1`

**What is the solution**
- Remove finDOMNode because of the deprecation.
- Create reference object from React.createRef
- Override the children props by passing the reference

Using <Button> results in "findDOMNode is deprecated in StrictMode" warning

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Perfermance optimization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
There's an open issue here: 
https://github.com/ant-design/ant-design/issues/22493

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     x      |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/_util/wave.tsx
+++ b/components/_util/wave.tsx
@@ -1,5 +1,4 @@
 import * as React from 'react';
-import { findDOMNode } from 'react-dom';
 import TransitionEvents from '@ant-design/css-animation/lib/Event';
 import raf from './raf';
 import { ConfigConsumer, ConfigConsumerProps, CSPConfig, ConfigContext } from '../config-provider';
@@ -22,8 +21,12 @@ function isNotGrey(color: string) {
   }
   return true;
 }
+interface WaveProps {
+  insertExtraNode?: boolean;
+  ref?: React.RefObject<HTMLElement>;
+}
 
-export default class Wave extends React.Component<{ insertExtraNode?: boolean }> {
+export default class Wave extends React.Component<WaveProps> {
   static contextType = ConfigContext;
 
   private instance?: {
@@ -44,8 +47,10 @@ export default class Wave extends React.Component<{ insertExtraNode?: boolean }>
 
   context: ConfigConsumerProps;
 
+  private nodeReference = React.createRef<HTMLElement>();
+
   componentDidMount() {
-    const node = findDOMNode(this) as HTMLElement;
+    const node = this.nodeReference.current as HTMLElement;
     if (!node || node.nodeType !== 1) {
       return;
     }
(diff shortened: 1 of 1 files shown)


MostafaBalata:
@zombieJ is this PR going to be in consideration soon?

yoyo837:
Maybe not only the buttons.

components/_util/wave.tsx
```diff
@@ -44,8 +47,10 @@ export default class Wave extends React.Component<{ insertExtraNode?: boolean }>
 
   context: ConfigConsumerProps;
 
+  private nodeReference = React.createRef<HTMLElement>();
+
   componentDidMount() {
-    const node = findDOMNode(this) as HTMLElement;
+    const node = this.nodeReference.current as HTMLElement;
```
zombieJ, line 53: This is not equal to `findDOMNode`. Wave children is not always dom node.

MostafaBalata: but it does the same effect, what is the benefits from findDOMNode, that we can't get after assigning the reference explicitly?

MostafaBalata: @zombieJ also the tests are passing, which means it wont harm any children that doesn't include dom nodes.
I would like to solve the problem completely, can you guid me with more details?

zombieJ: > but it does the same effect, what is the benefits from findDOMNode, that we can't get after assigning the reference explicitly?

`findDOMNode` always return dom node but `ref` is not.  It is not same effect.

You can consider to add prop in ConfigProvider to disable wave instead.

hengkx:
#24222

HuanxinHu:
I follow the document https://ant.design/docs/react/use-in-typescript, also have this warning in console, when there is only a button component.

hengkx:
ping

zombieJ:
Close since no more response

PR #25691, opened 2020-07-19
feat: textArea -  add word count ( same as antd mobile)  example: 20/500

### LINKED ISSUE #24940: Display character counter for input component optionally

I think having a character counter when there is a limit on the input field would be very important for UX and useful for projects using ant. It's a very common practice and material design promotes it too. Of course you can use the suffix or addon props with a bit of custom logic to achieve this, but that would create some problems. It would not be possible for the TextArea component, it wouldn't have the option of showing the counter outside of the input, and most importantly I find it sub-optimal to have to create custom components that wrap ant inputs in the projects that need this and have to use those instead.
I'd be more than happy to do this and create a PR for it if you think it's a good idea.

Perhaps we can have a showCounter boolean prop to the input which would show the number of characters entered if there is no maxLength prop, and show the ratio of characters used and the total character limit if a maxLength was specified. I attached a design for how the counter might look like.

Screen Shot 2020-06-12 at 6 20 04 AM

komagic:

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

在业务开发中，经常需要使用到textArea 的计数功能（如：20/100），之前发现antd-mobile有count字数功能，而antd-design没有，就考虑加上一个。
### 💡 Background and solution

截屏2020-07-19 上午11 54 18

close #24940

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

-----
[View rendered components/input/index.en-US.md](https://github.com/komagic/ant-design/blob/feature/components/input/index.en-US.md)
[View rendered components/input/index.zh-CN.md](https://github.com/komagic/ant-design/blob/feature/components/input/index.zh-CN.md)

### DIFF
--- a/components/comment/__tests__/__snapshots__/demo.test.js.snap
+++ b/components/comment/__tests__/__snapshots__/demo.test.js.snap
@@ -156,10 +156,15 @@ exports[`renders ./components/comment/demo/editor.md correctly 1`] = `
               <div
                 class="ant-form-item-control-input-content"
               >
-                <textarea
-                  class="ant-input"
-                  rows="4"
-                />
+                <div
+                  class="ant-input-control"
+                  value=""
+                >
+                  <textarea
+                    class="ant-input"
+                    rows="4"
+                  />
+                </div>
               </div>
             </div>
           </div>
--- a/components/input/TextArea.tsx
+++ b/components/input/TextArea.tsx
@@ -115,7 +115,7 @@ class TextArea extends React.Component<TextAreaProps, TextAreaState> {
       prefixCls: customizePrefixCls,
       bordered = true,
       showCount = false,
-      maxLength = showCount ? 500 : undefined,
+      maxLength = showCount ? 200 : undefined,
     } = this.props;
     const prefixCls = getPrefixCls('input', customizePrefixCls);
     return (
--- a/components/input/TextArea.tsx
+++ b/components/input/TextArea.tsx
@@ -9,7 +9,7 @@ import { fixControlledValue, resolveOnChange } from './Input';
 export interface TextAreaProps extends RcTextAreaProps {
   allowClear?: boolean;
   bordered?: boolean;
-  hasCount?: boolean;
+  showCount?: boolean;
 }
 
 export interface TextAreaState {
@@ -17,13 +17,14 @@ export interface TextAreaState {
 }
 
 function countSymbols(text = ''): number {
-  if (text == null || text == undefined) return 0;
+  if (text == null) return 0;
   const regexAstralSymbols = /[\uD800-\uDBFF][\uDC00-\uDFFF]|\n/g;
   return text.toString().replace(regexAstralSymbols, '_').length;
 }
 
 class TextArea extends React.Component<TextAreaProps, TextAreaState> {
   resizableTextArea: ResizableTextArea;
+
   clearableInput: ClearableLabeledInput;
 
   constructor(props: TextAreaProps) {
@@ -80,12 +81,12 @@ class TextArea extends React.Component<TextAreaProps, TextAreaState> {
   renderTextArea = (
     prefixCls: string,
     bordered: boolean,
-    hasCount: boolean,
+    showCount: boolean,
     value: string,
     maxLength: number | undefined,
   ) => {
     const characterLength = countSymbols(value);
-    const maxLen = hasCount ? maxLength : undefined;
+    const maxLen = showCount ? maxLength : undefined;
 
     return (
       <div className={`${prefixCls}-control`}>
@@ -99,7 +100,7 @@ class TextArea extends React.Component<TextAreaProps, TextAreaState> {
           onChange={this.handleChange}
           ref={this.saveTextArea}
         />
-        {hasCount && (
+        {showCount && (
           <span className={`${prefixCls}-control-count`}>
             <span>{characterLength}</span>/{maxLength}
           </span>
@@ -113,8 +114,8 @@ class TextArea extends React.Component<TextAreaProps, TextAreaState> {
     const {
       prefixCls: customizePrefixCls,
       bordered = true,
-      hasCount = false,
-      maxLength = hasCount ? 500 : undefined,
+      showCount = false,
+      maxLength = showCount ? 500 : undefined,
     } = this.props;
     const prefixCls = getPrefixCls('input', customizePrefixCls);
     return (
@@ -124,7 +125,7 @@ class TextArea extends React.Component<TextAreaProps, TextAreaState> {
         direction={direction}
         inputType="text"
         value={fixControlledValue(value)}
-        element={this.renderTextArea(prefixCls, bordered, hasCount, value, maxLength)}
+        element={this.renderTextArea(prefixCls, bordered, showCount, value, maxLength)}
         handleReset={this.handleReset}
         ref={this.saveClearableInput}
         triggerFocus={this.focus}
--- a/components/input/TextArea.tsx
+++ b/components/input/TextArea.tsx
@@ -16,14 +16,14 @@ export interface TextAreaState {
   value: any;
 }
 
-function countSymbols(text = '') {
+function countSymbols(text = ''): number {
+  if (text == null || text == undefined) return 0;
   const regexAstralSymbols = /[\uD800-\uDBFF][\uDC00-\uDFFF]|\n/g;
-  return text.replace(regexAstralSymbols, '_').length;
+  return text.toString().replace(regexAstralSymbols, '_').length;
 }
 
 class TextArea extends React.Component<TextAreaProps, TextAreaState> {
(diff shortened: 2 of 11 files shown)


fireairforce (commented):
加一下 test case 吧~

fireairforce:
lint也挂了，修改一下。

components/input/TextArea.tsx
```diff
@@ -71,23 +77,34 @@ class TextArea extends React.Component<TextAreaProps, TextAreaState> {
     resolveOnChange(this.resizableTextArea.textArea, e, this.props.onChange);
   };
 
-  renderTextArea = (prefixCls: string, bordered: boolean) => {
+  renderTextArea = (prefixCls: string, bordered: boolean,hasCount:boolean,value:string,maxLength:number) => {
+    let characterLength = countSymbols(value);
```
07akioni, line 81: make it const

components/input/index.en-US.md
```diff
@@ -49,6 +49,8 @@ The rest of the props of Input are exactly the same as the original [input](http
 | allowClear | If allow to remove input content with clear icon | boolean | false |  |
 | onResize | The callback function that is triggered when resize | function({ width, height }) | - |  |
 | bordered | Whether has border style | boolean | true | 4.5.0 |
+| hasCount | Whether has word count | boolean | false |  |
```
07akioni, line 52: please add version 4.5.0

components/input/index.zh-CN.md
```diff
@@ -50,6 +50,7 @@ Input 的其他属性和 React 自带的 [input](https://facebook.github.io/reac
 | allowClear | 可以点击清除图标删除内容 | boolean | false |  |
 | onResize | resize 回调 | function({ width, height }) | - |  |
 | bordered | 是否有边框 | boolean | true | 4.5.0 |
+| hasCount | 是否显示字数统计 | boolean | false |  |
```
07akioni, line 53: add version

07akioni (commented):
Fix lint & add test cases

komagic:
1. added two test count words and without hasCount 
2. fixed lint

components/input/TextArea.tsx
```diff
@@ -71,23 +77,45 @@ class TextArea extends React.Component<TextAreaProps, TextAreaState> {
     resolveOnChange(this.resizableTextArea.textArea, e, this.props.onChange);
   };
 
-  renderTextArea = (prefixCls: string, bordered: boolean) => {
+  renderTextArea = (
+    prefixCls: string,
+    bordered: boolean,
+    hasCount: boolean,
+    value: string,
+    maxLength: number|undefined,
```
xrkffgg, line 85:
```suggestion
    maxLength: number | undefined,
```

components/input/index.en-US.md
```diff
@@ -49,6 +49,8 @@ The rest of the props of Input are exactly the same as the original [input](http
 | allowClear | If allow to remove input content with clear icon | boolean | false |  |
 | onResize | The callback function that is triggered when resize | function({ width, height }) | - |  |
 | bordered | Whether has border style | boolean | true | 4.5.0 |
+| hasCount | Whether has word count | boolean | false | 4.5.0 |
+
```
xrkffgg, line 53:
```suggestion

```

components/input/TextArea.tsx
komagic: thanks

components/input/TextArea.tsx
```diff
@@ -9,15 +9,21 @@ import { fixControlledValue, resolveOnChange } from './Input';
...
   bordered?: boolean;
+  hasCount?: boolean;
 }
 
 export interface TextAreaState {
   value: any;
 }
 
+function countSymbols(text = ''): number {
+  if (text == null || text == undefined) return 0;
```
07akioni, line 20: `if (!text) return 0;` is enough, there is no way for `text === undefined`

components/input/TextArea.tsx
```diff
@@ -71,23 +77,45 @@ class TextArea extends React.Component<TextAreaProps, TextAreaState> {
...
   renderComponent = ({ getPrefixCls, direction }: ConfigConsumerProps) => {
-    const { value } = this.state;
-    const { prefixCls: customizePrefixCls, bordered = true } = this.props;
+    const { value = '' } = this.state;
+    const {
+      prefixCls: customizePrefixCls,
+      bordered = true,
+      hasCount = false,
+      maxLength = hasCount ? 500 : undefined,
+    } = this.props;
```
07akioni, line 118: why `maxLength` is not documented?

components/input/TextArea.tsx
komagic: thanks,what should I do, only keep typeof text ===’null‘ ?

components/input/TextArea.tsx
komagic: I thought maxLength has been documented by Input 
截屏2020-07-22 上午10 41 31

components/input/TextArea.tsx
```diff
@@ -9,15 +9,21 @@ import { fixControlledValue, resolveOnChange } from './Input';
 export interface TextAreaProps extends RcTextAreaProps {
   allowClear?: boolean;
   bordered?: boolean;
+  hasCount?: boolean;
```
shaodahong, line 12: Is better naming `showCount`?

komagic: The name is from antd-mobile, i prefer yours or enableCount

shaodahong: https://github.com/ant-design/ant-design/wiki/API-Naming-rules#props

@07akioni What do you think?

components/input/TextArea.tsx
07akioni: that is ok

components/input/TextArea.tsx
07akioni: `showCount`

components/input/TextArea.tsx
07akioni: You are adding feature to `Input.TextArea`, not `Input` itself. So adding it to API docs is in need.

components/input/index.zh-CN.md
```diff
@@ -50,7 +50,8 @@ Input 的其他属性和 React 自带的 [input](https://facebook.github.io/reac
 | allowClear | 可以点击清除图标删除内容 | boolean | false |  |
 | onResize | resize 回调 | function({ width, height }) | - |  |
 | bordered | 是否有边框 | boolean | true | 4.5.0 |
-| hasCount | 是否显示字数统计 | boolean | false | 4.5.0 |
+| showCount | 是否显示字数统计 | boolean | false | 4.5.0 |
+| maxLength | 最大长度 | number | - | 4.5.0 |
```
07akioni, line 54: how about default 500. Works when only `showCount` is `true`? The behavior doesn't follow docs now.

komagic: 大佬，1、我就是想在有showCount的时候，maxlength才显示，该如何优雅的code出来？ 2、测试总是报错误，有啥办法修改？

07akioni: 我觉得是这样，默认 maxLength 为空，showCount 只展示字数，然后如果加上一起展示，而且对非 showCount 状况也生效。

07akioni: 现在这样行为不一致

07akioni: 也不清晰

07akioni: 测试的话本地测一下或者点 details，看看自己哪个用例挂了，改掉

components/comment/__tests__/__snapshots__/demo.test.js.snap
```diff
@@ -156,10 +156,15 @@ exports[`renders ./components/comment/demo/editor.md correctly 1`] = `
               <div
                 class="ant-form-item-control-input-content"
               >
-                <textarea
-                  class="ant-input"
-                  rows="4"
-                />
+                <div
+                  class="ant-input-control"
+                  value=""
```
xiaohuoni, line 161: why value=""?

07akioni:
CI Failed

afc163:
Closing for no response, feel free to recreate another PR.

mxye:
So, do I understand it correctly that "showCount" will not be added as a feature to the generic Input, and if I would still like "showCount", a workaround would be needed, such as using "TextArea" but limiting the number of Rows to 1 to mimic the generic Input? -- The TextArea doesn't exactly look the same as generic Input, and in some cases generic Input would look better -- if only it has showCount...

PR #25827, opened 2020-07-27
fix(tabs): focus state color when disabled

hsuanxyz:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

https://codesandbox.io/s/serene-wiles-gtu01

### 💡 Background and solution

屏幕录制2020-07-27 下午2 45 35 mov

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   🐞 Fix tab  focus state color when disabled   |
| 🇨🇳 Chinese |  🐞 修复 tab 禁用时 focus 状态的颜色  |

### DIFF
--- a/components/tabs/style/index.less
+++ b/components/tabs/style/index.less
@@ -183,6 +183,14 @@
       cursor: not-allowed;
     }
 
+    &&-disabled &-btn,
+    &&-disabled &-remove {
+      &:focus,
+      &:active {
+        color: @disabled-color;
+      }
+    }
+
     & &-remove .@{iconfont-css-prefix} {
       margin: 0;
     }


components/tabs/style/index.less
```diff
@@ -183,6 +183,14 @@
       cursor: not-allowed;
```
MrHeer, line 183: why not write after this line?

hsuanxyz: sorry, I'm not familiar with less, do you have any good suggestions for me?

like this? I think the readability is not good.

```css
  &&-disabled {
      color: @disabled-color;
      cursor: not-allowed;
      
      .@{tab-prefix-cls}-tab-btn,
      .@{tab-prefix-cls}-tab--remove {
        &:focus,
        &:active {
          color: @disabled-color;
        }
      }
    }
```

xrkffgg: How about `pointer-events: none`?

07akioni (approved):

PR #25871, opened 2020-07-28
add onClick to DropDownProps

k-yle:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

`onClick` is missing from `<Dropdown />`'s props. It is usefull if you want to stop event propogation e.g. `<Dropdown onClick={e => e.stopPropagation()} />`

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Add `onClick` to `<Dropdown />` props |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/dropdown/dropdown.tsx
+++ b/components/dropdown/dropdown.tsx
@@ -39,7 +39,7 @@ export interface DropDownProps {
   trigger?: ('click' | 'hover' | 'contextMenu')[];
   overlay: React.ReactElement | OverlayFunc;
   onVisibleChange?: (visible: boolean) => void;
-  onClick?: (e: MouseEvent) => void;
+  onClick?: (e: MouseEvent<HTMLButtonElement, MouseEvent>) => void;
   visible?: boolean;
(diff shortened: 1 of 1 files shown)


07akioni:
have you try the prop? I haven't find related prop on rc-dropdown and rc-trigger

k-yle:
> have you try the prop? I haven't find related prop on rc-dropdown and rc-trigger

Yes, we use it already. It's just not included in the type definitions.

07akioni:
You shouldn't append event listener on Dropdown directly. You should append listener on its `children`. Appending listener directly to it is ambiguous (should content be listened to?) and may cause bugs.

k-yle:
Thanks, not sure why I didn't think of that! This PR can be closed then

PR #26479, opened 2020-08-29
fix: rlt style of columns

TrueMoein:
This fix should exist in rtl styles.

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [x] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Fix the rtl style of Col.         |
| 🇨🇳 Chinese |  修复 Col RTL 样式。          |

### DIFF
--- a/components/grid/style/rtl.less
+++ b/components/grid/style/rtl.less
@@ -10,7 +10,6 @@
 .@{ant-prefix}-col {
   &&-rtl {
     float: right;
-    max-width: 100%;
   }
 }
 
--- a/components/grid/style/rtl.less
+++ b/components/grid/style/rtl.less
@@ -10,7 +10,7 @@
 .@{ant-prefix}-col {
   &&-rtl {
     float: right;
-    width: 100%;
+    max-width: 100%;
   }
 }
 


xrkffgg:
Why did you remove it? Could you provide a example

TrueMoein:
> Why did you remove it? Could you provide a example

Hi, in the original style we have ( `max-width: 100%` ) and it's fine. but in RTL `width: 100%` added recently and it's broke the grid.

xrkffgg:
I add this because https://ant.design/components/list/#components-list-demo-resposive

If don't have `widht: 100%`, it will be

image

Maybe should add this only in List Component

TrueMoein:
Yes, thats right. I capture this columns screenshot with and without this style in the original document of antd right now.

chrome-capture

TrueMoein:
Dear @xrkffgg 

Maybe we don't need ` float: right;` style anymore. `flex` and main `direction: rlt;` handle it for us.
try it. then you can remove whole:

```
.@{ant-prefix}-col {
  &&-rtl {
    float: right;
    width: 100%;
  }
}
```

It's fixed the issue for List Components too.

Issue #26976, opened 2020-09-30
Typescript error for <Slider step={null} />


ranbena:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/antd-slider-step-null-l1h1i?file=/src/containers/App/index.tsx)

### Steps to reproduce
Create `<Slider>` with prop `step={null}` and typescript yields an error that "null" is not an acceptable value.

### What is expected?
That "null" value IS an acceptable value.
According to the docs it has a special meaning, and there's even an example in the AntD site with it - https://codesandbox.io/s/jxny9

### What is actually happening?
In https://github.com/ant-design/ant-design/pull/26884 (v4.6.6) the ts definition was wrongly changed from `step?: number | null` to `step?: number`.

| Environment | Info |
|---|---|
| antd | 4.6.6 |
| React | 16.13.1 |
| System | MacOS |
| Browser | Chrome |

---
🔔 Probably was undetected since the error appears for `--strictnullchecks`/`strict: true` in tsconfig.json.

ranbena:
@Kermit-Xuan plz revert this line

https://github.com/ant-design/ant-design/blob/216337219654315e6868317caafe9a750ce222ad/components/slider/index.tsx#L35

Kermit-Xuan:
you can use ‘null!’

ranbena:
I'm afraid that's not true for typescript configured with [strict null checks](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-0.html#--strictnullchecks).

[This example](https://codesandbox.io/s/antd-slider-step-null-l1h1i?file=/src/containers/App/index.tsx) clearly demonstrates the typescript error on the Slider component.

PR #26984, opened 2020-10-01
fix: Slider step prop not accepting null value

shangyilim:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

1. [#26976](https://github.com/ant-design/ant-design/issues/26976)

### 💡 Background and solution

The `step` prop for `Slider` will trigger typescript error: `Type null is not assignable to type 'number | undefined' `.  This is because the type definition does not allow `null` value, eventhough in the [documents it allowed as mentioned](https://ant.design/components/slider/):

```
the granularity the slider can step through values. Must greater than 0, and be divided by (max - min) . When marks no null, step can be null
```

This can be fixed by changing the definition of `step` to allow `null` value. 
Since underlying component used in `Slider` is `react-component/slider` its `step` props do not allow `null`, although in their documentation [it is allowed.](https://github.com/react-component/slider).
A quick fix will force the value `null!` within the library, while client apps will still be able to pass in `step={null}` without typescript errors. Until the underlying prop interface is resolved, this should a sufficient enough fix.

Updated typescript test to catch this error, and also added tests to ensure slider behaves as expected when null is passed. 
### 📝 Changelog

No breaking changes

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |Fixed `step` prop in `Slider` to accept null as a valid value in accordance to the documentation.|
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/slider/index.tsx
+++ b/components/slider/index.tsx
@@ -143,6 +143,7 @@ const Slider = React.forwardRef<unknown, SliderSingleProps | SliderRangeProps>(
       return (
         <RcRange
           {...(restProps as SliderRangeProps)}
+          step={restProps.step!}
           className={cls}
           ref={ref}
           handle={(info: HandleGeneratorInfo) =>
@@ -159,6 +160,7 @@ const Slider = React.forwardRef<unknown, SliderSingleProps | SliderRangeProps>(
     return (
       <RcSlider
         {...(restProps as SliderSingleProps)}
+        step={restProps.step!}
         className={cls}
         ref={ref}
         handle={(info: HandleGeneratorInfo) =>
--- a/components/slider/index.tsx
+++ b/components/slider/index.tsx
@@ -32,7 +32,7 @@ export interface SliderBaseProps {
   reverse?: boolean;
   min?: number;
   max?: number;
-  step?: number;
+  step?: null | number;
   marks?: SliderMarks;
   dots?: boolean;
   included?: boolean;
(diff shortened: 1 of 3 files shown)


components/slider/index.tsx
```diff
@@ -159,6 +160,7 @@ const Slider = React.forwardRef<unknown, SliderSingleProps | SliderRangeProps>(
     return (
       <RcSlider
         {...(restProps as SliderSingleProps)}
+        step={restProps.step!}
```
afc163, line 163: RcSlider 的定义也需要修正一下，避免用 `!`。

afc163:
image

Could you fill the changelog part?

components/slider/index.tsx
shangyilim: Unfortunately until RcSlider fixes their definition, there is nothing we can do on our end and to avoid typescript lint errors.
Furthermore I have added another test to safeguard againts `undefined`:
```
it('step should not crash with undefined value', () => {
    [undefined, null].forEach(value => {
      mount(<Slider step={value} tooltipVisible />);
    });
  });
```

PR #28090, opened 2020-11-30
fix: Make dontAnimate optional in CarouselRef interface

jarretmoses:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

v4.8.6 introduced a new CarouselRef type which is great but it did not correctly allow the `dontAnimate` argument of the `goTo` method to be optional which according to the source code it is. This just makes the argument optional.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/carousel/index.tsx
+++ b/components/carousel/index.tsx
@@ -23,7 +23,7 @@ export interface CarouselProps extends Omit<Settings, 'dots' | 'dotsClass'> {
 }
 
 export interface CarouselRef {
-  goTo: (slide: number, dontAnimate: boolean) => void;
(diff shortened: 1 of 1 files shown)


eugene-matvejev:
👍

PR #28092, opened 2020-11-30
fix: Collapse.Panel collapsible should be optional

sammarks:
It looks like this was broken in https://github.com/ant-design/ant-design/pull/27790. According to `rc-collapse`, the `collapsible` prop inside Collapse.Panel is optional, so it should be optional here as well if we're just forwarding props directly to the child component.

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

This was broken in https://github.com/ant-design/ant-design/pull/27790

### 💡 Background and solution

Upon using Collapse.Panel inside a TypeScript project and updating to the latest version of `antd`, Typescript now complains that `collapsible` is required in all usages of Collapse.Panel.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fixed Typescript definitions for Collapse.Panel collapsible property. |
| 🇨🇳 Chinese | 修复了Collapse.Panel可折叠属性的Typescript定义。 |

### DIFF
--- a/components/collapse/CollapsePanel.tsx
+++ b/components/collapse/CollapsePanel.tsx
@@ -18,7 +18,7 @@ export interface CollapsePanelProps {
   forceRender?: boolean;
   id?: string;
   extra?: React.ReactNode;
-  collapsible: CollapsibleType;
+  collapsible?: CollapsibleType;
 }
 
 const CollapsePanel: React.FC<CollapsePanelProps> = props => {


yoyo837 (approved):

eugene-matvejev:
👍

PR #28730, opened 2021-01-06
fix: unnecessary stylelint plugin

akijoey:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [x] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

Since [`stylelint-config-rational-order`](https://github.com/constverum/stylelint-config-rational-order) adds `stylelint-order` to plugins,
we don't have to do this when extending `stylelint-config-rational-order`.

image

It still work.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/.stylelintrc.json
+++ b/.stylelintrc.json
@@ -4,7 +4,7 @@
     "stylelint-config-rational-order",
     "stylelint-config-prettier"
   ],
-  "plugins": ["stylelint-order", "stylelint-declaration-block-no-ignored-properties"],
+  "plugins": ["stylelint-declaration-block-no-ignored-properties"],
   "rules": {
     "comment-empty-line-before": null,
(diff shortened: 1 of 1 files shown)


yoyo837:
Need remove it from the package.json file else.

AkiJoey:
Plugin `stylelint-declaration-block-no-ignored-properties` is never used.
Should I remove it together?

AkiJoey:
Plugin `stylelint-declaration-block-no-ignored-properties` is never used.
Should I remove it together?

PR #28792, opened 2021-01-09
fix: add memorization for ConfigProvider (#27617)

### LINKED ISSUE #27617: ConfigProvider can cause all antd components to re-render even its props are not changed

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/focused-knuth-o9wh6?file=/src/App.js)

### Steps to reproduce
1. Open [sandbox's full window mode](https://o9wh6.csb.app/)
2. Open react devtool profiler
3. Start profiling, click the first button, stop profiling, see flamegraph
4. Start profiling again, click the second button, stop profiling, compare flamegraph to the previous one

### What is expected?
Both flamegraphs should be identical since we have memoed `Text` component to prevent it from render.

On the first run, when `ConfigProvider` is inside updated component, even `Text` is not rendered, `Tooltip` under it is rendered, but on the second run all components are not rendered.

| Environment | Info |
|---|---|
| antd | 4.8.0 |
| React | 17.0.0 |
| System | macOS |
| Browser | Chrome 86 |

---
This performance issue can be prevented by putting `ConfigProvider` on top of components tree, however I believe many developers are not aware of this, and a complex page can contain hundreds of antd components, by each of them rendering for 1ms, a big and blocking render occurs.

It seems `ConfigProvider` lacks some memoization of its context value, maybe we can improve it.

zxc0328:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

resolve #27617

### 💡 Background and solution

The issue mentioned in https://github.com/ant-design/ant-design/issues/27617 will cause noticeable performance problem in large and complex web apps (In my case, a ConfigProvider inside root redux container case the whole App's ant components to re-render, which wastes hundreds of milliseconds for each render).

image

*notice all the discrete renders in the bottom level, the parent is not re-rendered. It's antd component's render caused by Context value change*

The problem can be fixed by placing ConfigProvider outside the redux container(or other root state container). But the placement of ConfigProvider is up to the user. So antd should fix this.

The fix keeps ref for the context value rendered last time. When ConfigProvider renders, it checks whether context value changed or not. Because the props passed in ConfigProvider can be modified by both immutable or mutable method, the equality checking is performed by deep comparison. And the memoed context value is deep cloned for equality checking.

The decision to support both immutable and mutable way of props change keeps the behavior of this component unchanged. Otherwise, the user changes props of ConfigProvider in a mutable way will encouter bugs when upgraded.


components/config-provider/index.tsx
```diff
@@ -114,9 +120,8 @@ const ConfigProvider: React.FC<ConfigProviderProps> & {
       dropdownMatchSelectWidth,
     } = props;
 
-    const config: ConfigConsumerProps = {
+    let config: ConfigConsumerProps = {
```
zombieJ, line 123: Use `useMemo` to create cache config object instead.

components/_util/hooks/useDeepMemo.ts
```diff
@@ -0,0 +1,30 @@
+import * as React from 'react';
```
zombieJ, line 1: It's no different with prev code. Why not use `React.useMemo` and realize memo self? ConfigProvider only need to care the passed prop is same ref and should not compare in deep. It will case additional perf issue since developer may wrap many ConfigProvider in real code. Shallow equal is good enough:
```tsx
// No additional render
const pageHeader  = { ghost: true };
<ConfigProvider pageHeader={pageHeader} />

// Additional render
<ConfigProvider pageHeader={{ ghost: true }} />
```

zxc0328: It would be nice to just use useMemo and shallow compare. But it would introduce a breaking change. Consider the following situation（[one of the test case](https://github.com/ant-design/ant-design/pull/28792/files#diff-90fc6ceffa60c11a3b5f758ae2514534599c065943ad3507bd61d2110c8fadd4R58)）:

```jsx
const App = () => {
      const [, forceRender] = React.useReducer(v => v + 1, 1);
      const pageHeader = React.useRef({ ghost: false });
      return (
        <ConfigProvider pageHeader={pageHeader.current}>
          <button
            type="button"
            onClick={() => {
              pageHeader.current.ghost = true;
              forceRender();
            }}
            className="mutable"
          >
            Change Direction
          </button>
          <MemoedSibling />
        </ConfigProvider>
      );
    };
```
If user change the props mutablely. In the current antd version, this change will be delivered to the Consumer. Because ConfigProvider generate new object **every single render**. But if we use useMemo, this kind of mutable change will not trigger Consumer's re-render.

We may have users change props in mutable way. So theoretically, the shallow equal implementation would introduce a bug .

So if you thought it is a bad prictice(it is) and should not be supported, I will use the shallow equal implementation.

zombieJ: antd v4 only promise immutable data for correct render behavior.

https://ant.design/docs/react/faq#Why-does-modifying-props-in-mutable-way-not-trigger-a-component-update

components/config-provider/index.tsx
```diff
@@ -68,19 +70,31 @@ export interface ConfigProviderProps {
...
-        rtl: props.direction === 'rtl',
-      });
-    }
-  }, [props.direction]);
+interface ProviderChildrenProps extends ConfigProviderProps {
+  parentContext: ConfigConsumerProps;
+  legacyLocale: Locale;
+}
+
+const ProviderChildren: React.FC<ProviderChildrenProps> = props => {
```
zombieJ, line 78: Why create another FC to handle this?

components/config-provider/index.tsx
```diff
@@ -95,37 +109,22 @@ const ConfigProvider: React.FC<ConfigProviderProps> & {
...
-      input,
-      locale,
-      pageHeader,
-      componentSize,
-      direction,
-      space,
-      virtual,
-      dropdownMatchSelectWidth,
-    } = props;
+  const getPrefixCls = React.useMemo(() => getPrefixClsWrapper(parentContext), [
```
zombieJ, line 112: `useCallback` instead.

components/config-provider/index.tsx
```diff
@@ -38,6 +35,11 @@ export const configConsumerProps = [
   'autoInsertSpaceInButton',
   'locale',
   'pageHeader',
+  'form',
+  'virtual',
+  'input',
+  'dropdownMatchSelectWidth',
+  'space',
```
zombieJ, line 42: `configConsumerProps` is legacy usage since hooks not ready when coding this. May not extend this list since it may be removed in future.

components/config-provider/index.tsx
```diff
@@ -150,42 +149,73 @@ const ConfigProvider: React.FC<ConfigProviderProps> & {
...
-      validateMessages = { ...validateMessages, ...form.validateMessages };
-    }
+  const config = getConfig();
 
-    if (Object.keys(validateMessages).length > 0) {
-      childNode = <RcFormProvider validateMessages={validateMessages}>{children}</RcFormProvider>;
-    }
+  // https://github.com/ant-design/ant-design/issues/27617
+  const memoedConfig = React.useMemo(
+    () => getConfig(),
```
zombieJ, line 159: Not need to call `getConfig` again since it already exist in `config`.

components/config-provider/index.tsx
zxc0328: Current implementation use a fuction to render children, where parent context's config is passed in.

Since we can't use Hooks inside a nested function. It has to be another component.

components/config-provider/index.tsx
zxc0328: useMemo's deps has to be a const size array, so we need a const array for all the ConsumerProps keys. If `configConsumerProps` will be removed, should I create a new const array for useMemo's deps?

zombieJ: Got it. you can use `rc-util/lib/hooks/useMemo` which provides additional compare function of shallow equal.

components/config-provider/index.tsx
```diff
@@ -68,64 +66,57 @@ export interface ConfigProviderProps {
...
+
+      return suffixCls ? `${mergedPrefixCls}-${suffixCls}` : mergedPrefixCls;
+    },
+    [parentContext.getPrefixCls],
+  );
 
-    const config: ConfigConsumerProps = {
-      ...context,
-      getPrefixCls: getPrefixClsWrapper(context),
+  const getConfig = (): ConfigConsumerProps => {
```
zombieJ, line 108: Now we can save this function declare. Get the config object directly. : )

components/config-provider/__tests__/memo.test.js
```diff
@@ -0,0 +1,81 @@
...
+    const [count, setCount] = React.useState(0);
+    React.useEffect(() => setCount(c => c + 1), [config]);
+    return <div id="counter">{count}</div>;
+  };
+
+  const Sibling = () => (
+    <ConfigConsumer>{config => <RenderCounter config={config} />}</ConfigConsumer>
+  );
+
+  it('should not generate new context config when render', () => {
```
zombieJ, line 18: ConfigConsumer is internal component which may remove if some refactor. Test should focus on the final behavior. You can ref this to create a test case instead: https://codesandbox.io/s/focused-knuth-o9wh6?file=/src/App.js:476-789

zombieJ: To be clear. Just remove `ConfigConsumer` in test case : )

PR #28865, opened 2021-01-14
fix: dont use document.body as slider's popup container

rinick:
when getPopupContainer  is not defined, it should rely on rc-tooltip's default behavior instead of forcing it to use document.body

### 🤔 This is a ...

- [x] Bug fix

### 🔗 Related issue link

https://github.com/ant-design/ant-design/issues/28435

### 💡 Background and solution

The Slider component shows tooltip in document.body by default. That means when slider is in a different document, like popup window or iframe, tooltip will be shown in the wrong location.

Solution:
Since this is already handled properly by rc-trigger and rc-tooltip, we only need to make sure slider doesn't override the default getPopupContainer  behavior

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Slider should not use document.body as default tooltip container, it should rely on rc-tooltip's default behavior  |
| 🇨🇳 Chinese |  Slider不应使用document.body作为默认的tooltip容器，应该留空让rc-tooltip默认机制决定  |

### DIFF
--- a/components/slider/index.tsx
+++ b/components/slider/index.tsx
@@ -116,7 +116,7 @@ const Slider = React.forwardRef<unknown, SliderSingleProps | SliderRangeProps>(
           transitionName="zoom-down"
           key={index}
           overlayClassName={`${prefixCls}-tooltip`}
-          getPopupContainer={getTooltipPopupContainer || getPopupContainer || (() => document.body)}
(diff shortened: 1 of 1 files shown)


afc163:
图片

rinick:
@afc163  updated

Issue #29235, opened 2021-02-04
Table pagination overrides setting top and bottom position to 'none'


zachfedor:
- [X] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://ant.design/components/table/#components-table-demo-dynamic-settings](https://ant.design/components/table/#components-table-demo-dynamic-settings)

### Steps to reproduce
Pagination config option for the Table component allows setting position for top and bottom pagination component like so:

```<Table pagination={{ position: ['topLeft', 'bottomLeft']}} />```

This position array is painfully confusing. It can contain up to two strings prefixed with `top` or `bottom`, placing up to two pagination components for the top and bottom according to the first two valid positions it sees.

In the documentation, it offers the choice to turn off pagination completely by replacing the config object with `false`. But it also shows the option of setting these positions to `'none'`, which works for one of the positions but is disregarded when set to _both_.

### What is expected?
```<Table pagination={{ position: ['none', 'none']}} />```

Should disable pagination component.

### What is actually happening?
Pagination component is displayed at the default position of bottom right of the table.

| Environment | Info |
|---|---|
| antd | 4.12.2 |
| React | 17.0.1 |
| System | macOS 10.15.7 |
| Browser | Chrome 88.0 / Firefox 85.0 |

mumiao:
i mention pr

PR #29251, opened 2021-02-05
fix: no default pagination if pagination is set with position array

headwindz:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

[Table pagination overrides setting top and bottom position to 'none'](https://github.com/ant-design/ant-design/issues/29235)

### 💡 Background and solution

When `pagination` is set with position array，a default pagination will be rendered to bottom right if none of the positions matches either 'top' or 'bottom'. Therefore, `{ pagination: ['none', 'none'] }` will result in a default pagination

Solution: Don't render default pagination when `pagination` is set with position array. Render a top pagination if one of the positions matches 'top'. The same logic applies for 'bottom' pagination.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Don't render default pagination when pagination is set with position array, use exact position match instead   |
| 🇨🇳 Chinese |  当使用位置信息作为页码设置时，如果没有匹配到位置信息，不要渲染默认页码  |

### DIFF
--- a/components/badge/style/index.less
+++ b/components/badge/style/index.less
@@ -190,12 +190,12 @@
     display: inline-block;
     height: @badge-height;
     transition: all @animation-duration-slow @ease-in-out;
-    .safari-fix-motion;
+    .safari-fix-motion();
 
     > p.@{number-prefix-cls}-only-unit {
       height: @badge-height;
       margin: 0;
-      .safari-fix-motion;
+      .safari-fix-motion();
     }
   }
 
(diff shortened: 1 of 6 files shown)


n0ruSh:
duplicate with https://github.com/ant-design/ant-design/pull/29256

PR #29615, opened 2021-03-04
feat: add onCancel and onEnd option for editable

### LINKED ISSUE #29371: Typography editable on cancel event

Typography editable has two events: `onStart`, `onChange`, but in my situation, I need an `onCancel`/`onEnd` event too.

This is so I can keep track of whether the element is being edited or not.
In my specific use case, I am replacing `\n` with `<br />` **only** when the text is **not** being edited. And when it **is** being edited, I will not replace the `\n` characters.

An example of editable paragraph:
```js
<Typography.Paragraph
  editable={{
    onStart={() => setIsEditing(true)}
    onEnd={() => setIsEditing(false)} // <- This is the new prop
    onChange={() => setIsEditing(false)}
  }}
>
```

With `onEnd` property will fire when the user ends the editing of the field. It could also be `onCancel` which would only fire when they cancel editing.

jueinin:

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

close https://github.com/ant-design/ant-design/issues/29371
### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Typography `editable` 新增 `onCancel` 和 `onEnd` 回调。 |
| 🇨🇳 Chinese | Typography `editable` supports `onCancel` and `onEnd`.  |

### DIFF
--- a/components/typography/Base.tsx
+++ b/components/typography/Base.tsx
@@ -37,6 +37,8 @@ interface EditConfig {
   tooltip?: boolean | React.ReactNode;
   onStart?: () => void;
   onChange?: (value: string) => void;
+  onCancel?: () => void;
+  onEnd?: () => void;
   maxLength?: number;
   autoSize?: boolean | AutoSizeType;
 }
@@ -210,6 +212,7 @@ class Base extends React.Component<InternalBlockProps, BaseState> {
   };
 
   onEditCancel = () => {
+    this.getEditable().onCancel?.();
     this.triggerEdit(false);
   };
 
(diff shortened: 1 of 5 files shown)


yoyo837:
请提交到feature分支

jueinin:
好的 不好意思没注意

afc163 (approved):

PR #29832, opened 2021-03-19
enhancement: (Upload) stopPropagation on file drop

jojonarte:
Call e.stopPropagation on file drop to support nested drag and drop upload use-cases

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [x] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

1. I was facing a use-case problem on my project, where I need nested upload drop targets, I couldn't find a way to make them work as the onDrop can't be overwritten from the client. 
2. The scenario that happens is if I drop a file into the child Upload target the event bubbles and the parent Upload target also receives the upload drop event, which results in double upload with unintended upload consequences.
3.  I could get the event and listen to the event via `addEventListener` but it's not enough to stop the event bubbling.
4. I came up with this solution to stop the propagation on the library's drop event itself directly.

### 📝 Changelog

Jut added a call to e.stopPropagation() on `onFileDrop` callback in `Upload`.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Prevent drop event bubbling on drop event         |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/upload/Upload.tsx
+++ b/components/upload/Upload.tsx
@@ -266,6 +266,7 @@ const InternalUpload: React.ForwardRefRenderFunction<unknown, UploadProps> = (pr
   };
 
   const onFileDrop = (e: React.DragEvent<HTMLDivElement>) => {
+    e.stopPropagation();
     setDragState(e.type);
   };
 


fgblomqvist:
This should have been put behind a prop (e.g. `propagateEvents` with a default of true) instead of changing the default case. This is a breaking change that made it into a minor version. Our code broke because of this and I'm certain we're not alone 😕 . This should either be reverted or fixed with a new PR that puts this behind a prop. @afc163 @jojonarte

ebonow:
+1 to the comments. I am looking for a way to notify the user if any of the dropped files were not an accepted file type. While it would be nice if rc-upload had a native method to return rejected files which could be exposed through this component, I think it is also useful to have a mechanism to run code on the `onDrop` event which now runs in isolation since this change.

I will submit a PR with this feature with the hopes that we can expose this functionality for users as an opt-in feature that shouldn't break existing users.

PR #29916, opened 2021-03-26
feat: allow data and aria props to input, button, checkbox, notification

emmajam:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

Allow data-* and aria-* props to be passed through to the Input, Button, Checkbox and Notification components, to allow for e2e UI testing e.g. cypress or testcafe in the proper way.

### 💡 Background and solution

Currently Antd components don't all accept the data-* or aria-* props, which makes it very difficult to create e2e tests properly, coinciding with good practices. As it means they must be tied to JS and CSS selectors, rather than to data-test-ids for example.

This takes advantage of an already existing function for the Form component and extends it to the Input, Button, Checkbox and Notification. 

### 📝 Changelog

🆕 Input supports `data-*` and `aria-*` props
🆕 Checkbox supports `data-*` and `aria-*` props
🆕 Notification accepts `dataTestId` prop

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  🆕 Input supports `data-*` and `aria-*` props 🆕 Checkbox supports `data-*` and `aria-*` props 🆕 Notification accepts `dataTestId` prop          |
| 🇨🇳 Chinese |  🆕 Input supports `data-*` and `aria-*` props 🆕 Checkbox supports `data-*` and `aria-*` props 🆕 Notification accepts `dataTestId` prop          |

### DIFF
--- a/components/button/button.tsx
+++ b/components/button/button.tsx
@@ -11,7 +11,6 @@ import devWarning from '../_util/devWarning';
 import SizeContext, { SizeType } from '../config-provider/SizeContext';
 import LoadingIcon from './LoadingIcon';
 import { cloneElement } from '../_util/reactNode';
-import getDataOrAriaProps from '../_util/getDataOrAriaProps';
 
 const rxTwoCNChar = /^[\u4e00-\u9fa5]{2}$/;
 const isTwoCNChar = rxTwoCNChar.test.bind(rxTwoCNChar);
@@ -270,7 +269,6 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
       </a>
     );
   }
-  const dataOrAriaProps = getDataOrAriaProps(props);
 
   const buttonNode = (
     <button
@@ -279,7 +277,6 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
       className={classes}
       onClick={handleClick}
       ref={buttonRef}
-      {...dataOrAriaProps}
     >
       {iconNode}
       {kids}
--- a/components/button/button.tsx
+++ b/components/button/button.tsx
@@ -11,6 +11,7 @@ import devWarning from '../_util/devWarning';
 import SizeContext, { SizeType } from '../config-provider/SizeContext';
 import LoadingIcon from './LoadingIcon';
 import { cloneElement } from '../_util/reactNode';
+import getDataOrAriaProps from '../_util/getDataOrAriaProps';
 
 const rxTwoCNChar = /^[\u4e00-\u9fa5]{2}$/;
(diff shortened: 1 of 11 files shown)


components/button/button.tsx
```diff
@@ -269,6 +270,7 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
       </a>
     );
   }
+  const dataOrAriaProps = getDataOrAriaProps(props);
```
zombieJ, line 273: This code is useless. `rest` already pass the props: https://codesandbox.io/s/anniuleixing-antd4141-forked-wj54d?file=/index.js

zombieJ (changes requested):

components/checkbox/__tests__/__snapshots__/checkbox.test.js.snap
```diff
@@ -17,3 +17,26 @@ exports[`Checkbox rtl render component should be rendered correctly in RTL direc
...
+  class="ant-checkbox-wrapper"
+  data-text="123"
+>
+  <span
+    class="ant-checkbox"
+  >
+    <input
+      aria-hidden="true"
+      class="ant-checkbox-input"
+      data-text="123"
```
zombieJ, line 33: Seems internal input already get the `data-*`, parent `label` should not duplicated of this props.

components/input/Input.tsx
```diff
@@ -293,6 +295,7 @@ class Input extends React.Component<InputProps, InputState> {
           },
         )}
         ref={this.saveInput}
+        {...dataOrAriaProps}
```
zombieJ, line 298: Same as Button. Input already pass the `otherProps` in. No need to get additional logic of this.

components/notification/index.tsx
```diff
@@ -221,7 +223,7 @@ function getRCNoticeProps(args: ArgsProps, prefixCls: string) {
 
   return {
     content: (
-      <div className={iconNode ? `${prefixCls}-with-icon` : ''} role="alert">
+      <div id="notification" data-test-id={dataTestId} className={iconNode ? `${prefixCls}-with-icon` : ''} role="alert">
```
zombieJ, line 226: `dataTestId` is too specific. Please consider provides another prop instead. Or just?
```
notification.open({
      message: <div data-test-id="test">Some notification</div>,
    });
```

PR #29944, opened 2021-03-29
fix: Not try to modify lock file

### LINKED ISSUE #29233: Upload - uid is set even if provided - issue when fileList is immutable from ReduxToolkit selector | Cannot assign to read only property 'uid' of object '#<Object>'

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/antd-reproduction-template-forked-8juv9)

### Steps to reproduce
Check reproduction link - there is my explanation of the problem.
The problem exists because I am using ReduxToolkit and getting attachements from the store (I am setting uid property) but this object is frozen (I want only show attached attachements (read-only))

### What is expected?
There is react effect to set uid on each file if not provided

```file.uid = (_a = file.uid) !== null && _a !== void 0 ? _a : "__AUTO__".concat(timestamp, "_").concat(index, "__");```

should be sth like this

```
if (!file.uid)
   file.uid = "__AUTO__".concat(timestamp, "_").concat(index, "__");
```

Cannot assign to read only property 'uid' of object '#<Object>'

| Environment | Info |
|---|---|
| antd | 4.12.2 |
| React | 16.3.1 |
| System | Windows |
| Browser | Chrome/Edge chromium/Firefox |

### PR DESCRIPTION
**Pull request by zombieJ**

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

resolve #29233

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    Fix Upload not crash when passed `fileList` is frozen object.      |
| 🇨🇳 Chinese |     修复 Upload 在 `fileList` 为冻结对象时不再崩溃。      |

### DIFF
--- a/components/upload/Upload.tsx
+++ b/components/upload/Upload.tsx
@@ -77,7 +77,7 @@ const InternalUpload: React.ForwardRefRenderFunction<unknown, UploadProps> = (pr
     const timestamp = Date.now();
 
     (fileList || []).forEach((file, index) => {
-      if (!file.uid) {
+      if (!file.uid && !Object.isFrozen(file)) {
         file.uid = `__AUTO__${timestamp}_${index}__`;
       }
     });
@@ -254,7 +254,7 @@ const InternalUpload: React.ForwardRefRenderFunction<unknown, UploadProps> = (pr
         currentFile = { ...file, status: 'removed' };
         mergedFileList?.forEach(item => {
           const matchKey = currentFile.uid !== undefined ? 'uid' : 'name';
-          if (item[matchKey] === currentFile[matchKey]) {
+          if (item[matchKey] === currentFile[matchKey] && !Object.isFrozen(item)) {
             item.status = 'removed';
           }
         });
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: components/upload/Upload.tsx [hunk 1]
```diff
@@ -77,7 +77,7 @@ const InternalUpload: React.ForwardRefRenderFunction<unknown, UploadProps> = (pr
     const timestamp = Date.now();
 
     (fileList || []).forEach((file, index) => {
-      if (!file.uid) {
+      if (!file.uid && !Object.isFrozen(file)) {
         file.uid = `__AUTO__${timestamp}_${index}__`;
```
yoyo837, line 81: 是不是价格warning好一些？提示尝试初始化uid失败，因为啥啥啥的。

PR #30111, opened 2021-04-12
style(Button): lint

liuzi6612:
Condition 'loading.delay' is always true at this point because it is already checked at line 176.

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [x] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution
Condition 'loading.delay' is always true at this point because it is already checked at line 176.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   Condition 'loading.delay' is always true at this point because it is already checked at line 176.        |
| 🇨🇳 Chinese |   此时条件'loading.delay'始终为true，因为在第176行已对其进行了检查         |

### DIFF
--- a/components/button/button.tsx
+++ b/components/button/button.tsx
@@ -174,7 +174,7 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
   // =============== Update Loading ===============
   let loadingOrDelay: Loading;
   if (typeof loading === 'object' && loading.delay) {
-    loadingOrDelay = loading.delay || true;
+    loadingOrDelay = true;
   } else {
     loadingOrDelay = !!loading;
   }


components/button/button.tsx
```diff
@@ -174,7 +174,7 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
   // =============== Update Loading ===============
   let loadingOrDelay: Loading;
   if (typeof loading === 'object' && loading.delay) {
-    loadingOrDelay = loading.delay || true;
```
xrkffgg, line 177: 如果为数字，会优先使用数字

PR #30179, opened 2021-04-16
fix(FormItem): add name as default variables.label value

### LINKED ISSUE #30146: When I add the locale property to the ConfigProvider,  Form.item triggers the rule check, validation message contains undefined

### Reproduction link
[Edit on CodeSandbox](https://codesandbox.io/s/cold-wave-2p77h?file=/index.js)

### Steps to reproduce
 add the locale property to the ConfigProvider

### What is expected?
When Form.item triggers rule verification, name props is not undefined

When Form.item triggers rule verification, name props is  undefined

| Environment | Info |
|---|---|
| antd | 4.15.1 |
| React | ^17.0.2 |
| System | manjaro 20 |
| Browser | chrome 89.0.4389.128 |

jameslahm:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
Fix #30146. When `label` is not given, `${label}` will be replaced by `undefined`. I think we could add `name` as a default `variables.label` value. 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | add `name` as a default label value in `validationMessages`. |
| 🇨🇳 Chinese | 增加`name`作为验证消息`label`的默认值|

### DIFF
--- a/components/form/FormItem.tsx
+++ b/components/form/FormItem.tsx
@@ -265,6 +265,8 @@ function FormItem<Values = any>(props: FormItemProps<Values>): React.ReactElemen
   let variables: Record<string, string> = {};
   if (typeof label === 'string') {
     variables.label = label;
+  } else if (name) {
+    variables.label = String(name);
   }
   if (messageVariables) {
     variables = { ...variables, ...messageVariables };
--- a/components/form/__tests__/index.test.js
+++ b/components/form/__tests__/index.test.js
@@ -564,6 +564,23 @@ describe('Form', () => {
     expect(wrapper.find('.ant-form-item-explain').first().text()).toEqual('Bamboo is good!');
   });
 
+  it('`name` support template when label is not provided', async () => {
+    const wrapper = mount(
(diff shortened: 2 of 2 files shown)


components/form/FormItem.tsx
```diff
@@ -265,6 +265,8 @@ function FormItem<Values = any>(props: FormItemProps<Values>): React.ReactElemen
   let variables: Record<string, string> = {};
   if (typeof label === 'string') {
     variables.label = label;
+  } else if (name) {
+    variables.label = String(name);
```
zombieJ, line 269: Could you help to add a test case about this?

jameslahm: Ok, I will add the test case as soon as possible ~

PR #31343, opened 2021-07-10
Update type for Menu onOpenChange in docs.

kku1993:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [X] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

N/A

### 💡 Background and solution

Looks like `react-component` has updated the argument type to `React.Key[]` instead of `string[]` [1]

[1] https://github.com/react-component/menu/blob/571c52f35a9d0ba582f58c26e06f44c0619d4253/src/Menu.tsx#L112

### 📝 Changelog

N/A

### DIFF
--- a/components/menu/index.en-US.md
+++ b/components/menu/index.en-US.md
@@ -53,7 +53,7 @@ More layouts with navigation: [Layout](/components/layout).
 | triggerSubMenuAction | Which action can trigger submenu open/close | `hover` \| `click` | `hover` |  |
 | onClick | Called when a menu item is clicked | function({ item, key, keyPath, domEvent }) | - |  |
 | onDeselect | Called when a menu item is deselected (multiple mode only) | function({ item, key, keyPath, selectedKeys, domEvent }) | - |  |
-| onOpenChange | Called when sub-menus are opened or closed | function(openKeys: string\[]) | - |  |
+| onOpenChange | Called when sub-menus are opened or closed | function(openKeys: React.Key\[]) | - |  |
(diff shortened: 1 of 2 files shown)


afc163:
I suggest using string instead of React.Key which same as the type of defaultOpenKeys and openKeys.

图片

kku1993:
@afc163 sorry for being unclear, but the change is needed because the underlying [react-component library](https://github.com/react-component/menu/blob/571c52f35a9d0ba582f58c26e06f44c0619d4253/src/Menu.tsx#L112) has changed the type to `React.Key`, so I'm afraid it's not up to the docs to decide that the type should be `string`.

afc163:
`string` is safe to use.

kku1993:
@afc163 unfortunately typescript will not typecheck your code if you use `onOpenChange = (s: string[]) => ...`

afc163:
@kku1993 Could you provide a reproduce codesandbox?

kku1993:
> @kku1993 Could you provide a reproduce codesandbox?

Of course: https://codesandbox.io/s/magical-sunset-3w5ps?file=/src/App.tsx

You need to wait for typescript to load the type definitions. Once it does, you should see:

Screenshot from 2021-07-12 13-47-03

PR #31479, opened 2021-07-21
Fix Input component styles when using prefixCls

spawnia:

### 🤔 This is a ...

- [x] Bug fix

### 🔗 Related issue link

I am trying to use two versions of antd in a project simultanously.
To separate the styles, I am trying to use `prefixCls` for one of them.

Most components do work correctly, receiving an appropriate class name, and the less styles
match the class name. However the `Input` component does not seem to have any matching styles for the prefixed class name.

### 💡 Background and solution

Changed according to https://github.com/ant-design/ant-design/blob/master/components/input-number/style/index.less

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Input component styles when using prefixCls |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/input/style/index.less
+++ b/components/input/style/index.less
@@ -4,15 +4,17 @@
 @import './affix';
 @import './allow-clear';
 
+@input-prefix-cls: ~'@{ant-prefix}-input';
+
 // Input styles
-.@{ant-prefix}-input {
+.@{input-prefix-cls} {
   .reset-component();
   .input();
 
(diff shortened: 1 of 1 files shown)


spawnia:
I just tried this locally, with this change the class name is now generated correctly: 
image

Previously, the HTML was correct but the CSS did not contain styles for `.mll-ant-input`.

PR #31661, opened 2021-08-04
chore(all): delete useless Omit define

Dreamerryao:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [x] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 💡 Background and solution

TS includes `Omit` type
TS内置了Omit类型

### DIFF
--- a/components/_util/type.ts
+++ b/components/_util/type.ts
@@ -1,4 +1,3 @@
-export type Omit<T, K extends keyof T> = Pick<T, Exclude<keyof T, K>>;
 // https://stackoverflow.com/questions/46176165/ways-to-get-string-literal-type-of-array-values-without-enum-overhead
 export const tuple = <T extends string[]>(...args: T) => args;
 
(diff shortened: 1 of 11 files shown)


Dreamerryao:
> You just broke our code.
> 
> In your official documentation, there's a
> 
> ```
> export interface TimePickerProps extends Omit<PickerTimeProps<Dayjs>, 'picker'> {}
> ```
> 
> which you broken by deleting the Omit

Emmm, so what happens? I delete useless Omit because TS support it after 3.5
https://www.typescriptlang.org/docs/handbook/utility-types.html

Dreamerryao:
And Typescript 3.5 pushlished since 2019 April. So I think we should embrace new features rather than add redundant code for compatibility. Please check your dependencies in `package.json` , `typescript` should >= 3.5.1

Issue #32961, opened 2021-11-21
ConfigProvider context not using the globalConfig for prefixCls


sirajulm:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/antd-reproduction-template-forked-nls4l?file=/index.js)

### Steps to reproduce
- `import ConfigProvider`
- Use `ConfigProfider.config({prefixCls: 'custom'})` in the root of the project folder before any usage of `ConfigProvider`
- Check the component class name prefix, it still uses `defaultGetPrefixCls` from `components/config-provider/context`

### What is expected?
ConfigProfider.config should be able to configure the globals for the ConfigProvider and it's context.

I think the order of resolution for `prefixCls` should be as (cascaded):
1. `defaultPrefixCls`
2. `ConfigProvider.config(...)`
3. `<ConfigProvider prefixCls="...">`
4. `<Component prefixCls="...">`

Examples:
1. default prefix
    ```typescript
    <Button>Click me!</Button>  // <button class="ant-btn" />
    ```
2. with global configuration
    ```typescript
    ConfigProvider.config({prefixCls: 'custom'});
    <Button>Click me!</Button>  // <button class="custom-btn" />
    ```
3. with global config   Provider
    ```typescript
    ConfigProvider.config({prefixCls: 'custom'});
    <ConfigProvider prefixCls="test">
      <Button>Click me!</Button>  // <button class="test-btn" />
    <ConfigProvider>
    ```
4. with global config   Provider   component override (no change, as currently applied)
    ```typescript
    ConfigProvider.config({prefixCls: 'custom'});
    <ConfigProvider prefixCls="test">
      <Button prefixCls="rand">Click me!</Button>  // <button class="rand" />
    <ConfigProvider>
    ```

### What is actually happening?
Currently `config-provider/context` does not consume the configs passed from `ConfigProvider.config`.

| Environment | Info |
|---|---|
| antd | 4.17.0 |
| React | 16+  |
| System | All |
| Browser | All |

---
`prefixCls` prop on the component seem to ignore the suffix to be used, even in the current implementation. I'm not sure if that is a bug or is expected to work the same way as implemented. Hence I'm not making any changes on it.

sirajulm:
@afc163 Looking forward for your feedback. 
[...]

PR #32962, opened 2021-11-21
fix: update ConfigProvider context to  use globalConfig 

sirajulm:
- Move global config functions to seperate file
- Reuse global config prefixCls for the context
- Move types to seperate file

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

- https://github.com/ant-design/ant-design/issues/32961

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | `ConfigProvider.config({...})` can now be used to globally set the `prefixCls` for the components.  |
| 🇨🇳 Chinese | # Need help in translation 🙂 |

### DIFF
--- a/components/config-provider/context.tsx
+++ b/components/config-provider/context.tsx
@@ -3,6 +3,7 @@ import defaultRenderEmpty, { RenderEmptyHandler } from './renderEmpty';
 import { Locale } from '../locale-provider';
 import { SizeType } from './SizeContext';
 import { RequiredMark } from '../form/Form';
+import { globalConfig } from './global-config';
 
 export interface Theme {
   primaryColor?: string;
@@ -46,15 +47,9 @@ export interface ConfigConsumerProps {
   };
 }
 
-const defaultGetPrefixCls = (suffixCls?: string, customizePrefixCls?: string) => {
-  if (customizePrefixCls) return customizePrefixCls;
-
-  return suffixCls ? `ant-${suffixCls}` : 'ant';
-};
-
 export const ConfigContext = React.createContext<ConfigConsumerProps>({
   // We provide a default function for Context without provider
-  getPrefixCls: defaultGetPrefixCls,
+  getPrefixCls: globalConfig().getPrefixCls,
 
   renderEmpty: defaultRenderEmpty,
 });
--- a/components/config-provider/global-config.ts
+++ b/components/config-provider/global-config.ts
@@ -1,0 +1,61 @@
+import { Theme } from './context';
+import { registerTheme } from './cssVariables';
+import { ConfigProviderProps } from './type';
+
+export const defaultPrefixCls = 'ant';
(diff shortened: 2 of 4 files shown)


crazyair:
ConfigProvider.config  是为了静态方法使用的（可能 antd v5 会把静态方法类移除掉就不需要 ConfigProvider.config 了）。

组件类推荐 <ConfigProvider prefixCls ...

sirajulm:
@crazyair Thank you for the reply. 
Does that mean once the static method `ConfigProvider.config`, there would be no other way to change the `prefixCls` statically?
`<ConfigProvider prefixCls="" />` does help to change the prefix dynamically, having a static configuration is good to have, in cases of creating a customized component library on top of antd. It look simpler to call the static `config()` once instead of wrapping each exposed component with `<ConfigProvider ... />

Let me know your thoughts. 🙂

crazyair:
组件库的话，也不需要配置`prefixCls`，而是在使用的项目去配置。

sirajulm:
The solution would work with simple and medium scale projects, but would not scale for complex projects using multiple packages/micro-frontends which in turn uses different versions of antd.

That would not help  in  projects where the host application uses a different version of `antd` compared to the one the custom component library uses, adding up to style conflicts. It would be quite difficult to keep all the projects running with same version of `antd` throughout its lifecycle.

Also,
```javascript
import {Button} from '@org/components'; // seperate repo with customization over antd components.
import {ConfigProvider, Input} from 'antd';
<ConfigProvider prefixCls="custom">
  <Input {...props} />
  <Button {...props} />
</ConfigProvider>
```
The above code would render
```html
<input class="custom-input" />
<button class="ant-btn"> 
```
with no effect of ConfigProvider on the button (as expected)

Which would require the `@org/components` to wrap each exposed component with a `ConfigProvider`

sirajulm:
I'm interested to know if there is any public accessible roadmap on antd v5? 🙂

crazyair:
```tsx
import {ConfigProvider, Input} from 'antd';
import {OrgConfigProvider, Button} from '@org/components';

<ConfigProvider prefixCls="ant">
    <OrgConfigProvider prefixCls="custom">
      <Input  />
      <Button />
    </OrgConfigProvider>
</ConfigProvider>
```

sirajulm:
@crazyair Hmm... That also should be working. It was my mistake that I didn't expose the ConfigProvider from the custom library. This should be working for me Though I see that the host application is made responsible for setting the context of a customized library, I think that is fine.

We can close this PR and the related issue raised 🙂

PR #33131, opened 2021-12-02
refactor: replace let with const

sunkuangdong:


sunkuangdong:
replace let with const

Issue #33572, opened 2022-01-05
table sorting should set aria-sort


dgreene1:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### What problem does this feature solve?
`aria-sort` allows screen readers to tell users that a table is being sorted

### What does the proposed API look like?
Following the standard set in [Section 3.23 of WAI-ARIA Authoring Practices 1.2
](https://w3c.github.io/aria-practices/#table_roles_states_props) 

An example of this can be seen in https://w3c.github.io/aria-practices/examples/table/sortable-table.html where the screen reader will read "Last Name column 2 sorted ascending"

The can be accomplished by:
1. Install NVDA
2. navigate to https://w3c.github.io/aria-practices/examples/table/sortable-table.html
3. type T on your keyboard to hear it describe the table
4. press the down arrow on your keyboard to get to the headers
5. press ctrl, alt, and the right arrow

So if we were to automatically put aria-sort on the header tag, we could get the same behavior. The value of this attribute is described in https://w3c.github.io/aria/#aria-sort

I would like to submit the PR for this issue in the next week or so. Please confirm that a PR would be considered?

PR #33603, opened 2022-01-06
feat: resolves issue 33572 re aria-sort on tables

dgreene1:

### 🤔 This is a ...

- [X] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

closes https://github.com/ant-design/ant-design/issues/33572

### 💡 Background and solution

The problem was described in https://github.com/ant-design/ant-design/issues/33572.

No API changes were made.

The only difference is that the `aria-sort` attribute shows up now on the header tag. For example: `aria-sort="descending"`

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Table adds `aria-sort` attribute for screen readers.  |
| 🇨🇳 Chinese | Table 增加 aria-sort 属性以优化屏幕阅读器的使用体验。 |

- [X] Doc is updated/provided or **not needed**
- [X] **Demo is updated/provided** or not needed
- [X] TypeScript definition is updated/provided or **not needed**
- [X] **Changelog is provided** or not needed

### DIFF
--- a/components/table/__tests__/Table.sorter.test.js
+++ b/components/table/__tests__/Table.sorter.test.js
@@ -68,19 +68,55 @@ describe('Table.sorter', () => {
       ),
     );
 
+    const getNameColumn = () => wrapper.find('th').at(0);
+
     expect(renderedNames(wrapper)).toEqual(['Tom', 'Lucy', 'Jack', 'Jerry']);
+    expect(getNameColumn().prop('aria-sort')).toEqual('descending');
+  });
+
+  it('should change aria-sort when default sort order is set to descend', () => {
+    const wrapper = mount(
+      createTable(
+        {
+          sortDirections: ['descend', 'ascend'],
+        },
+        {
+          defaultSortOrder: 'descend',
+        },
+      ),
+    );
+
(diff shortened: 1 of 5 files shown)


afc163:
图片

afc163 (approved):

afc163:
图片

coverage percent is down.

dgreene1:
@afc163 I’m surprised that coverage went down since I added unit tests and enhanced many other tests. I will bring it back to 100% tomorrow (Monday).

dgreene1:
@afc163 I removed the one line that isn't actually reachable by tests. I had put it in there as an attempt at an "assert never" type of approach. But I think give that there is only 3 sort states (up, down, and inactive) it's fine to remove the line that I had added.

useSorter should be at 100% code coverage now.

PR #33649, opened 2022-01-10
fix: Warning: Encountered two children with the same key, 'seperator'

dimbslmh:

### 🤔 This is a ...

- [ ] New feature
- [X] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#33183

### 💡 Background and solution

> Warning: Encountered two children with the same key, 'seperator'. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.

Append index to seperator key

### 📝 Changelog
No breaking changes

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/cascader/index.tsx
+++ b/components/cascader/index.tsx
@@ -45,7 +45,7 @@ function highlightKeyword(str: string, lowerKeyword: string, prefixCls: string |
 
     if (index % 2 === 1) {
       originWorld = (
-        <span className={`${prefixCls}-menu-item-keyword`} key="seperator">
+        <span className={`${prefixCls}-menu-item-keyword`} key={`seperator-${index}`}>
           {originWorld}
         </span>
       );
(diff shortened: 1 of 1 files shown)


dimbslmh:
> Preview Build / build preview (pull_request) Failing after 4m — build preview 


Check fail inherited from https://github.com/ant-design/ant-design/commit/b18c09cf32c80dca63533fc931925ea50d2e0d0c

PR #33736, opened 2022-01-16
Fix sub menu collapse bug 

KennyMacheka:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link
#32718 

### 💡 Background and solution
After investigation, it appears this 'bug' is a by-product of the sliding animation of a dropdown menu if this option is selected. Actual changes to this would be required inside react-component and possible even further up.

A solution I propose to this is for a dropdown menu to instantly appear, as opposed to the sliding down animation.
Then it won't be possible for the sub-menu to collapse as described in the issue.

This may be an undesirable UI, so I've submitted this as a draft PR to see if the proposal is sound. If so, I can also modify some of the failing tests for this new code (and write new ones where necessary).

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/dropdown/dropdown.tsx
+++ b/components/dropdown/dropdown.tsx
@@ -67,18 +67,6 @@ const Dropdown: DropdownInterface = props => {
     direction,
   } = React.useContext(ConfigContext);
 
-  //   const getTransitionName = () => {
-  //     const rootPrefixCls = getPrefixCls();
-  //     const { placement = '', transitionName } = props;
-  //     if (transitionName !== undefined) {
-  //       return transitionName;
-  //     }
-  //     if (placement.indexOf('top') >= 0) {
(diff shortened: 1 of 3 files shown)


li-jia-nan:
Hey! Conflicting files, and it seems this is now really outdated. If this is something that still makes sense to the latest master version of antd, please send a new PR with this change :)

PR #34178, opened 2022-02-23
style: remove overriding !important from input border right width styles

aghArdeshir:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [x] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

There is actually no issue for it, but I left a comment on this PR: https://github.com/ant-design/ant-design/pull/10492#issuecomment-1041154579

### 💡 Background and solution

This is actually a fix on a style that is merged in a PR: https://github.com/ant-design/ant-design/pull/10492

This `!important` added here to `border-right-width` of input causes problems with our styling. We can't easily override it without having `!important`. And if we add `!important` to our input styling, it will mess up our application as it should be configurable by some settings.

I believe we should not specify border-width for hover state of an element at all. because it won't happen usually that in hover/focus state somebody wants to change right-border-width specifically; unless in animations, which they will face this `border-base-width 1px !important` that they need to override.

I tried to demonstrate the problem in this video:

https://user-images.githubusercontent.com/5755214/154209194-0d81d02e-6746-4dc8-8a0d-477f0ac82096.mp4

I will just mention the people contributed to the PR, so that they make sure if my changes break anything or not: 

@tfwww 
@afc163 
@morenyang 

### 📝 Changelog

I don't know if any part of the app is actually dependent heavily on this border-right-width. So I don't know if this is a breaking change or not.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Remove Input `!important` style.  |
| 🇨🇳 Chinese | 移除 Input 一处 `!important` 样式。    |

- [x] Doc is updated/provided or not needed (**Not Needed**)
- [x] Demo is updated/provided or not needed (**Not Needed**, instead the thing I did in the video will work correctly after this fix)
- [x] TypeScript definition is updated/provided or not needed (**Not Needed**)
- [x] Changelog is provided or not needed (**Not Needed**)

### DIFF
--- a/components/input/style/mixin.less
+++ b/components/input/style/mixin.less
@@ -30,14 +30,14 @@
     border-color: @hoverBorderColor;
     box-shadow: @input-outline-offset @outline-blur-size @outline-width @outlineColor;
   }
-  border-right-width: @border-width-base !important;
+  border-right-width: @border-width-base;
   outline: 0;
 }
 
 // == when hover
 .hover(@color: @input-hover-border-color) {
   border-color: @color;
-  border-right-width: @border-width-base !important;
+  border-right-width: @border-width-base;
 }
 
 .disabled() {


yoyo837:
Maybe `feature` branch is better.

aghArdeshir:
@yoyo837 I doubt it. The way I see it, it was a defect preventing user custom styles from being applied

PR #34529, opened 2022-03-16
fix: [table] ColumnsType should be readonly

tychenjiajun:
It's readonly in rc-table https://github.com/react-component/table/blob/master/src/interface.ts. It should be readonly in antd as well.

### 🤔 This is a ...

TypeScript definition update

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

TypeScript definition is updated.


tychenjiajun:
dup of #31888, closed

Issue #34369, opened 2022-03-08
[textarea] cursor jumped to the end when exceed max length


boomler:
- [x] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/qq6c74?file=/index.js)

### Steps to reproduce
input anything in the middle text

### What is expected?
cursor's position not changed

### What is actually happening?
cursor's position jumps to the end

| Environment | Info |
|---|---|
| antd | undefined |
| React | All |
| System | All |
| Browser | All |

boomler:
I will take this.

EmilyyyLiu:
Hi @boomler 

Thanks for the bug report! I've tested with antd v6.3.7 and the issue seems to be resolved. When typing in the middle of the text and exceeding maxLength, the cursor now stays at the correct position instead of jumping to the end.

Demo: https://ant-design.antgroup.com/components/input#input-demo-textarea

Could you please help verify? If everything works fine on your end, I'll close this issue.

PR #34627, opened 2022-03-21
fix(input): cursor should not jump to the end after inputing

boomler:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/ant-design/ant-design/issues/34369

### 💡 Background and solution
Antd will set a formatted value  after inputing, this will make browser can't reset cursor's position.

I add a state to store cursor's position and reset it back after back. It's quick like what input-number does: https://github.com/react-component/input-number/blob/master/src/hooks/useCursor.ts

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |      fix  textarea's cursor jumps to the end when inputing     |
| 🇨🇳 Chinese |  修复某些输入情况下，textarea 跳跃到末尾的问题    |

### DIFF
--- a/components/input/TextArea.tsx
+++ b/components/input/TextArea.tsx
@@ -14,6 +14,7 @@ import {
   getMergedStatus,
 } from '../_util/statusUtils';
 import ClearableLabeledInput from './ClearableLabeledInput';
+import { useCursor } from './hooks/useCursor';
 import { fixControlledValue, InputFocusOptions, resolveOnChange, triggerFocus } from './Input';
 
 interface ShowCountProps {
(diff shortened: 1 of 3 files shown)


components/input/TextArea.tsx
```diff
@@ -84,14 +69,47 @@ const TextArea = React.forwardRef<TextAreaRef, TextAreaProps>(
 
     const innerRef = React.useRef<RcTextArea>(null);
     const clearableInputRef = React.useRef<ClearableLabeledInput>(null);
+    const compositing = React.useRef(false);
```
boomler, line 72: useRef to get composition status immediately.

components/input/TextArea.tsx
```diff
@@ -139,15 +143,9 @@ const TextArea = React.forwardRef<TextAreaRef, TextAreaProps>(
     };
 
     const handleChange = (e: React.ChangeEvent<HTMLTextAreaElement>) => {
-      let triggerValue = e.target.value;
-      if (!compositing && hasMaxLength) {
-        // 1. 复制粘贴超过maxlength的情况 2.未超过maxlength的情况
-        const isCursorInEnd =
-          e.target.selectionStart >= maxLength! + 1 ||
-          e.target.selectionStart === triggerValue.length ||
-          !e.target.selectionStart;
```
boomler, line 148: `!e.target.selectionStart` is not correct, when cursor at the start position: e.target.selectionStart = 0

li-jia-nan:
Hey! Conflicting files, and it seems this is now really outdated. If this is something that still makes sense to the latest master version of antd, please send a new PR with this change :)

PR #35792, opened 2022-05-28
fix: Fix style when both `disabled` and `danger` are specified for dropdown menu

MasaoBlue:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

#### Background
When both `disabled: true` and `danger: true` are specified for a dropdown menu, the menu will be inactive, but the style of danger will be applied in priority as shown in the image.

image

#### Solution
Add `important` to the `.ant-dropdown-menu-item-disabled` style. This is similar to the definition of [`.ant-menu-item-disabled`](https://github.com/ant-design/ant-design/blob/fd2ebe5/components/menu/style/index.less#L658).

No test is implemented because only the less file is modified.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Add `important` to the `.ant-dropdown-menu-item-disabled` style |
| 🇨🇳 Chinese | 在 `.ant-dropdown-menu-item-disabled` 样式中添加 `important` |

### DIFF
--- a/components/dropdown/style/index.less
+++ b/components/dropdown/style/index.less
@@ -227,12 +227,12 @@
       }
 
       &-disabled {
-        color: @disabled-color;
+        color: @disabled-color !important;
         cursor: not-allowed;
 
         &:hover {
-          color: @disabled-color;
-          background-color: @dropdown-menu-submenu-disabled-bg;
+          color: @disabled-color !important;
+          background-color: @dropdown-menu-submenu-disabled-bg !important;
           cursor: not-allowed;
         }
 


components/dropdown/style/index.less
```diff
@@ -227,12 +227,12 @@
...
       &-disabled {
-        color: @disabled-color;
+        color: @disabled-color !important;
         cursor: not-allowed;
 
         &:hover {
-          color: @disabled-color;
-          background-color: @dropdown-menu-submenu-disabled-bg;
+          color: @disabled-color !important;
+          background-color: @dropdown-menu-submenu-disabled-bg !important;
```
afc163, line 235: 能不用 !important 么？

MasaoBlue: Thank you for your comment.

This option is necessary. If this important is removed, the background color will be red when hovering.

image

li-jia-nan:
Hey! Conflicting files, and it seems this is now really outdated. If this is something that still makes sense to the latest master version of Antd, please send a new PR with this change :)

PR #36288, opened 2022-06-29
fix: Fixed an error when the loading property of the Button component…

kejianfeng:
… was passed into null

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

[https://github.com/ant-design/ant-design/issues/36282](https://github.com/ant-design/ant-design/issues/36282)

### 💡 Background and solution

- background:  An error occurs when the loading attribute of the Button component is passed into null
- solution： I hope to change typeof loading === 'object' in the source code to Object.prototype.toString.call(loading)
This will not report an error even if it is null. In this way, the scene component itself that will report an error should be avoided. No matter what parameters are passed in, the error will not cause the page to crash directly.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     When determining whether the passed loading parameter is an object, change typeof to Object.prototype.toString.call     |
| 🇨🇳 Chinese |      判断传入的loading参数是否为对象时，把 typeof 改为 Object.prototype.toString.call     |

### DIFF
--- a/components/button/button.tsx
+++ b/components/button/button.tsx
@@ -190,7 +190,7 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
 
   // =============== Update Loading ===============
   const loadingOrDelay: Loading =
-      loading?.delay ?? !!loading;
+      typeof loading === 'boolean' ? loading : loading?.delay;
 
   React.useEffect(() => {
     let delayTimer: number | null = null;
--- a/components/button/button.tsx
+++ b/components/button/button.tsx
@@ -190,7 +190,7 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
 
   // =============== Update Loading ===============
   const loadingOrDelay: Loading =
-    typeof loading === 'object' && loading.delay ? loading.delay || true : !!loading;
+    Object.prototype.toString.call(loading) === '[object Object]' && loading.delay ? loading.delay || true : !!loading;
 
   React.useEffect(() => {
     let delayTimer: number | null = null;
--- a/components/button/button.tsx
+++ b/components/button/button.tsx
@@ -190,7 +190,7 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
 
   // =============== Update Loading ===============
   const loadingOrDelay: Loading =
-      typeof loading === 'boolean' ? loading : loading?.delay;
+      typeof loading === 'boolean' ? loading : (loading?.delay || true);
 
   React.useEffect(() => {
     let delayTimer: number | null = null;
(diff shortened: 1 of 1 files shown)


yoyo837:
How about using `Optional Chaining Operator`？

components/button/button.tsx
```diff
@@ -190,7 +190,7 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
 
   // =============== Update Loading ===============
   const loadingOrDelay: Loading =
-    typeof loading === 'object' && loading.delay ? loading.delay || true : !!loading;
+    Object.prototype.toString.call(loading) === '[object Object]' && loading.delay ? loading.delay || true : !!loading;
```
afc163, line 193:
```suggestion
      loading?.delay ?? !!loading;
```

这样应该就行了。

components/button/button.tsx
```diff
@@ -190,7 +190,7 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
 
   // =============== Update Loading ===============
   const loadingOrDelay: Loading =
-    Object.prototype.toString.call(loading) === '[object Object]' && loading.delay ? loading.delay || true : !!loading;
+      loading?.delay ?? !!loading;
```
afc163, line 193:
```suggestion
    loading?.delay ?? !!loading;
```

MadCcc: 断言不如还是判断下类型？我觉得断言都是有隐患的。。

afc163:
```suggestion
     typeof loading === 'boolean' ? loading : (loading?.delay || true);
```

components/button/button.tsx
```diff
@@ -190,7 +190,7 @@ const InternalButton: React.ForwardRefRenderFunction<unknown, ButtonProps> = (pr
 
   // =============== Update Loading ===============
   const loadingOrDelay: Loading =
-    typeof loading === 'object' && loading.delay ? loading.delay || true : !!loading;
+      typeof loading === 'boolean' ? loading : loading?.delay;
```
afc163, line 193:
```suggestion
      typeof loading === 'boolean' ? loading : (loading?.delay || true);
```

kejianfeng:
Using Optional Chaining Operator can solve the problem, but in order to show that the purpose is to judge a real object, it is more reasonable to use Object.prototype.toString.call.


在 2022年6月29日 ***@***.***> 写道：


How about using Optional Chaining Operator？

PR #37001, opened 2022-08-10
feat: custom slider handle margin left in the vertical direction

### LINKED ISSUE #36989: Slider: expose "@slider-handle-margin-left" instead of hardcoded value

We have created an override system for ant-design that relies on "modify vars" of less. Part of our overrides is changing @slider-handle-size to 20px. Which requires us to change @slider-handle-margin-top (which takes care of horizontal case), But the vertical state is broken, since there is no @slider-handle-margin-left variable

In [here](https://github.com/ant-design/ant-design/blob/master/components/slider/style/index.less#L181) Instead of hard coded margin left, it would be great to use "@slider-handle-margin-left" variable, just like we do in [here](https://github.com/ant-design/ant-design/blob/master/components/slider/style/index.less#L43)

denghuacc:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [x] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

close #36989

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  add `@slider-handle-margin-left` to custom type |
| 🇨🇳 Chinese |  支持使用 `@slider-handle-margin-left` 定制样式 |

### DIFF
--- a/components/slider/style/index.less
+++ b/components/slider/style/index.less
@@ -178,7 +178,7 @@
 
     .@{slider-prefix-cls}-handle {
       margin-top: -6px; // we chould consider border width as well: (10 + 2 ) / 2
-      margin-left: -5px;
+      margin-left: @slider-handle-margin-left;
     }
 
     .@{slider-prefix-cls}-mark {
--- a/components/style/themes/default.less
+++ b/components/style/themes/default.less
@@ -893,6 +893,7 @@
 @slider-handle-color-tooltip-open: @primary-color;
 @slider-handle-size: 14px;
 @slider-handle-margin-top: -5px;
+@slider-handle-margin-left: -5px;
 @slider-handle-shadow: 0;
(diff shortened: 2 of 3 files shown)


components/slider/style/index.less
```diff
@@ -178,7 +178,7 @@
 
     .@{slider-prefix-cls}-handle {
       margin-top: -6px; // we chould consider border width as well: (10 + 2 ) / 2
```
alanhaledc, line 180: is it an invalid style? remove it?

afc163 (approved):

Issue #37134, opened 2022-08-18
Allow for passing key prop to breadcrumb items


lopno:
### What problem does this feature solve?

If two routes in a breadcrumb have the same path you will get the following warning.
```
Warning: Encountered two children with the same key, `KEY`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in div (created by Breadcrumb)
...
```

A similar issue was raised because earlier the key was based on the `breadcrumbName` prop: https://github.com/ant-design/ant-design-pro/issues/3904
This is essentially the same issue, just for path.

### What does the proposed API look like?

The route interface would be expanded with a new optional field called `key`, which would be the first choice for a key of a breadcrumb route.
[Route interface](https://ant.design/components/breadcrumb/#routes)
```
interface Route {
  path: string;
  breadcrumbName: string;
  children: Array<{
    path: string;
    breadcrumbName: string;
  }>;
  key?: string;
}
```

The logic for determining the key would look like this:
```
...
key: child.key || child.path || child.breadcrumbName,
...
```
[Current implementation](https://github.com/ant-design/ant-design/blob/master/components/breadcrumb/Breadcrumb.tsx#L103)

This would not break any existing implementations, but would fix the issue I am facing.

MadCcc:
`path` should be unique, what's your scenario?

lopno:
@MadCcc `customers / :customerId / specific_subpage / :subpage_id`

I want `specific_subpage` to point to the same page as `customers` because I don't have a page for this route without having a subpage_id. Because path is a required field I cannot omit it.

federicojasson:
In my particular case, I was trying to use:
```node
const routes = [
  {
    path: "/",
    breadcrumbName: <antdIcons.HomeOutlined />,
  },
  ...
]
```

But because the `/` is trimmed, and the `breadcrumbName` is converted to string as `[object Object]`, the `key` ended up being `[object Object]`, which isn't the best.

Here's a workaround defining a wrapper component:
```node
import { Breadcrumb } from "antd";

export default function BreadcrumbWrapper(props) {
  const { items } = props;

  const paths = {}; // Map: keys to paths
  const routes = [];

  for (const item of items) {
    paths[item.key] = item.path;

    routes.push({
      path: item.key,
      breadcrumbName: item.name,
    });
  }

  const params = {
    paths,
  };

  const render = (route, params, routes) => {
[...]

PR #37141, opened 2022-08-18
Add key prop to breadcrumb in order to support multiple routes with s…

lopno:

### 🤔 This is a ...

- [x] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

https://github.com/ant-design/ant-design/issues/37134

### 💡 Background and solution

If two routes in a breadcrumb have the same path you will get the following warning.
```
Warning: Encountered two children with the same key, `KEY`. Keys should be unique so that components maintain their identity across updates. Non-unique keys may cause children to be duplicated and/or omitted — the behavior is unsupported and could change in a future version.
    in div (created by Breadcrumb)
...
```

This solution adds a new optional field called `key`, which would be the first choice for a key of a breadcrumb route.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🆕 Breadcrumb Route supports `key` |
| 🇨🇳 Chinese | 🆕 Breadcrumb Route 支持 `key` |

### DIFF
--- a/components/breadcrumb/Breadcrumb.tsx
+++ b/components/breadcrumb/Breadcrumb.tsx
@@ -11,6 +11,7 @@ import BreadcrumbSeparator from './BreadcrumbSeparator';
 export interface Route {
   path: string;
   breadcrumbName: string;
+  key?: string;
   children?: Omit<Route, 'children'>[];
 }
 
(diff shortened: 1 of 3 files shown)


lopno:
Implemented here https://github.com/ant-design/ant-design/pull/41434

PR #38249, opened 2022-10-27
feat: DatePicker support `presets`

zombieJ:

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   Adjust DatePicker style and support `presets` prop.        |
| 🇨🇳 Chinese |   调整 DatePicker 布局并且支持 `presets` 属性。        |


gabschne:
While I like this list style more than the current tag one, I think the fact that the text can be dynamic might be problematic, especially if it's a longer one:

image

This means that, depending on the (length of the) preset texts, the calendar section of the picker is sometimes further to the left and sometimes further to the right, which is not really good for the user experience.

Have you thought about putting the list on the right side of the calendar part or below it (below would likely require to truncate the text using `text-overflow: ellipsis` if it's longer than the width of the picker popup) or even making this configurable?

Below could also be better for smaller screen sizes.

heiyu4585 (approved):

afc163:
覆盖率掉了

zombieJ:
> While I like this list style more than the current tag one, I think the fact that the text can be dynamic might be problematic, especially if it's a longer one:
> 
> image
> 
> This means that, depending on the (length of the) preset texts, the calendar section of the picker is sometimes further to the left and sometimes further to the right, which is not really good for the user experience.
> 
> Have you thought about putting the list on the right side of the calendar part or below it (below would likely require to truncate the text using `text-overflow: ellipsis` if it's longer than the width of the picker popup) or even making this configurable?
> 
> Below could also be better for smaller screen sizes.

Thanks for feedback. Let me bring this to designer.

Issue #38922, opened 2022-11-23
Spin component with tip property is using incorrect styles for non-default sizes


sribich:
### Reproduction link

[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/s/antd-reproduction-template-forked-hllyqg?file=/index.js)

### Steps to reproduce

Create a spin component with a `tip` and a `size` that is not `default`

```jsx
<Spin tip="Loading" size="large">
    content
</Spin>
```

### What is expected?

The spinning dots should be centered on the tip text.

### What is actually happening?

The spinning dots are misaligned with the tip text.

| Environment | Info |
| --- | --- |
| antd | 5.0.1 |
| React | 18.2.0 |
| System | Arch Linux |
| Browser | Firefox 108.0b5  |

PR #38923, opened 2022-11-23
fix: adjust spin styles to support sm/lg variants

sribich:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 🔗 Related issue link

#38922 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Spin alignment when using `tip`. |
| 🇨🇳 Chinese | 修复 Spin 包裹模式时的样式偏移问题。         |

### DIFF
--- a/components/spin/style/index.tsx
+++ b/components/spin/style/index.tsx
@@ -71,7 +71,7 @@ const genSpinStyle: GenerateStyle<SpinToken> = (token: SpinToken): CSSObject => 
           marginTop: -(token.spinDotSize / 2) - 10,
         },
 
-        [`> div > ${token.componentCls}-sm`]: {
+        '&-sm': {
           [`${token.componentCls}-dot`]: {
             margin: -token.spinDotSizeSM / 2,
           },
@@ -83,7 +83,7 @@ const genSpinStyle: GenerateStyle<SpinToken> = (token: SpinToken): CSSObject => 
           },
         },
 
-        [`> div > ${token.componentCls}-lg`]: {
+        '&-lg': {
           [`${token.componentCls}-dot`]: {
             margin: -(token.spinDotSizeLG / 2),
           },
--- a/components/spin/__tests__/__snapshots__/demo-extend.test.ts.snap
+++ b/components/spin/__tests__/__snapshots__/demo-extend.test.ts.snap
@@ -288,57 +288,209 @@ exports[`renders ./components/spin/demo/size.tsx extend context correctly 1`] = 
 
 exports[`renders ./components/spin/demo/tip.tsx extend context correctly 1`] = `
 <div
-  class="ant-spin-nested-loading"
+  class="ant-space ant-space-vertical"
+  style="width:100%"
 >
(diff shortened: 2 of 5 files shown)


sribich:
Since this is a style only change, this cannot be tested using normal component tests.

I also don't think it's possible to add an image test either due to the animated nature of the component. I would test to see if this is in fact the case, but unfortunately `npm run test-image` is failing for me.

Do you have any recommendations for adding tests, or is this an instance it's okay to skip?

sribich:
Also is adding a native Chinese translation something a maintainer will do, or how should I go about adding those? 
I would prefer to not add a low quality machine translation, but if that's the norm then I can do that.

sribich:
I have updated the docs to show sized tip examples:

image

components/spin/style/index.tsx
```diff
@@ -71,7 +71,7 @@ const genSpinStyle: GenerateStyle<SpinToken> = (token: SpinToken): CSSObject =>
           marginTop: -(token.spinDotSize / 2) - 10,
         },
 
-        [`> div > ${token.componentCls}-sm`]: {
+        '&-sm': {
```
afc163, line 74: Why update this?

sribich: Removing all of the css rules to illustrate the problem gives us this:

```typescript
const genSpinStyle: GenerateStyle<SpinToken> = (token: SpinToken): CSSObject => ({
  [`${token.componentCls}`]: {
    '&-nested-loading': {
      [`> div > ${token.componentCls}`]: {
        [`> div > ${token.componentCls}-sm`]: {},
        [`> div > ${token.componentCls}-lg`]: {},
      },
    },
  },
});
```

## Proposed solution 1

```typescript
const genSpinStyle: GenerateStyle<SpinToken> = (token: SpinToken): CSSObject => ({
  [`${token.componentCls}`]: {
    '&-nested-loading': {
      [`> div > ${token.componentCls}`]: {
        '&-sm': {},
        '&-lg': {},
      },
    },
  },
});
```

## Proposed solution 2

```typescript
const genSpinStyle: GenerateStyle<SpinToken> = (token: SpinToken): CSSObject => ({
  [`${token.componentCls}`]: {
    '&-nested-loading': {
      [`> div > ${token.componentCls}`]: {},
      [`> div > ${token.componentCls}-sm`]: {},
      [`> div > ${token.componentCls}-lg`]: {},
    },
  },
});
```

afc163:
Grate fix, we need one last step to merge, could you run `npm test -- components/spin -u` in local and commit snapshot update?

sribich:
Snapshots have been updated

afc163 (approved):

PR #16437, opened 2019-05-06
fix: Print reject error when ActionButton catch promise rejects

jas0ncn:
### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / document update
- [ ] Component style update
- [ ] TypeScript definition update
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other (about what?)

### 👻 What's the background?

There is nothing output on browser console when an error occurs while `<Modal>` components call `onOk` callback function.

### 💡 Solution

Emit error manually by add `console.error(e)`

### 📝 Changelog

- English Changelog:
  - Print reject error when ActionButton catch promise rejects
  - Print reject error when `Modal`'s `onOk` callback retures an promise rejects
- Chinese Changelog (optional):
  - 当 `ActionButton` 捕获到 Promise 的 Reject 的时候，打印错误。
  - 当 `Modal` 组件 `onOk` 回调返回的 Promise 被 Reject 的时候，打印错误。

### ☑️ Self Check before Merge

- [x] Doc is updated/provided or not needed
- [x] Demo is updated/provided or not needed
- [x] TypeScript definition is updated/provided or not needed
- [x] Changelog is provided or not needed

PR #40018, opened 2023-01-04
fix: Allow users to handle promise rejections in ActionButton's onOk callback

### LINKED ISSUE #40017: ActionButton swallows rejected promises
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/s/antd-reproduction-template-forked-m32ijg?file=%2Findex.js)

### Steps to reproduce

1. Open a modal dialog with `Modal.confirm`,
2. That has an asynchronous `onOk` handler,
3. That returns a Promise that rejects and is not caught.

### What is expected?

The browser logs an unhandled promise rejection warning.

- The modal dialog stays open
- An error message is logged
- The rejected promise is swallowed silently and does not bubble up so the browser can handle it

| Environment | Info |
| --- | --- |
| antd | 5.1.2 |
| React | react-18 |
| System | macOS |
| Browser | Firefox |

---
In issue [#16437](https://github.com/ant-design/ant-design/pull/16437) a fix has been added to log an error in case the `onOk` handler rejects. While this allows a developer to see the problem, it still does not allow us to handle such an error in code and, for example, log the error remotely or present the user with a global error message.

While it is important that a promise rejection is handled in ActionButton to clean up, it should not be silently swallowed.

This PR proposes to let the promise rejection bubble up by re-throwing it after cleanup has been performed. It removes the logging code from #16437 and lets the browser log an unhandled promise rejection instead.

Regarding the tests: due to an [issue in Jest](https://github.com/facebook/jest/issues/5620), unhandled promise rejections can't be tested for. There is a [workaround](https://johann.pardanaud.com/blog/how-to-assert-unhandled-rejection-and-uncaught-exception-with-jest/) that I have implemented here. I see two options: keep the workaround and accept that there is more test scaffolding that needs to be done, or remove the test entirely.

grossbart:
Components that rely on ActionButton swallow rejected promises. This makes it impossible for userland code to handle them.

### 🤔 This is a ...

- [x] Bug fix

### 🔗 Related issue link

- Closes #40017 
- Builds on #16437

### 💡 Background and solution

Components that rely on ActionButton swallow rejected promises in their `onOk` handler. This makes it impossible for userland code to react to these errors, e.g. to log them to a remote server or to present users with an error message.

Rejected promises should not be swallowed so userland code or the browser can handle them.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Allow users to handle promise rejections in ActionButton's `onOk` callback |
| 🇨🇳 Chinese | Allow users to handle promise rejections in ActionButton's `onOk` callback |


afc163:
图片

components/_util/ActionButton.tsx
```diff
@@ -55,12 +55,10 @@ const ActionButton: React.FC<ActionButtonProps> = (props) => {
         clickedRef.current = false;
       },
       (e: Error) => {
-        // Emit error when catch promise reject
-        // eslint-disable-next-line no-console
-        console.error(e);
         // See: https://github.com/ant-design/ant-design/issues/6183
         setLoading(false, true);
         clickedRef.current = false;
+        throw e;
```
zombieJ, line 61: Seems just return `Promise.reject` instead is much meaningful here : )

grossbart:
@afc163 I've investigated why it works for me but not on CI: The problem is that as soon as Jest runs tests in parallel the proposed workaround doesn't work anymore.

My pragmatic suggestion would be to remove the test and the workaround: it just doesn't seem to be worth the trouble and in the experiments I did today I could not find a way that works across Jest processes. If you feel very strongly that a test should exist and chances are high that you would accept this PR I would investigate further.

At this moment I'm not sure a solution is possible without going deep into the internals of Jest, which would make a workaround brittle. Jest has had this issue open for four years, so I'm not sure that is on their roadmap: https://github.com/facebook/jest/issues/5620.

zombieJ:
Remove is not a good way. Could you help to find way out instead just remove it?

grossbart:
Me and a colleague dug deep into Jest's source code for several hours to see whether we could find a workaround. We could not 😞 Jest a) does not provide a real process object and b) handles `unhandledRejection` events itself, making it impossible in our understanding to change.

So, I don't see a way that this can be tested with Jest and see no other option than to remove the test …

zombieJ:
Let me try other way : )

zombieJ:
Update. I found that this update is meanless. Nowhere to handle the return Promise.reject on `onOk`. If not care about rejection warning. Just:
https://codesandbox.io/s/antd-reproduction-template-forked-reiyjf

grossbart:
We're actively using this patch in production, so I know it's not meaningless 🤷 

Of course: in most situations that promise should be handled in `onOk` like you've shown. But we do not handle promise rejections in situations where no expected domain exceptions occur in the `onOk` handler. Instead we want the error to bubble up to our error boundary to handle it there. (We log it to our servers and show an error message to users.)

Without this patch, errors are swallowed and users won't know that their action didn't work. Also, as seen in #16437, it is helpful to developers to see when an error occurred. And letting an unhandled promise rejection bubble up to the browser would be the correct way to do so.

With this PR I'm trying to say: Antd should not catch the promise in library code but let developers handle it.

zombieJ (approved):

PR #40112, opened 2023-01-09
feat: improve theme config to support custom tokens

### LINKED ISSUE #40006: Ant theme custom tokens v5

The main issue is that we need a few background tokens that we could change and that would not change the other components.
In the documentation, I found these variables for the background: `colorBgContainer, colorBgElevated, colorBgLayout, colorBgSpotlight, colorBgMask`. The problem with these tokens is that we cannot use them as primary, secondary, tertiary and quaternary backgrounds for custom components or use them as background for custom container components, and by changing them we change how antd components look.
In v4 we had a custom logic that parsed the `less` files to extract `less variables` and then we could override variables or define new variables. Then we exported this to files for each theme.

So we had more flexibility for migration from v3 to v4. Now it looks like we need to have another definition file (for each theme) and then custom logic that would merge antd tokens with our custom variables. This also means that we cannot use `theme.useToken()` and need to have some custom implementation for that.

- The best solution (and future-proof) would be that we have the ability to modify `ThemeConfig` so that it allows us to define custom variables in the theme file and we could get them with `theme.useToken()`. Maybe not the whole `ThemeConfig` but `ThemeConfig.token` this would mean that type cannot be hardcoded to `token?: Partial<AliasToken>` but it should be generic or inferred.

Maybe something like this:
```ts
interface ThemeConfig<T = unknown> {
  token?: Partial<AliasToken> & T;
  components?: OverrideToken;
  algorithm?: MappingAlgorithm | MappingAlgorithm[];
  hashed?: boolean;
  inherit?: boolean;
}

type CustomVariables = {
  colorBg: string;
  colorBgSecondary: string;
  colorBgTertiary: string;
  colorBgQuaternary: string;
};

const theme: ThemeConfig<CustomVariables> = {
  token: {
    colorPrimary: '#1a6ae3',
    colorBgMask: 'rgba(34, 46, 84, 0.95)',
    colorBg: '#FFFFFF',
    colorBgSecondary: '#F3F4F9',
    colorBgTertiary: '#E5E7EB',
    colorBgQuaternary: '#D1D5DB',
  },
  components: {
    Tabs: {
      fontSize: 12,
    },
  },
  algorithm: antTheme.defaultAlgorithm,
};

export default theme;

```

- If not, for now, it would be ok to create 4 new tokens the same as we have for text components. (`colorText, colorTextSecondary, colorTextTertiary, colorTextQuaternary`)
Something like this:
```js
colorBg,
colorBgSecondary,
colorBgTertiary,
colorBgQuaternary
```

fihorvat:

### 🤔 This is a ...

- [] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [x] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

- Close #40006 

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  Add theme custom tokens |
| 🇨🇳 Chinese |  添加主题自定义令牌  |

### DIFF
--- a/components/config-provider/__tests__/theme.test.tsx
+++ b/components/config-provider/__tests__/theme.test.tsx
@@ -117,6 +117,60 @@ describe('ConfigProvider.Theme', () => {
     expect(tokenRef?.colorPrimaryText).toBe('#177ddc');
   });
 
+  it('should support custom tokens with internal useToken hook', () => {
+    type CustomTokens = { customToken: string; customToken2: number };
+    let customTokenRef: any;
+    const Demo = () => {
+      const [, , , customTokens] = useToken<CustomTokens>();
+      customTokenRef = customTokens;
+
+      return null;
+    };
+    render(
+      <ConfigProvider
+        theme={{
+          token: { colorPrimary: '#1890ff' },
+          algorithm: [defaultAlgorithm],
+          customTokens: {
+            customToken: '#D3D3D3',
+            customToken2: 12,
+          } as CustomTokens,
+        }}
+      >
+        <Demo />
+      </ConfigProvider>,
+    );
+    expect(customTokenRef?.customToken).toBe('#D3D3D3');
+    expect(customTokenRef?.customToken2).toBe(12);
+  });
+
+  it('should support custom tokens', () => {
+    type CustomTokens = { customToken: string; customToken2: number };
+    let customTokenRef: any;
+    const Demo = () => {
+      const tokens = theme.useToken<CustomTokens>().customTokens;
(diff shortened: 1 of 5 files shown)


MadCcc:
https://github.com/ant-design/ant-design/issues/40006#issuecomment-1376665354

PR #40269, opened 2023-01-16
💄 STYLE: Button's color should override child anchor's color

ds1371dani:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [x] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution

image
https://codesandbox.io/s/antd-reproduction-template-forked-ecdy4f?file=/index.js

When there is an `a` tag inside a Button component, it should follow the Button's color

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | :lipstick: Fix Button text color when containing an a tag. |
| 🇨🇳 Chinese | :lipstick: 修复包含 a 标签时的 Button 文本颜色. |

### DIFF
--- a/components/button/style/index.ts
+++ b/components/button/style/index.ts
@@ -46,6 +46,10 @@ const genSharedButtonStyle: GenerateStyle<ButtonToken, CSSObject> = (token): CSS
         marginInlineStart: token.marginXS,
       },
 
+      '> a': {
+        color: 'currentColor',
+      },
+
       '&:not(:disabled)': {
         ...genFocusStyle(token),
       },


li-jia-nan:
Rather than nesting `<a>` inside the `<Button>`, it is better to use the href attribute directly, The advantage of doing this is the `<Button>` will be displayed as `<a>` in the browser

li-jia-nan (approved):

PR #40729, opened 2023-02-15
fix: adjust onBackground outputs to support semi-transparent colors

Jungzl:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution

I want to configure a theme that makes the background color of the Table component hover with transparency, but the previous version of the [`onBackground`](https://github.com/scttcper/tinycolor/pull/236#issue-1582100623) method had an error in the alpha compositing result for two semi-transparent colors. I also add a [`toHexShortString`](https://github.com/scttcper/tinycolor/pull/237#issue-1583642171) method to upstream so that this fix won't change related snapshots if output is not a semi-transparent color.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | adjust `onBackground` method outputs to support semi-transparent colors |
| 🇨🇳 Chinese | 调整 `onBackground` 方法的输出以支持两个半透明色叠加 |

### DIFF
--- a/components/empty/simple.tsx
+++ b/components/empty/simple.tsx
@@ -1,6 +1,6 @@
+import { TinyColor } from '@ctrl/tinycolor';
 import * as React from 'react';
 import { useMemo } from 'react';
-import { TinyColor } from '@ctrl/tinycolor';
 import { useToken } from '../theme/internal';
 
 const Simple = () => {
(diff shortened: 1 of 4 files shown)


Jungzl:
I also find some usages in code like 
```typescript
color.setAlpha(*).toHexString()
```
which lost alpha channel, is it intentional?

afc163 (approved):

PR #41319, opened 2023-03-18
fix(input): missing rtl className for count in TextArea

ds1371dani:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [x] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution

When `showCount` is used, it should be positioned at the end of the element

image

Examples of the problem:

- https://ant.design/components/form?direction=rtl#components-form-demo-validate-static
- https://ant.design/components/input?direction=rtl#components-input-demo-textarea-show-count

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | :lady_beetle: Fix Input.TextArea position in RTL mode when using `showCount`. |
| 🇨🇳 Chinese | :lady_beetle: 使用 `showCount` 时修复了 RTL 模式下 Input.TextArea 的位置。 |

### DIFF
--- a/components/input/TextArea.tsx
+++ b/components/input/TextArea.tsx
@@ -104,6 +104,7 @@ const TextArea = forwardRef<TextAreaRef, TextAreaProps>(
           countWrapper: classNames(
             `${prefixCls}-textarea`,
             `${prefixCls}-textarea-show-count`,
+            { [`${prefixCls}-textarea-show-count-rtl`]: direction === 'rtl' },
             hashId,
           ),
           textarea: classNames(


components/input/TextArea.tsx
```diff
@@ -104,6 +104,7 @@ const TextArea = forwardRef<TextAreaRef, TextAreaProps>(
           countWrapper: classNames(
             `${prefixCls}-textarea`,
             `${prefixCls}-textarea-show-count`,
+            { [`${prefixCls}-textarea-show-count-rtl`]: direction === 'rtl' },
```
yoyo837, line 107:
```suggestion
            `${prefixCls}-textarea-show-count-rtl`: direction === 'rtl',
```
How about this?

ds1371dani: We can't do that, we need the object to be able to use our direction condition

afc163 (approved):

PR #44163, opened 2023-08-10
Fix notification config typings

StefanoMagrassi:
### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution

The global `notification.config()` method is defined as `any` (really bad...).

To fix it, I've just add the signature `(config: GlobalConfigProps) => void` to the `config` method of `BaseMethods` interface.

The signature is taken from `setNotificationGlobalConfig()` function which is the default implementation of `config` method.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Add signature to global `notification.config()` method |
| 🇨🇳 Chinese |           |

---

### <samp>🤖 Generated by Copilot at bd9bae4</samp>

Added a `.node-version` file and fixed some bugs and typos in the notification component and its tests. Improved the type definition of `notification.config` and its usage.

### <samp>🤖 Generated by Copilot at bd9bae4</samp>

*  Add `.node-version` file to specify Node.js version for the project ([link](https://github.com/ant-design/ant-design/pull/44163/files?diff=unified&w=0#diff-2e3b78afed59645cf5dec2a3b273df0ef8e815a1190c6106ab867f886c808af9R1))
*  Fix notification component bug by using `undefined` instead of `null` for `prefixCls` and `getContainer` properties in `notification.config` method and related tests ([link](https://github.com/ant-design/ant-design/pull/44163/files?diff=unified&w=0#diff-15d16dcef957890a3f45c164db9818dcc95d2fd79868850c7ff7d3ce44f0641dL20-R21), [link](https://github.com/ant-design/ant-design/pull/44163/files?diff=unified&w=0#diff-ed8976c7e832d9befa4199cc192b025c91cb3ff525a3649834d7545fdd89f690L23-R24), [link](https://github.com/ant-design/ant-design/pull/44163/files?diff=unified&w=0#diff-ed8976c7e832d9befa4199cc192b025c91cb3ff525a3649834d7545fdd89f690L156-R156), [link](https://github.com/ant-design/ant-design/pull/44163/files?diff=unified&w=0#diff-54f5ce6e753e0976aac6abeac40add61e61be5d71d134c97d2e2c06db6898d59L39-R40))
*  Improve type safety and documentation of `notification.config` method by using `GlobalConfigProps` type for its parameter ([link](https://github.com/ant-design/ant-design/pull/44163/files?diff=unified&w=0#diff-ec815250b7c2f7267ba9b8a93ea48315da09b7060dc31f38646dd7db8964a965L230-R230))


.node-version
```diff
@@ -0,0 +1 @@
+16.20.1
```
afc163, line 1: 为啥加这个

li-jia-nan: image

li-jia-nan:
I don't think we need `.node-version` file

StefanoMagrassi:
> I don't think we need `.node-version` file

@li-jia-nan can you explain why?

li-jia-nan:
> @li-jia-nan can you explain why?

In fact, for the contribution value, we not only support `node 16.20.1`, but also support more versions, which will cause ambiguity

StefanoMagrassi:
> In fact, for the contributors, we not only support `node 16.20.1`, but also support more versions, which will cause ambiguity

`.node-version` file is only used by `nodenv` and by the way is _almost always_ a good practice to set a common version for development tools.

But if you **really** do not want a `.node-version` in the source code, it should be added a note in the documentation for the supported Node versions (`16.x` maybe...)

afc163:
I prefer not to add `.node-version`.

StefanoMagrassi:
> I prefer not to add `.node-version`.

So, please add some kind of info about supported Node version for development in the documentation...

afc163:
> So, please add some kind of info about supported Node version for development in the documentation...

PR is welcome~

li-jia-nan (approved):

Issue #44194, opened 2023-08-13
Carousel doesn't respect `useTransform` prop


Gherciu:
### Reproduction link

[https://ant.design/components/carousel](https://ant.design/components/carousel)

### Steps to reproduce

Long story short:
Slick carousel has a property `useTransform` which is by default true, and this property is meant to restrict the usage of `transform` CSS property when `useTransform` is false, and This is needed only in few edge cases when peoples want to have a children in their carousel which may have fixed position, but there is Bug in CSS which mess the fixed position if any parent has a transform CSS property and because of that Slick implemented this `useTransform` to give the power to developers to restrict the usage of `transform` CSS properties if needed and give control to them

It doesn't matter at the moment if we give that as True or False, This is used and Respected by Slick carousel, but not by Ant design wrapper, there are some styles using Transform CSS properties even if we declare `useTransform={false}`

### What is expected?

`useTransform` prop should be respected and when it is FALSE no transform properties in CSS should be used same as SLick carousel does

### What is actually happening?

`useTrandform` property is just passed to slick carousel, but the wrapper doesn't respect that and in its styles it uses transform CSS even if we make `useTrandform={false}`
We need to respect this property even in ant design anyway users can't use it

| Environment | Info |
| --- | --- |
| antd | 5.8.3 |
| React | 18 |
| System | All |
| Browser | All |

zombieJ:
Which part not using `useTransform` ? Could provide more detail about this?

### useTransform=true
<img width="733" alt="截屏2023-08-14 10 10 27" src="https://github.com/ant-design/ant-design/assets/5378891/764ece1b-0642-41f8-8b3f-2c968fbc16ab">

### useTransform=false
<img width="589" alt="截屏2023-08-14 10 09 59" src="https://github.com/ant-design/ant-design/assets/5378891/0042c1f9-dcfb-4708-8421-5847b2bd85c0">

PR #44195, opened 2023-08-13
fix: respect use transform from slick

gherciu:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link
https://github.com/ant-design/ant-design/issues/44194

### 💡 Background and solution

I had no idea how to pass props to styles
Slick has a prop named `useTransform` and ideally we need to pass that to styles and use
but I had o idea how to do that so I used tokens, at least developers will have a solution for this edge case

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    Added a useTransform token to carousel so developers can restrict the usage of transform prop in carousel if they need that       |
| 🇨🇳 Chinese |           |

---

### <samp>🤖 Generated by Copilot at 8c9dda0</samp>

Added a new option to the carousel component to toggle CSS3 transforms. Updated the style generation logic to handle the option.

### <samp>🤖 Generated by Copilot at 8c9dda0</samp>

*  Add `useTransform` property to `ComponentToken` interface to control whether to use CSS3 transforms for carousel component ([link](https://github.com/ant-design/ant-design/pull/44195/files?diff=unified&w=0#diff-24c8a7b3adad15c6014a09fdd0200792fa3915c53a19bb5b9e702b6d4f0574fdR23-R27))
*  Modify `genCarouselStyle` function to conditionally apply `transform` property to `.slick-track` and `.slick-list` selectors based on `useTransform` token value ([link](https://github.com/ant-design/ant-design/pull/44195/files?diff=unified&w=0#diff-24c8a7b3adad15c6014a09fdd0200792fa3915c53a19bb5b9e702b6d4f0574fdL50-R56))
*  Set default value of `useTransform` token to `true` in `defaultToken` object to match existing behavior and ensure backward compatibility ([link](https://github.com/ant-design/ant-design/pull/44195/files?diff=unified&w=0#diff-24c8a7b3adad15c6014a09fdd0200792fa3915c53a19bb5b9e702b6d4f0574fdR374-R375))

### DIFF
--- a/components/carousel/style/index.ts
+++ b/components/carousel/style/index.ts
@@ -20,6 +20,11 @@ export interface ComponentToken {
    * @descEN Width of active indicator
    */
   dotActiveWidth: number;
+  /**
+   * @desc 是否使用 CSS3 变换
+   * @descEn Whether to use CSS3 transforms or not
+   */
+  useTransform: boolean;
 }
 
 interface CarouselToken extends FullToken<'Carousel'> {
@@ -47,7 +52,8 @@ const genCarouselStyle: GenerateStyle<CarouselToken> = (token) => {
         WebkitTapHighlightColor: 'transparent',
 
         '.slick-track, .slick-list': {
-          transform: 'translate3d(0, 0, 0)',
+          // Transforms should not be applied at all, if user has asked for it
+          ...(token.useTransform && { transform: 'translate3d(0, 0, 0)' }),
           touchAction: 'pan-y',
         },
       },
@@ -365,6 +371,8 @@ export default genComponentStyleHook(
       dotHeight: 3,
       dotWidthActive: dotActiveWidth,
       dotActiveWidth,
+      // In Slick this property is `true` by default, so making Antd's default value also as `true`
+      useTransform: true,
     };
   },
   {


Gherciu:
This is the bug I'm trying to fix https://github.com/ant-design/ant-design/issues/44194

If you guys know how to pass prop to styles so we can use that instead of tokens, I would love your help.

components/carousel/style/index.ts
```diff
@@ -47,7 +52,8 @@ const genCarouselStyle: GenerateStyle<CarouselToken> = (token) => {
         WebkitTapHighlightColor: 'transparent',
 
         '.slick-track, .slick-list': {
-          transform: 'translate3d(0, 0, 0)',
+          // Transforms should not be applied at all, if user has asked for it
+          ...(token.useTransform && { transform: 'translate3d(0, 0, 0)' }),
```
zombieJ, line 56: `transform` props is auto generate with inline style by `react-slick` which is not owned by antd part.

Gherciu: Yes, and when whe give `useTransform `as False to react slick it respects that and doesn't generate any transform

The problem is wit these styles in Antd it doesn't have any IF statement and just adds these to styles without checking if transforms are needed or not

https://react-slick.neostack.com/docs/api#useTransform

AbdullahWins (approved):
good job!

PR #45184, opened 2023-10-05
fix(FormItem):As a subcomponent of FormItem, the aria-label prop of the Select component should correspond to the label prop of the FormItem.

### LINKED ISSUE #45168: Select has unnecessary aria-label
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/s/l72926)

### Steps to reproduce

1. Use a Select inside a Form.Item
2. Assign a label to the form item

### What is expected?

The form item's label should be the label of the `<input />` inside the `Select`.

The input has an aria-label "search" which overrides the label assigned to the form item.

| Environment | Info |
| --- | --- |
| antd | 5.9.4 |
| React | 18.2.0 |
| System | Windows 10 |
| Browser | Firefox 117 |

---
The workaround is to set aria-label to undefined in the Select, but it shouldn't be necessary to do so.

liangkuku:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link
close https://github.com/ant-design/ant-design/issues/45168

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     fix(FormItem):As a subcomponent of FormItem, the aria-label prop of the Select component should correspond to the label prop of the FormItem.      |
| 🇨🇳 Chinese |     修复 FormItem 的 bug，Select 作为 FormItem 的子组件，其 `aria-label` 属性应该对应于 FormItem 的 `label` 属性。      |

---

### 🚀 Summary

Generated by Copilot at [dcf5ef](https://github.com/liangkuku/ant-design/commit/dcf5efda8246241c750c3573e0a76e7ea078ed58)
Assign the label prop of the FormItem to the aria-label prop of the child component.

### 🔍 Walkthrough

Assign the label prop of the FormItem to the aria-label prop of the child component.
To override Select default aria-label values

### DIFF
--- a/components/auto-complete/__tests__/__snapshots__/demo-extend.test.ts.snap
+++ b/components/auto-complete/__tests__/__snapshots__/demo-extend.test.ts.snap
@@ -692,6 +692,7 @@ exports[`renders components/auto-complete/demo/form-debug.tsx extend context cor
             class="ant-form-item-control-input-content"
           >
             <div
+              aria-label="单独 AutoComplete"
               class="ant-select ant-select-in-form-item ant-select-auto-complete ant-select-single ant-select-show-search"
             >
               <div
@@ -705,7 +706,7 @@ exports[`renders components/auto-complete/demo/form-debug.tsx extend context cor
                     aria-controls="rc_select_TEST_OR_SSR_list"
                     aria-expanded="false"
                     aria-haspopup="listbox"
-                    aria-label="Search"
+                    aria-label="单独 AutoComplete"
                     aria-owns="rc_select_TEST_OR_SSR_list"
                     autocomplete="off"
                     class="ant-select-selection-search-input"
@@ -763,6 +764,7 @@ exports[`renders components/auto-complete/demo/form-debug.tsx extend context cor
             class="ant-form-item-control-input-content"
           >
             <div
+              aria-label="单独 TreeSelect"
               class="ant-select ant-tree-select ant-select-in-form-item ant-select-single ant-select-show-arrow"
             >
               <div
@@ -776,7 +778,7 @@ exports[`renders components/auto-complete/demo/form-debug.tsx extend context cor
                     aria-controls="rc_select_TEST_OR_SSR_list"
                     aria-expanded="false"
                     aria-haspopup="listbox"
-                    aria-label="Search"
+                    aria-label="单独 TreeSelect"
                     aria-owns="rc_select_TEST_OR_SSR_list"
                     autocomplete="off"
                     class="ant-select-selection-search-input"
@@ -1099,6 +1101,7 @@ exports[`renders components/auto-complete/demo/form-debug.tsx extend context cor
             class="ant-form-item-control-input-content"
           >
             <div
+              aria-label="包含 search 图标正常"
               class="ant-select ant-select-in-form-item ant-select-auto-complete ant-select-single ant-select-customize-input ant-select-show-search"
             >
               <div
@@ -1115,7 +1118,7 @@ exports[`renders components/auto-complete/demo/form-debug.tsx extend context cor
                       aria-controls="rc_select_TEST_OR_SSR_list"
                       aria-expanded="false"
                       aria-haspopup="listbox"
-                      aria-label="Search"
+                      aria-label="包含 search 图标正常"
                       aria-owns="rc_select_TEST_OR_SSR_list"
                       autocomplete="off"
                       class="ant-input"
--- a/components/auto-complete/__tests__/__snapshots__/demo-extend.test.ts.snap
+++ b/components/auto-complete/__tests__/__snapshots__/demo-extend.test.ts.snap
@@ -677,9 +677,9 @@ exports[`renders components/auto-complete/demo/form-debug.tsx extend context cor
       >
         <label
           class=""
-          title="单独 AutoComplete"
+          title="Single AutoComplete."
         >
-          单独 AutoComplete
+          Single AutoComplete.
         </label>
       </div>
       <div
@@ -692,7 +692,7 @@ exports[`renders components/auto-complete/demo/form-debug.tsx extend context cor
             class="ant-form-item-control-input-content"
           >
             <div
-              aria-label="单独 AutoComplete"
+              aria-label="Single AutoComplete."
               class="ant-select ant-select-in-form-item ant-select-auto-complete ant-select-single ant-select-show-search"
             >
               <div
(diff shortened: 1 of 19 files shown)


yoyo837:
@khalibloo Is this what you expected?

khalibloo:
@yoyo837 I just had a look in the preview of the Form component. It works correctly now. Thanks

components/form/__tests__/index.test.tsx
```diff
@@ -778,7 +800,7 @@ describe('Form', () => {
     );
 
     // if getByLabelText can get element, then it is a validate field with form control and label
-    expect(screen.queryByLabelText('test')).not.toBeInTheDocument();
+    expect(screen.queryByLabelText('test')).toBeInTheDocument();
```
yoyo837, line 803: Why change this?

liangkuku: The Form.Item has an prop label 'test', the input element has an aria-label attribute also set to 'test' and it can locate based on the input's aria-label attribute by screen.queryByLabelText('test').

yoyo837: @zombieJ Please help to check this.

components/form/__tests__/__snapshots__/demo-extend.test.ts.snap
```diff
@@ -7532,6 +7588,7 @@ exports[`renders components/form/demo/label-debug.tsx extend context correctly 1
             class="ant-form-item-control-input-content"
           >
             <input
+              aria-label="[object Object]"
```
afc163, line 7591: This should be a wrong `aria-label`.

afc163, line 7591: This should be a wrong aria-label.

afc163 (changes requested):

components/form/__tests__/__snapshots__/demo-extend.test.ts.snap
liangkuku: sorry It was my negligence.The problem 'aria-label="[object Object]' arises because the label property of FormItem is a component.if the prop label of FormItem is component , then the aria-label of its child components should be what?

liangkuku: @ yoyo837 @afc163  @zombieJ Sorry,it was my negligence.The problem 'aria-label="[object Object]' arises because the label property of FormItem is a component.if the prop label of FormItem is component , then the aria-label of its child components should be what?
image

liangkuku: @yoyo837 @afc163 @zombieJ Sorry,it was my negligence.The problem 'aria-label="[object Object]' arises because the label property of FormItem is a component.if the prop label of FormItem is component , then the aria-label of its child components should be what?

yoyo837: `renderToString`?

liangkuku: @yoyo837  Extracting text from a component? In the example,regard 'longtextlongtext...' as input aria-label? What if label is an empty tag for example '<div></div>'?

liangkuku: @yoyo837 Extracting text from a component? In the example,regard 'longtextlongtext...' as input aria-label? What if label is an empty tag for example `<div></div>`?

yoyo837: Maybe other members have good suggestions... 🤔

liangkuku: I have an idea. if label is a component, the children property of props is searched recursively until children are of type String.If not , the default 'Search ' is used as the value of the aria-label property .

khalibloo: Not sure I understand completely, but how about `domElement.innerText`? That should handle both string and component children.

liangkuku: I know what you mean.But label is React.ReactNode not HTMLElement.What I want to do is as follows.
```
function getTextFromElement(element) {
  let text = '';

  React.Children.map(element.props.children, child => {
    if (React.isValidElement(child)) {
      text += getTextFromElement(child);
    } else if (typeof child === 'string') {
      text += child;
    }
  });

  return text;
}

const A = (
  <div>
    Hello <span>world</span>!
  </div>
);

const text = getTextFromElement(A);
console.log(text); // Output: "Hello world!"
```

liangkuku: @yoyo837 @afc163 @zombieJ I fix wrong aria-label. If you have time, please review Pr.

components/auto-complete/__tests__/__snapshots__/demo-extend.test.ts.snap
```diff
@@ -692,6 +692,7 @@ exports[`renders components/auto-complete/demo/form-debug.tsx extend context cor
             class="ant-form-item-control-input-content"
           >
             <div
+              aria-label="单独 AutoComplete"
```
kiner-tang, line 695: 为啥会出现中英文混合的内容，感觉不太好，demo 里面还是用纯英文吧

components/form/util.ts
```diff
@@ -59,3 +60,20 @@ export function getStatus<DefaultValue>(
   }
   return status;
 }
+
+/**
+ * Get element inner text
+ */
+export function getTextFromElement(element: React.ReactElement): string {
```
zombieJ, line 67: This will not work as expect since React support HOC:
```tsx
const My = () => 'Hello World';
```

`getTextFromElement` will get nothing.

zombieJ: And this will also break in some locale case:
```tsx
const Counter = ({ unit }) => {
  const counter = useContext(parentContext);
  return `${counter} ${unit}s missing`;
};
```

get `unit` instead of `3 units missing`

zombieJ: Another case is customize component use another prop instead of `children`:

```tsx
<Badge status="success" text="Success" />
```

zombieJ:
Could you help to check using `aria-labelledby` instead?

components/auto-complete/__tests__/__snapshots__/demo-extend.test.ts.snap
yoyo837: 这个应该是demo的自己的锅

kiner-tang: 感觉可以改一下 demo

liangkuku:
> Could you help to check using `aria-labelledby` instead?

Sure,but I did some research and found that `aria-labelledby` requires the element's id as an identifier.If the element does not have an id, it can pose a problem when trying to identify it.

PR #47874, opened 2024-03-14
fix: DatePicker itemHeight calculation

### LINKED ISSUE #47550: Date Picker with mode 'multiple' computes wrong heights
### Reproduction link

[CodeSandbox](https://codesandbox.io/p/sandbox/multiple-antd-5-14-1-forked-sq3ygc)

### Steps to reproduce

1. Overwrite ConfigProvider theme property with `theme: { token: { controlHeight: 40 }`
2. Use DatePicker with property multiple to true
3. See the input height is 46px instead of 40px

### What is expected?

DatePicker input with default `size='middle'` should be the height of the value passed in the theme token for control heights. In this scenario the height should be 40px and NOT 46px
Also, I would expect the placeholder to work, doesn't seem to work. Also included that in the repro sandbox

The height of date picker in mode multiple is based on some padding we apply and the height   line-height of the tags items or simply the empty picker selector div. That padding   height/line-height is wrong and gives the wrong height (46px instead of the 40px expected from the `controlHeight` token)

| Environment | Info |
| --- | --- |
| antd | 5.14.1 |
| React | React 18.2.0 |
| System | MacOS Ventura version 13.4 |
| Browser | Version 119.0.6045.199 (Official Build) (arm64) |

---
It is an exciting update cause the UX of our workaround to have a multi date picker wasn't as good. This is a small bump in the road and the style is quite complex to understand and I wouldn't be confortable changing values there. Otherwise I would be down to open a PR

**This is explained only in the context of components with default size type ('middle') but the bug applies to other sizes too.**
**Also, NOTE that this is BEFORE we select value(s). After that it get more complex cause there is also the styles applied to the tags that will apply a wrong height and mess with the control height**

This might not be the exact bug and all the cases the height is broken but I think it gives a good start point to understand why this bug happens.

Hierarchy important to understand the height bug is only the 2 top level DIVs: `div (ant-picker-multiple, other clases...) > div (ant-picker-selector)`
With Antd default configuration this is what happens:
1. `.ant-picker-multiple` applies (basically the same for padding-top and bottom): 
`padding-top: max(calc((var(--ant-control-height) - calc(var(--ant-font-height)   2px) - var(--ant-line-width) * 2)/2),0px);` with `controlHeight` of `24px`, `fontHeight` of `22px` and `lineWidth` of `1px` => `3px` (Total of `6px` top   bottom)
2. `.ant-picker-multiple .ant-picker-selector:after` applies a `line-height`:
`line-height: var(--ant-date-picker-multiple-item-height);` Where this value is equal the `controlHeightSM` value (`24px`). [Line of code assigning the itemHeight value](https://github.com/ant-design/ant-design/blob/31b421d739d0c8eaa261117d214227062c12a6a7/components/date-picker/style/token.ts#L158)

**Total height with antd token values** => `24   6 = 30px` which is the expected `controlHeight` since that design token is set to `30px` by default
Now, in our case we define `controlHeight: 40` and `controlHeightSM: 32` which will give a height of `48px` which isn't equal to the expected `40px`

zombieJ:

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link

resolve https://github.com/ant-design/ant-design/issues/47550

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   Fix DatePicker component height not correct when customize `controlHeight`.         |
| 🇨🇳 Chinese |   修复 DatePicker 配置自定义 `controlHeight` 时，展示高度不正确的问题。        |

### DIFF
--- a/components/date-picker/style/token.ts
+++ b/components/date-picker/style/token.ts
@@ -139,7 +139,9 @@ export const initPickerPanelToken = (token: TokenWithCommonCls<GlobalToken>): Pi
 };
 
 export const initPanelComponentToken = (token: GlobalToken): PanelComponentToken => {
-  const { colorBgContainerDisabled, controlHeightSM, controlHeightLG } = token;
+  const { colorBgContainerDisabled, controlHeight, controlHeightSM, controlHeightLG, paddingXXS } =
+    token;
+
   return {
     cellHoverBg: token.controlItemBgHover,
     cellActiveWithRangeBg: token.controlItemBgActive,
@@ -155,8 +157,9 @@ export const initPanelComponentToken = (token: GlobalToken): PanelComponentToken
     withoutTimeCellHeight: controlHeightLG * 1.65,
     multipleItemBg: token.colorFillSecondary,
     multipleItemBorderColor: 'transparent',
-    multipleItemHeight: controlHeightSM,
-    multipleItemHeightLG: token.controlHeight,
+    multipleItemHeight: controlHeight - paddingXXS * 2,
+    multipleItemHeightSM: controlHeightSM - paddingXXS * 2,
+    multipleItemHeightLG: controlHeightLG - paddingXXS * 2,
     multipleSelectorBgDisabled: colorBgContainerDisabled,
     multipleItemColorDisabled: token.colorTextDisabled,
     multipleItemBorderColorDisabled: 'transparent',
--- a/components/date-picker/style/token.ts
+++ b/components/date-picker/style/token.ts
@@ -142,6 +142,8 @@ export const initPanelComponentToken = (token: GlobalToken): PanelComponentToken
   const { colorBgContainerDisabled, controlHeight, controlHeightSM, controlHeightLG, paddingXXS } =
     token;
 
+  console.log('??!!!', controlHeightLG, paddingXXS * 2);
+
   return {
     cellHoverBg: token.controlItemBgHover,
     cellActiveWithRangeBg: token.controlItemBgActive,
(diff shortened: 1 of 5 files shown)


components/date-picker/style/token.ts
```diff
@@ -139,7 +139,11 @@ export const initPickerPanelToken = (token: TokenWithCommonCls<GlobalToken>): Pi
 };
 
 export const initPanelComponentToken = (token: GlobalToken): PanelComponentToken => {
-  const { colorBgContainerDisabled, controlHeightSM, controlHeightLG } = token;
+  const { colorBgContainerDisabled, controlHeight, controlHeightSM, controlHeightLG, paddingXXS } =
+    token;
+
+  console.log('??!!!', controlHeightLG, paddingXXS * 2);
```
crazyair, line 145: 还在调试阶段啊？

components/select/style/multiple.ts
```diff
@@ -20,6 +20,44 @@ type SelectItemToken = Pick<
...
+ * Get multiple selector needed style. The calculation:
+ *
+ * ContainerPadding = BasePadding - ItemMargin
+ *
+ * Border:                    ╔═══════════════════════════╗                 ┬
+ * ContainerPadding:          ║                           ║                 │
+ *                            ╟───────────────────────────╢     ┬           │
+ * Item Margin:               ║                           ║     │           │
+ *                            ║             ┌──────────┐  ║     │           │
+ * Item(multipleItemHeight):  ║ BasePadding │   Item   │  ║  Overflow  Container(ControlHeight)
```
crazyair, line 33: 这个厉害了

components/date-picker/style/token.ts
li-jia-nan, line 145: console

li-jia-nan (approved):

PR #48499, opened 2024-04-16
Feat/droprdown accessibility

### LINKED ISSUE #48132: Dropdown is missing critical accessibility attributes

Currently the Dropdown component does not meet WAI-ARIA standards for Menu Button Pattern (https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/):

Specifically:
- The element with role button has aria-haspopup set to either menu or true.
- When the menu is displayed, the element with role button has aria-expanded set to true. When the menu is hidden, aria-expanded is set to false.
- The element that contains the menu items displayed by activating the button has role menu.
Optionally, the element with role button has a value specified for aria-controls that refers to the element with role menu.

And furthermore from the Menu and MenuBar Pattern (https://www.w3.org/WAI/ARIA/apg/patterns/menubar/), though neither of these roles are used in Dropdown, it can be reasonably deduced that in Dropdown, the element with role=button should accept an aria-label prop to assist those with screen readers properly identifying it.

The Dropdown component currently does accept an aria-label prop, but it does not seem to actually get applied anywhere in the DOM structure.

Most of the above is satisfied by internal definitions, but the Dropdown component must expose an aria-label prop that maps to the aria-label attribute on the element that has role=button.

KirilCycle:
### 🤔 This is a ...
- [x] accessibility improvement

### 🔗 Related issue link
Fix #48132

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  I added all needed attributes to dropdown components, and now dropdown menu can be opened by space or enter,  keypress |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/dropdown/dropdown.tsx
+++ b/components/dropdown/dropdown.tsx
@@ -199,19 +199,10 @@ const Dropdown: CompoundedComponent = (props) => {
       child.props.className,
     ),
     disabled,
-    onFocus: (event: Event) => {
-      const targetElement = event.target as HTMLElement;
-
-      const handleKeyPress = (e: KeyboardEvent) => {
+    onKeyDown: (e: KeyboardEvent) => {
         if (e.key === 'Enter' || e.code === 'Space') {
-          targetElement.removeEventListener('keypress', handleKeyPress);
           onInnerOpenChange(true)
-        }
       };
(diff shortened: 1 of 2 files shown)


KirilCycle (commented):
Is that okay that dropdown need additional tab press to focus on first menu element ?

PR #48886, opened 2024-05-12
refactor: refactor Drawer loading

li-jia-nan:

### 🤔 This is a ...

- [ ] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [x] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | refactor: refactor Drawer loading |
| 🇨🇳 Chinese | 重构 Drawer 组件的 loading 状态，将 Spin 替换为 Skeleton，和 Modal 组件保持一致 |


Enigama:
It might be not user friendly to have header while `loading` is true, for instance: drawer gets title and content from the server and  while we are waiting for response from server we would see header without title and with [x] button.
image

li-jia-nan:
> It might be not user friendly to have `header` while `loading` is true, for instance: drawer gets title and content from the server and while we are waiting for response from server we would see header without title and with [x] button

I think this is the expected behavior, and I support the `loading` prop of the Modal component. The Drawer should be consistent with the behavior of Modal components. If you don't need header, You can customize the header:

```tsx
<Drawer
  open={open}
  loading={isLoading}
  closable={!isLoading}
  title={isLoading? null : <span>Loading Drawer</span>}
>
  <p>Some contents...</p>
  <p>Some contents...</p>
  <p>Some contents...</p>
</Drawer>
```

zombieJ (approved):

kiner-tang (approved):

zombieJ:
> It might be not user friendly to have header while loading is ...

In most case, Title is no need to load by async. But yes, it can still use spin as loading if needed.

PR #49160, opened 2024-05-31
feat: Add noDynamicTableData props for which the entire data can be rendered even if the table data source goes beyond the pageSize of the page

seung-juv:
Add noDynamicTableData props for which the entire data can be rendered even if the table data source goes beyond the pageSize of the page

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Support noDynamicTableData in Table |
| 🇨🇳 Chinese | 支持表中的 noDynamicTableData |


afc163:
How about set pageSize to Number.Infinity?

seung-juv:
> How about set pageSize to Number.Infinity?

I needed this function is that I wanted to show the data by combining two Rows on a table. There are 10 pieces of data on a page, but through a separate logic, it was hoped that there would be 20 pieces of data on 10 Rows to make 1 Row, but the dynamicTableData data showed only 10 pieces of data. As you said, if the value of pageSize is changed to Number.Infinity, the function of paging is not expected to work properly.

seung-juv:
> > How about set pageSize to Number.Infinity?
> 
> I needed this function is that I wanted to show the data by combining two Rows on a table. There are 10 pieces of data on a page, but through a separate logic, it was hoped that there would be 20 pieces of data on 10 Rows to make 1 Row, but the dynamicTableData data showed only 10 pieces of data. As you said, if the value of pageSize is changed to Number.Infinity, the function of paging is not expected to work properly.

As you mentioned, the function is not supported, so we did not expose the paging to the Table Component by working with pageSize as Number.Infinity. And even though a separate paging component was called to handle the logic, if the function exists, it is expected to be easy to handle not only in this situation but also in any situation.

afc163:
> show the data by combining two Rows on a table

What does the table look like?

seung-juv:
> > show the data by combining two Rows on a table
> 
> What does the table look like?

image

As shown in the image above, there are 1 order and n ordered products on the list to briefly explain. To distribute this to the list and to select each ordered product, there were cases where the number of pages did not match the page size because only n ordered products must be judged by one Row to not be sold out in Pagination.

PR #49172, opened 2024-06-01
feat: control Drawer close button placement

Enigama:

### 🤔 This is a ...

- [x] New feature
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 💡 Background and solution
It would be nice to have option to control close button in drawer, now it's possible only by adding extra css. So solution is simple, add `closePlacement` prop to a drawer, property can have: `left` default or `right` value.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   Add property to a Drawer to control close button placement        |
| 🇨🇳 Chinese |     在抽屜中新增屬性以控制關閉按鈕的位置      |


Enigama:
@li-jia-nan can we merge it or something?

yoyo837:
Rebase again please.

Enigama:
@yoyo837 H'm I sync my fork

Enigama:
@yoyo837 Have I done something wrong?

Enigama:
I'll create a new one, i merged main into feature.

zombieJ:
Sorry for delay. In my opinion, there shouldn't be too many configurations for `xxxPlacement` props, as their implementation can become overly fragmented. A better approach would be to use semantic `classNames` to allow developers to customize styles themselves. e.g. to do like Modal:

```tsx
<Drawer classNames={{ header: 'my-header' }} />
```

```less
.my-header button {
  order: 1;
}
```

Enigama:
> Sorry for delay. In my opinion, there shouldn't be too many configurations for `xxxPlacement` props, as their implementation can become overly fragmented. A better approach would be to use semantic `classNames` to allow developers to customize styles themselves. e.g.:
> 
> 
> 
> ```tsx
> 
> <Drawer classNames={{ header: 'my-header' }} />
> 
> ```
> 
> 
> 
> ```less
> 
> .my-header button {
> 
>   order: 1;
> 
> }
> 
> ```
> 
> 
> 
> online demo: https://stackblitz.com/edit/react-csca2w?file=demo.tsx,index.css
> 
> 
> 
> In addition, you can use ConfigProvider to handle all the drawer:
> 
> 
> 
> ```tsx
> 
> <ConfigProvider drawer={{ classNames: { header: 'my-header' } }} />
> 
> ```

Thanks for example with config provider didn't think in that way. Also example with 'order:1' won't help when extra prop, so solution for it is use 'position: absolute' and move to right corner

PR #49836, opened 2024-07-11
Revert "type(token): height / width support string | number"

li-jia-nan:
Reverts ant-design/ant-design#49810


li-jia-nan:
这个放在 feature 分支，先回滚

afc163 (approved):

PR #51318, opened 2024-10-20
feat(select): support overriding default component behavior for the onBlur event in tags mode

aojunhao123:
### 🤔 This is a ...

- [x] 🆕 New feature

### 🔗 Related Issues
- close #51149 
- close https://github.com/react-component/select/issues/1075
- close https://github.com/react-component/select/issues/952

### 💡 Background
In the current tags mode of the Select component, the default behavior for the blur event after user input is to remove any spaces in the input value and automatically add the processed value to the Select. However, in certain scenarios, developers may want to override this default behavior of the component.

### 💡 Solution
In this update, two new parameters have been added to the Select component: onBlurRemoveSpace and onBlurAddValue, which are used to override the two default behaviors mentioned above, respectively.

**for more detail**, **trace to**  https://github.com/react-component/select/pull/1081

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Added onBlurRemoveSpace and onBlurAddValue to allow overriding the default onBlur behavior in tags mode  |
| 🇨🇳 Chinese | 新增 onBlurRemoveSpace 和 onBlurAddValue，支持覆盖标签模式下的默认 onBlur 行为  |


components/select/index.tsx
```diff
@@ -120,6 +130,8 @@ const InternalSelect = <
     tagRender,
     maxCount,
     prefix,
+    onBlurRemoveSpaces = true,
+    onBlurAddValue = true,
```
afc163, line 134:
```jsx
blurBehavior={
  trim: boolean,
  createTag: boolean,
}
```

这样的 API 如何？

components/select/index.zh-CN.md
```diff
@@ -125,6 +125,8 @@ return (
 | value | 指定当前选中的条目，多选时为一个数组。（value 数组引用未变化时，Select 不会更新） | string \| string\[] \| <br />number \| number\[] \| <br />LabeledValue \| LabeledValue\[] | - |  |
 | variant | 形态变体 | `outlined` \| `borderless` \| `filled` | `outlined` | 5.13.0 |
 | virtual | 设置 false 时关闭虚拟滚动 | boolean | true | 4.1.0 |
+| onBlurRemoveSpaces | 标签形式下，失去焦点时是否移除空格，仅在 `mode="tags"` 时生效 | boolean | true |  |
```
afc163, line 128: 另外这个 onBlurRemoveSpaces 我没有太 GET 到，现在我测试下来是保留空格的？

图片

components/select/index.tsx
aojunhao123: cc @zombieJ

aojunhao123: 不过这个pr还需要重新调整一下rc侧的逻辑，第一版实现还有点问题。调整完再review吧~

afc163:
Thanks for the contribution. Closing this PR because the rc-select side of the design was already identified as incomplete, and the maintainer feedback around the `blurBehavior` API has remained unresolved. A new proposal should first settle the upstream behavior and API contract.

PR #51696, opened 2024-11-19
chore: do not downgrade syntaxes when compiling

kermanx:
### 🤔 This is a ...

- [x] 📦 Bundle size optimization

### 🔗 Related Issues

- https://github.com/ant-design/ant-design/issues/39900#issuecomment-2482080737

### 💡 Background and Solution

Transforming syntax has many drawbacks:

- Much longer bundle size
- Impossible to analyze for Minifiers and Tree-shakers

And in (at least ESM) builds, users should be able to decide whether downgrade the syntax or not.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | No longer downgrade syntaxes, for smaller bundle size and better optimization |
| 🇨🇳 Chinese | 不再降级语法，以减小包体积并便于优化 |

### DIFF
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es6",
+    "target": "ESNext",
     "jsx": "react",
     "jsxFactory": "React.createElement",
     "jsxFragmentFactory": "React.Fragment",


afc163:
得 diff 一下 antd.js 看看差异点。

li-jia-nan:
这个要怎么 diff 呢？是需要 build 一下看构建产物吗？还是有什么工具链可以检测？

afc163:
手动看一下先。

package.json
```diff
@@ -158,7 +158,7 @@
   "devDependencies": {
     "@ant-design/compatible": "^5.1.3",
     "@ant-design/happy-work-theme": "^1.0.0",
-    "@ant-design/tools": "^18.0.2",
+    "@ant-design/tools": "18.0.3-alpha.1",
```
afc163, line 161:
```suggestion
    "@ant-design/tools": "18.1.3-alpha.2",
```

afc163:
图片

感觉不太行。

package.json
```diff
@@ -158,7 +158,7 @@
   "devDependencies": {
     "@ant-design/compatible": "^5.1.3",
     "@ant-design/happy-work-theme": "^1.0.0",
-    "@ant-design/tools": "^18.0.2",
+    "@ant-design/tools": "18.1.3-alpha.2",
```
afc163, line 161:
```suggestion
    "@ant-design/tools": "^18.0.2",
```

KermanX:
> It will break old browser like IE11.

That's true. However, I think at least the ESM format should not be downgraded, because it is usually bundled by downstream apps. The downstream apps can still downgrade the syntax to support IE11, but it is very hard to reverse this process if the apps want a smaller size.

afc163:
> I think at least the ESM format should not be downgraded

You are right, could this PR do that?

afc163:
Please send this PR to next branch for antd v6: https://github.com/ant-design/ant-design/discussions/51919

thinkasany:
@afc163 , Already checkout to the next branch.

afc163:
## size-limit report 📦 
| Path                            | Size                     |
| ------------------------------- | ------------------------ |
| `dist/antd.min.js`              | 487.49 KB (-11.62 KB 🔽) |
| `dist/antd-with-locales.min.js` | 555.56 KB (-12.14 KB 🔽) |

afc163:
Resolved in https://github.com/ant-design/ant-design/discussions

Issue #52077, opened 2024-12-20
Input.Search should render an HTML input of type "search"


dev-nicolaos:
### Reproduction link

[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/antd-search-input-wrong-input-type-q72pr6)

### Steps to reproduce

1. Render an instance of `<Input.Search>`
1. Inspect the rendered `<input>` element with your browser's dev tools

### What is expected?

The rendered HTML `<input>` element has its `type` attribute set to `"search"`, which gives the input an implicit `role` of `"searchbox"`.

### What is actually happening?

The rendered HTML `<input>` element has its `type` attribute set to `"text"`, which gives the input an implicit `role` of `"textbox"`.

| Environment | Info |
| --- | --- |
| antd | 5.22.5 |
| React | 18 |
| System | NA |
| Browser | NA |

PR #52083, opened 2024-12-21
feat: update input.search type (#52077)

Kaikiat1126:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [x] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

The rendered HTML <input> element has its type attribute set to "text", which gives the input an implicit role of "textbox".

fix https://github.com/ant-design/ant-design/issues/52077

### 💡 Background and Solution

add "search" type

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  add "search" type for Input.Search   |
| 🇨🇳 Chinese |  为input.search元素补充"search" type  |

### DIFF
--- a/components/config-provider/__tests__/__snapshots__/components.test.tsx.snap
+++ b/components/config-provider/__tests__/__snapshots__/components.test.tsx.snap
@@ -16125,7 +16125,6 @@ exports[`ConfigProvider components Input configProvider 1`] = `
       >
         <input
           class="config-input config-input-outlined"
-          role="searchbox"
           type="search"
           value=""
         />
@@ -16226,7 +16225,6 @@ exports[`ConfigProvider components Input configProvider componentDisabled 1`] = 
         <input
           class="config-input config-input-disabled config-input-outlined"
           disabled=""
-          role="searchbox"
           type="search"
           value=""
         />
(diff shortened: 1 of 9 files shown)


afc163 (dismissed):

components/config-provider/__tests__/__snapshots__/components.test.tsx.snap
```diff
@@ -16125,7 +16125,8 @@ exports[`ConfigProvider components Input configProvider 1`] = `
       >
         <input
           class="config-input config-input-outlined"
-          type="text"
+          role="searchbox"
```
afc163, line 16128: 图片

Including role="searchbox" when the form is a search and the label indicates the element is a search may result in assistive technology announcing something along the lines of "search search this site search box", which is redundant. **The inclusion of role="searchbox" is not necessary.**

https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/searchbox_role

afc163 (approved):

PR #52750, opened 2025-02-11
fix(Calendar): disable select item in year and month selector according to disabledDate

### LINKED ISSUE #52658: Calendar disable future dates but select years and months aren't disabled
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/sandbox/37c43f)

### Steps to reproduce

- create a calendar
- assign disabled dates end of the current day

### What is expected?

future years and months MUST NOT be selectable from select

future years and months ARE selectable from select

| Environment | Info |
| --- | --- |
| antd | 5.23.3 |
| React | 18 |
| System | mac |
| Browser | chrome |

typenoob:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

fix #52658

### 💡 Background and Solution

**Background:**

`Calendar` component only set `disabledDate` to its picker panel, while its header don't hold the attribute. As a result, `YearSelect` and `MonthSelect` won't be affected with `disabledDate`.

**Solution:**

- Extended attributes of `CalendarHeader` with `disabledDate`.

- Disabled part of select item according to this attribute both in `YearSelect` and `MonthSelect`. The logic is similar to [YearPanel](https://github.com/react-component/picker/blob/23ff742b9d557f522e60ea4dcc40066449a10f78/src/PickerPanel/YearPanel/index.tsx#L62) and [MonthPanel](https://github.com/react-component/picker/blob/23ff742b9d557f522e60ea4dcc40066449a10f78/src/PickerPanel/MonthPanel/index.tsx#L56) in rc-picker.

- Added test case relate to this pr.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   fix Calendar won't disable select item according to `disabledDate` in year and month selector      |
| 🇨🇳 Chinese |    修复Calendar年份月份选择器中不会根据`disabledDate`禁用选择项的问题       |

### DIFF
--- a/components/calendar/Header.tsx
+++ b/components/calendar/Header.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 import { useContext, useMemo } from 'react';
 import type { GenerateConfig } from 'rc-picker/lib/generate';
-import type { Locale } from 'rc-picker/lib/interface';
+import type { DisabledDate, Locale, PanelMode } from 'rc-picker/lib/interface';
 
 import { FormItemInputContext } from '../form/context';
 import { Button, Group } from '../radio';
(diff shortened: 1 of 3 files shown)


afc163:
Thanks for the contribution. Closing because the year/month disabling logic only checks period endpoints, so a period containing valid dates can still be disabled. This needs a different range-validity model before it can be merged.

PR #52775, opened 2025-02-13
feat:For Switch  add two apis checkedValue and uncheckedValue

zsqsmart:

### 🤔 This is a ...

- [x] 🆕 New feature

### 🔗 Related Issues

### 💡 Background and Solution

    Currently, the Switch only supports the boolean type. In fact, there are times when other types such as numbers (e.g., 0 and 1) and strings are used to represent the checked and unchecked states. Therefore, the function of adding custom values is required.
    Add two attributes, checkedValue and uncheckedValue, which represent the values for the checked and unchecked states respectively.

### 📝 Change Log

| Language   | Changelog                                                                                                                             |
| ---------- | ------------------------------------------------------------------------------------------------------------------------------------- |
| 🇺🇸 English | For the Switch, custom values can be added. You can set the checked value and unchecked value through `checkedValue` and `uncheckedValue` |
| 🇨🇳 Chinese | Switch 增加自定义值，可以通过 `checkedValue` 和 `uncheckedValue` 设置选中值和未选中值 |

### DIFF
--- a/components/switch/demo/custom-value.tsx
+++ b/components/switch/demo/custom-value.tsx
@@ -1,0 +1,39 @@
+import React, { useState } from 'react';
+import { Space, Switch } from 'antd';
+
+const App: React.FC = () => {
+  const [value1, setValue1] = useState(true);
+  const [value2, setValue2] = useState(1);
+  const [value3, setValue3] = useState('hello');
+
+  const onChange1 = (value: boolean) => {
+    setValue1(value);
+  };
+
+  const onChange2 = (value: number) => {
+    setValue2(value);
+  };
+
+  const onChange3 = (value: string) => {
+    setValue3(value);
+  };
+
+  return (
+    <Space direction="vertical">
+      <Space>
+        <Switch value={value1} onChange={onChange1} />
+        <span>value: {value1.toString()}</span>
+      </Space>
+      <Space>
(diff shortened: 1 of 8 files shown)


zombieJ:
直接 HOC 一个 Switch 就行了？看起来不像是一个常见的需求

components/switch/demo/custom-value.tsx
```diff
@@ -0,0 +1,39 @@
...
+  };
+
+  return (
+    <Space direction="vertical">
+      <Space>
+        <Switch value={value1} onChange={onChange1} />
+        <span>value: {value1.toString()}</span>
+      </Space>
+      <Space>
+        <Switch value={value2} checkedValue={1} uncheckedValue={0} onChange={onChange2} />
```
afc163, line 28: 这一层应该交给业务自行去转换，Switch 就是 boolean 的逻辑。

zsqsmart: 我觉得switch不仅仅是boolean的逻辑，应该是两个枚举值，业务自己转换太麻烦，需要转成boolean，然后再转回去，如果是ts，还涉及到类型转换，会有太多的冗余代码，所以我觉得很有必要增加自定义值的功能。
参考 element-plus：
image

afc163: Switch 是非常原子的组件，我认为逻辑应该尽可能保持简单。

PR #52997, opened 2025-02-27
fix: includes aria-description in table header when column is sortable

jon-cullison:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

### 💡 Background and Solution

Screen reader users have no way of knowing if the currently focused table column header is sortable or not.

This solution adds `aria-description="Sortable column"` to sortable column headers. This will make screen readers announce that columns are sortable.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Includes aria-description in table header when column is sortable |
| 🇨🇳 Chinese | Includes aria-description in table header when column is sortable |


jon-cullison:
There is a bug in React v17 that is causing the snapshot test failures in the legacy tests. React v17 does not recognize `aria-current` as a valid aria prop.

I can't fix the snapshot because then other tests will fail. Please advise how to proceed.

thinkasany:
> There is a bug in React v17 that is causing the snapshot test failures in the legacy tests. React v17 does not recognize `aria-current` as a valid aria prop.
> 
> I can't fix the snapshot for the legacy tests because then the tests will fail for other versions. Please advise how to proceed.

Considering compatibility, this function should not be provided to lower versions, but for better a11y, it seems that 18 needs this feature. Not sure if this api can help you.
image
image

afc163:
@mellis481 Could you help to review this one?

mellis481:
> @mellis481 Could you help to review this one?

@afc163 This PR looks good to me. This change is very important for screen readers users because, without it, they will not know that a column is sortable.

Looking at how [react-aria](https://react-spectrum.adobe.com/react-aria/Table.html#sorting) solves this problem, they add a visually-hidden element with the contents "sortable column" to the page and programmatically link it (via `aria-describedby`) to the `th` element in sortable columns. What is included in this PR I believe is a better solution because using `aria-description` on the header forgoes having to create one or more additional elements for each table instance.

mellis481:
@thinkasany That seems like a reasonable suggestion. Any idea when v6 will be released?

components/locale/en_US.ts
```diff
@@ -34,6 +34,7 @@ const localeValues: Locale = {
     triggerDesc: 'Click to sort descending',
     triggerAsc: 'Click to sort ascending',
     cancelSort: 'Click to cancel sorting',
+    sortableColumn: 'Sortable column',
```
mellis481, line 37: I would rename this prop to `sortableColumnHeaderAriaDescription`.

thinkasany:
> That seems like a reasonable suggestion. Any idea when v6 will be released?

The release depends on the completion time of the tasks. It will probably take a few months.
https://github.com/ant-design/ant-design/issues/52115

jon-cullison:
Closing this PR to reopen another in the next branch.

jon-cullison:
Opened https://github.com/ant-design/ant-design/pull/53087 in the next branch.

PR #53087, opened 2025-03-06
fix: includes aria-description in table header when column is sortable

jon-cullison:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 💡 Background and Solution

Screen reader users have no way of knowing if the currently focused table column header is sortable or not.

This solution adds `aria-description="Sortable column"` to sortable column headers. This will make screen readers announce that columns are sortable.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  ⌨️ Improve Table a11y behavior by adding `aria-description` when column is sortable. |
| 🇨🇳 Chinese |  ⌨️ 优化 Table `column` 为 `sortable` 时的 `aria-description` 可访问性属性。 |


jon-cullison:
Recreated the work done in https://github.com/ant-design/ant-design/pull/52997 in the next branch

components/table/interface.ts
```diff
@@ -60,6 +60,7 @@ export interface TableLocale {
   triggerDesc?: string;
   triggerAsc?: string;
   cancelSort?: string;
+  sortableColumnHeaderAriaDescription?: string;
```
zombieJ, line 63: Seems this API not follow antd naming standard. It's better to be:

```tsx
const columns = [{
  sorter: {
    'aria-description': 'Sort for name',
  },
}];
```

jon-cullison: I'm not sure what you mean. I'm happy to change the name to something else to meet antd's naming standard, but I don't know how you intended me to use the code you suggested.

mellis481: @zombieJ The prop/value being discussed is not something consumers are going to be setting/overriding.

mellis481:
@zombieJ Does anything else need to be done with this PR or can it be merged?

components/table/interface.ts
zombieJ:
```tsx
<Table
  columns={[{
    sorter: {
      'aria-description': 'Sort for name',
    },
  }]}
/>
```

zombieJ: or something not need add api version:

```tsx
onHeaderCell = () => ({
  'aria-description': 'Sort for name',
});
```

mellis481: @zombieJ I think there may be some confusion here. The prop that @jon-cullison has added is not one that consumer's should set, nor have the ability to override. Setting the generic string "sortable column" as the `aria-description` simply ensures that a sortable column is announced as sortable. It does not change the current behavior where the column header is announced. If the consumer wants to change how the column is announced, they can change the `title` prop.

image

@jon-cullison has added this code to `useSorter` where the rest of the sort-related functionality exists. He's also defined the text in the language files.

zombieJ: @mellis481 , OK. So we can do like this to make it as common aira props:

```tsx
// components/locale/index.tsx
// ...

export interface Locale {
  locale: string;
  // ...
  Table?: TableLocale;
  // ...
  a11y?: {
    sortable?: string;
  }
}
```

aojunhao123: Looks good! I agree with this approach for handling a11y as an optional property in the Locale interface. 👍

mellis481: Sounds good.

In that case, @jon-cullison, I recommend the very-descriptive name `sortableColumnHeaderAriaDescription`. `sortable` is not descriptive enough.

components/locale/en_US.ts
```diff
@@ -9,6 +9,9 @@ const typeTemplate = '${label} is not a valid ${type}';
 
 const localeValues: Locale = {
   locale: 'en',
+  a11y: {
+    sortableColumnHeaderAriaDescription: 'sortable column',
```
zombieJ, line 13: `sortable` and make a common desc that mean support sort. Which we can use same locale in other compnent

afc163: @zombieJ `sortable column` can only be used in Table.

zombieJ: > @zombieJ sortable column can only be used in Table.

```tsx
// EN
ally: {
  sortable: 'Sortable',
}

// CN
ally: {
  sortable: '可排序',
}
```

afc163, line 13:
```suggestion
    sortable: 'Sortable',
```

afc163 (approved):
LGTM

PR #53415, opened 2025-04-05
fix: Input OTP breaks with type number and mask

rajankonar:

### 🤔 This is a bugfix for Input.OTP, it breaks when mask is used with type number field.

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

> - Input.OTP breaks in UI, when set with type as number and mask.
> - For example : https://codesandbox.io/p/sandbox/otp-antd-5-24-6-forked-23m9n3

https://github.com/user-attachments/assets/4cbf9d86-9405-4b0e-b5ed-75ddbd8fb5d0

### 💡 Background and Solution

> - Mask doesnt work on number field, so i have handled it using css based approach
> - Type will be single source of truth, in the new approach we are not changing its type dynamically

Demo of fix :

https://github.com/user-attachments/assets/2e3761eb-f9b0-43e5-b202-e8a5da078c07

Code :

Screenshot 2025-04-05 at 5 15 14 PM

### 📝 Change Log

> - None


rajankonar:
@zombieJ @li-jia-nan  Please review.

rajankonar:
Hello @li-jia-nan, Visual Regression Diff was passed earlier, after updating branch, it got failed and my changes are related to input, it should not affect UI of list-vertical.

components/input/OTP/OTPInput.tsx
```diff
@@ -59,11 +65,21 @@ const OTPInput = React.forwardRef<InputRef, OTPInputProps>((props, ref) => {
     syncSelection();
   };
 
+  const setInputType = () => {
```
zombieJ, line 68: `useMemo` instead of function call

components/input/OTP/OTPInput.tsx
```diff
@@ -59,11 +65,21 @@ const OTPInput = React.forwardRef<InputRef, OTPInputProps>((props, ref) => {
...
+    let type = restProps.type;
+    if (mask === true) {
+      type = 'password';
```
zombieJ, line 71: User passed props `type` should be single source of truth. We should not override this

components/input/OTP/OTPInput.tsx
```diff
@@ -19,7 +19,13 @@ const OTPInput = React.forwardRef<InputRef, OTPInputProps>((props, ref) => {
   const internalValue = value && typeof mask === 'string' ? mask : value;
 
   const onInternalChange: React.ChangeEventHandler<HTMLInputElement> = (e) => {
-    onChange(index, e.target.value);
+    if (
+      e.target.value === '' ||
+      restProps.type !== 'number' ||
+      (restProps.type === 'number' && !isNaN(parseInt(e.target.value)))
```
zombieJ, line 25: This should deps on the follow mergedType logical.

components/input/OTP/OTPInput.tsx
rajankonar: Sure @zombieJ

components/input/OTP/OTPInput.tsx
rajankonar: Thanks for review @zombieJ will change it

components/input/OTP/OTPInput.tsx
```diff
@@ -14,18 +16,36 @@ export interface OTPInputProps extends Omit<InputProps, 'onChange'> {
...
+      const escapedMaskValue = maskValue
+        .toString()
+        .replace(/&/g, '&amp;')
+        .replace(/</g, '&lt;')
+        .replace(/>/g, '&gt;')
+        .replace(/"/g, '&quot;')
+        .replace(/'/g, '&apos;');
+      const inputBackgroundStyle = inputRef.current.input.style;
+      if (e.target.value && restProps.type === 'number' && typeof mask === 'string') {
+        inputBackgroundStyle.background = `url("data:image/svg+xml,<svg xmlns='http://www.w3.org/2000/svg' width='20' height='20'><text x='50%' y='50%' dominant-baseline='central' text-anchor='middle'>${escapedMaskValue}</text></svg>") center center / 26px no-repeat #fff`;
```
zombieJ, line 41: It's not allow to modify DOM style directly. It's not React way.

components/input/OTP/OTPInput.tsx
```diff
@@ -14,18 +16,36 @@ export interface OTPInputProps extends Omit<InputProps, 'onChange'> {
...
 
-  const internalValue = value && typeof mask === 'string' ? mask : value;
+  // ========================== Ref ===========================
+  const inputRef = React.useRef<InputRef>(null);
+  React.useImperativeHandle(ref, () => inputRef.current!);
 
+  // ========================= Input ==========================
   const onInternalChange: React.ChangeEventHandler<HTMLInputElement> = (e) => {
+    if (inputRef.current?.input && maskValue) {
+      const escapedMaskValue = maskValue
```
zombieJ, line 32: Why not just create a `:after` element to show the mask?

components/input/OTP/OTPInput.tsx
```diff
@@ -65,13 +85,16 @@ const OTPInput = React.forwardRef<InputRef, OTPInputProps>((props, ref) => {
       type={mask === true ? 'password' : 'text'}
       {...restProps}
       ref={inputRef}
-      value={internalValue}
+      value={restProps.type === 'number' ? value : internalValue}
```
zombieJ, line 88: When you use the `mask` element not filled in the input. You can always use `value` now.

components/input/OTP/OTPInput.tsx
rajankonar: Ok will change it.

components/input/OTP/OTPInput.tsx
rajankonar: @zombieJ  `:after`  doesnt work in input, so i went ahead with inputRef background css. 
I can wrap `span` on input then i can add pseudo `:after` to it.

components/input/OTP/OTPInput.tsx
rajankonar: I should use value here? 
Eg : `value={value}`

What i did here was only when type is number, i used value else i kept the functionality as it is. Since internalValue was masking properly with `input:text`

rajankonar:
@zombieJ thanks for the comments.

Changes implemented as per review comments, i will work on test cases and rest of the bot issues, kindly let me know, if this approach is good or i need to do any changes?

components/input/OTP/OTPInput.tsx
```diff
@@ -61,18 +65,30 @@ const OTPInput = React.forwardRef<InputRef, OTPInputProps>((props, ref) => {
...
-      ref={inputRef}
-      value={internalValue}
-      onInput={onInternalChange}
-      onFocus={syncSelection}
-      onKeyDown={onInternalKeyDown}
-      onKeyUp={onInternalKeyUp}
-      onMouseDown={syncSelection}
-      onMouseUp={syncSelection}
-    />
+    <span className={`${prefixCls}-wrapper`}>
```
zombieJ, line 68: `ant-otp-input-wrapper` not `ant-otp-wrapper`

components/input/style/otp.ts
```diff
@@ -14,6 +14,31 @@ const genOTPStyle: GenerateStyle<InputToken> = (token) => {
       flexWrap: 'nowrap',
       columnGap: paddingXS,
 
+      [`${componentCls}-wrapper`]: {
+        position: 'relative',
+        [`${componentCls}-mask-icon`]: {
```
zombieJ, line 19: `pointer-events: none`

components/input/style/otp.ts
```diff
@@ -14,6 +14,31 @@ const genOTPStyle: GenerateStyle<InputToken> = (token) => {
...
+          top: '50%',
+          right: '50%',
+          transform: 'translate(50%, -50%)',
+        },
+        [`${componentCls}-mask-input`]: {
+          color: 'transparent',
+          caretColor: 'var(--ant-color-text)',
+          transition: 'none',
+          textAlign: 'center',
+          letterSpacing: '10px',
```
zombieJ, line 31: these style seems no need to be mask input special

rajankonar: Yes, will remove line number 29 to 31, but caretColor is needed, since color is set as transparent.

components/input/__tests__/__snapshots__/otp.test.tsx.snap
```diff
@@ -4,41 +4,77 @@ exports[`Input.OTP rtl render component should be rendered correctly in RTL dire
...
-    size="1"
-    type="text"
-    value=""
-  />
+  <span
+    aria-label="Otp input"
+    class="ant-otp-input-wrapper"
+    role="group"
+  >
+    <input
```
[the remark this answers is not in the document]
rajankonar: @zombieJ any suggestions/help , tried couple of aria attributes, still getting errors.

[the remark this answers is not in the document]
zombieJ: I check the other OTP component, but not found better aria config of this. Maybe just ignore currently.

rajankonar:
@zombieJ Ignoring accesslint bot comments for now, please review and let me now, if any other changes required.

rajankonar:
Apologies for the mess, creating new pull request.

PR #54240, opened 2025-06-30
docs(Tree): add missing props from rc-tree

ofhope:
Adding missing props to the API section in docs. These props are inherited from the rc-tree package https://github.com/react-component/tree/blob/master/src/Tree.tsx

I can verify the EN version reads OK, I can't with the Chinese version 😢 its guess work with AI to translate.

### 🤔 This is a ...

- [ ] 🆕 New feature
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

### 💡 Background and Solution

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |           |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/tree/index.en-US.md
+++ b/components/tree/index.en-US.md
@@ -82,10 +82,15 @@ Common props ref：[Common props](/docs/react/common-props)
 | onDragOver | Callback function for when the onDragOver event occurs | function({event, node}) | - |  |
 | onDragStart | Callback function for when the onDragStart event occurs | function({event, node}) | - |  |
 | onDrop | Callback function for when the onDrop event occurs | function({event, node, dragNode, dragNodesKeys}) | - |  |
+| onDoubleClick | Callback function for when the user double clicks a treeNode | function(event, node) | - |  |
 | onExpand | Callback function for when a treeNode is expanded or collapsed | function(expandedKeys, {expanded: boolean, node}) | - |  |
 | onLoad | Callback function for when a treeNode is loaded | function(loadedKeys, {event, node}) | - |  |
+| onMouseEnter | Callback function for when the mouse enters a treeNode | function({event, node}) | - |  |
+| onMouseLeave | Callback function for when the mouse leaves a treeNode | function({event, node}) | - |  |
 | onRightClick | Callback function for when the user right clicks a treeNode | function({event, node}) | - |  |
 | onSelect | Callback function for when the user clicks a treeNode | function(selectedKeys, e:{selected: boolean, selectedNodes, node, event}) | - |  |
+| onScroll | Callback function for when the tree container scrolls | function(event) | - |  |
+| dropIndicatorRender | Customize the drop indicator when dragging | (props: {dropPosition, dropLevelOffset, indent}) => ReactNode | - |  |
 
 ### TreeNode props
 
--- a/components/tree/index.zh-CN.md
+++ b/components/tree/index.zh-CN.md
@@ -81,10 +81,15 @@ demo:
 | onDragOver | dragover 触发时调用 | function({event, node}) | - |  |
 | onDragStart | 开始拖拽时调用 | function({event, node}) | - |  |
 | onDrop | drop 触发时调用 | function({event, node, dragNode, dragNodesKeys}) | - |  |
+| onDoubleClick | 双击树节点时调用 | function(event, node) | - |  |
 | onExpand | 展开/收起节点时触发 | function(expandedKeys, {expanded: boolean, node}) | - |  |
 | onLoad | 节点加载完毕时触发 | function(loadedKeys, {event, node}) | - |  |
+| onMouseEnter | 鼠标移入节点时触发 | function({event, node}) | - |  |
+| onMouseLeave | 鼠标移出节点时触发 | function({event, node}) | - |  |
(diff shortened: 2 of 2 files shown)


components/tree/index.en-US.md
```diff
@@ -82,10 +82,15 @@ Common props ref：[Common props](/docs/react/common-props)
...
 | onDrop | Callback function for when the onDrop event occurs | function({event, node, dragNode, dragNodesKeys}) | - |  |
+| onDoubleClick | Callback function for when the user double clicks a treeNode | function(event, node) | - |  |
 | onExpand | Callback function for when a treeNode is expanded or collapsed | function(expandedKeys, {expanded: boolean, node}) | - |  |
 | onLoad | Callback function for when a treeNode is loaded | function(loadedKeys, {event, node}) | - |  |
+| onMouseEnter | Callback function for when the mouse enters a treeNode | function({event, node}) | - |  |
+| onMouseLeave | Callback function for when the mouse leaves a treeNode | function({event, node}) | - |  |
 | onRightClick | Callback function for when the user right clicks a treeNode | function({event, node}) | - |  |
 | onSelect | Callback function for when the user clicks a treeNode | function(selectedKeys, e:{selected: boolean, selectedNodes, node, event}) | - |  |
+| onScroll | Callback function for when the tree container scrolls | function(event) | - |  |
+| dropIndicatorRender | Customize the drop indicator when dragging | (props: {dropPosition, dropLevelOffset, indent}) => ReactNode | - |  |
```
thinkasany, line 93: https://github.com/ant-design/ant-design/blob/master/components/tree/Tree.tsx#L194
https://github.com/ant-design/ant-design/blob/master/components/tree/utils/dropIndicator.tsx

It seems that this configuration is defined internally by antd, but rc-tree allows users to customize it. So it shouldn't be shown in the document?


image

components/tree/index.en-US.md
```diff
@@ -82,10 +82,15 @@ Common props ref：[Common props](/docs/react/common-props)
 | onDragOver | Callback function for when the onDragOver event occurs | function({event, node}) | - |  |
 | onDragStart | Callback function for when the onDragStart event occurs | function({event, node}) | - |  |
 | onDrop | Callback function for when the onDrop event occurs | function({event, node, dragNode, dragNodesKeys}) | - |  |
+| onDoubleClick | Callback function for when the user double clicks a treeNode | function(event, node) | - |  |
```
thinkasany, line 85: Regarding the other common events, I looked through the documents and found that other components did not mention them either. I guess they are not particularly important, but it would be more cumbersome to add them all. Welcome to give your suggestions.

components/tree/index.en-US.md
ofhope: Great catch!
It does appear as an auto complete option when using the Tree. And looks like valid code. Which is a bit misleading when using the Tree.

Screenshot 2025-07-03 135121

It should be included in this list of Omitted props so it doesn't appear to the person implementing the Tree.

```
export interface TreeProps<T extends BasicDataNode = DataNode>
  extends Omit<
    RcTreeProps<T>,
    'prefixCls' | 'showLine' | 'direction' | 'draggable' | 'icon' | 'switcherIcon'
  > {
```

thinkasany: You are right. 

Can you help us implement this type of omit?

If possible, you can remove the document changes in this pr and replace it with type omit, which should be accepted.   :)

components/tree/index.en-US.md
ofhope: We are using the Double click event, which is awesome that Antd has such a complete set of life cycle functions.

I would document the other events (onScroll, onMouseEnter, onMouseLeave) while less valuable properties, and probably not used often, they are apart of the API. And communicates to future developers browsing the docs just how complete and robust a system Antd is.

image

afc163: There are [tons of mouse events](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5c0109cf35c750c85e85ebb3d3bee22ec9388f5b/types/react/v18/index.d.ts#L2503) in React, I don't think we should document it all which we support by default.

thinkasany: https://www.geeksforgeeks.org/reactjs/react-ondoubleclick-event/

image

The onDoubleClick event in React is a native DOM event that triggers when the user double-clicks on an element.

My point of view is that if native documents are added, other components may also need to be added, which will make the documents complicated. Of course, I may not be right, and I will ask other members for their opinions.

I have tried the following method to use this event.

```
const App = () => {
  return (
    <div className="content">
      <div
        onClick={() => console.log("one click")}
        onDoubleClick={() => console.log("double click")}
      >
        onClick
      </div>
    </div>
  )
}

export default App
```

But no matter what, I still should thank you for your PR and feedback.

thinkasany:
Mouse Event that are already supported by default will not be considered. Closing this for now. Feel free to submit a separate fix for type-related issues.

PR #54393, opened 2025-07-15
fix(Menu): Allow data attributes for menu items

### LINKED ISSUE #39518: Add data-* attributes to ItemType[] within menu object

Using React testing library we should be able to get an element by "getByTestId" and we need the ability to add "data-testid" attribute to every item inside ItemType[] configuration.

Option to add data-* attributes to items inside ItemType[]

msyavuz:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues
Fixes #39518

### 💡 Background and Solution
Menu items accept data attributes but in a convoluted and a hacky way:
```tsx
<Dropdown 
  menu={{
    items, 
    ...{'data-testid': 'user-menu'}
  }} 
/>
```
This PR aims to fix that by allowing props starting with `data-`. This also avoids having to allow any props for menu item #53128(although menu item allows anything, only typescript errors.)

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Menu items support data attributes |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/menu/interface.ts
+++ b/components/menu/interface.ts
@@ -6,7 +6,7 @@ import type {
 } from 'rc-menu/lib/interface';
 
 export type DataAttributes = {
-  [Key in `data-${string}`]: string;
+  [Key in `data-${string}`]: string | number;
 };
 
 export interface MenuItemType extends RcMenuItemType, DataAttributes {
--- a/components/menu/interface.ts
+++ b/components/menu/interface.ts
@@ -5,7 +5,11 @@ import type {
   SubMenuType as RcSubMenuType,
 } from 'rc-menu/lib/interface';
 
-export interface MenuItemType extends RcMenuItemType {
+export type DataAttributes = {
+  [Key in `data-${string}`]: string;
+};
+
+export interface MenuItemType extends RcMenuItemType, DataAttributes {
   danger?: boolean;
   icon?: React.ReactNode;
   title?: string;


afc163 (dismissed):

msyavuz:
@afc163 Can i get another review?

thinkasany (approved):

RivaJ-github:
This update caused an exception in my previously functioning code！ @afc163 @msyavuz 
I create an interface like 
``` typescript
export interface MenuItem {
  key: string
  label: string
  icon?: React.ReactNode
  children?: MenuItem[]
}
```
and set items prop as MenuItem[].
before this PR,  MenuItem can be automatically inferred to specific type of ItemType<MenuItemType>. 
But now I get an error below:
``` bash
error TS2322: Type 'MenuItem[]' is not assignable to type 'ItemType<MenuItemType>[]'.
  Type 'MenuItem' is not assignable to type 'ItemType<MenuItemType>'.
    Type 'MenuItem' is not assignable to type 'SubMenuType<MenuItemType>'.
      Types of property 'children' are incompatible.
        Type 'MenuItem[] | undefined' is not assignable to type 'ItemType<MenuItemType>[]'.
          Type 'undefined' is not assignable to type 'ItemType<MenuItemType>[]'.
```
I firmly believe that this merger has caused TS to lose its type inference capability.

msyavuz:
@RivaJ-github This change would not cause an error like that. I suggest you file a proper bug report with minimal reproduction, so people can take a look at it directly.

RivaJ-github:
> @RivaJ-github This change would not cause an error like that. I suggest you file a proper bug report with minimal reproduction, so people can take a look at it directly.

I accepted your advice and tried to make a minimal reproduction. But I'm not sure if this is a bug now. Cause by changing the interface declaration to a type declaration, I unexpectedly solved the problem.

``` typescript
import React from "react";
import { Menu } from "antd";

interface MenuItem {
  key: string;
  label: string;
  icon?: React.ReactNode;
  children?: MenuItem[];
}

const errItems: MenuItem[] = [];

const errMenu = <Menu items={errItems} />; // error TS2322: Type 'MenuItem[]' is not assignable to type 'ItemType<MenuItemType>[]'.

type MenuItemType = {
  key: string;
  label: string;
  icon?: React.ReactNode;
  children?: MenuItem[];
};

const okItems: MenuItemType[] = [];

const okMenu = <Menu items={okItems} />; // no error
```

msyavuz:
@RivaJ-github Also i think ant design already exports MenuItemType or something similar to use with menu items, that maybe helpful as well

hylin:
@msyavuz We got same lint error, similar to @RivaJ-github 's case, we defined some custom MenuItem type.
I think `[Key in `data-${string}`]: string | number;` is an optional prop.Why it has caused TS to lose it's type inference capability?
Error log: https://github.com/ant-design/x/actions/runs/16400585621/job/46339414382?pr=1050
Work around: https://github.com/ant-design/x/pull/1054/files

components/menu/interface.ts
```diff
@@ -5,7 +5,11 @@ import type {
   SubMenuType as RcSubMenuType,
 } from 'rc-menu/lib/interface';
 
-export interface MenuItemType extends RcMenuItemType {
+export type DataAttributes = {
+  [Key in `data-${string}`]: string | number;
```
gregor-mueller, line 9: This is a 'breaking change' (means the app does not build after the update without any code changes on our end).

Either, the value should be `any` or just `string`.

Let me evaluate:
`any`: Anything is allowed as a `data-` attribute value.
`string`: Any value, except for `null | undefined` will be a string in the DOM node. `null`, `undefined` and function values result in the attribute to not be present at all. As everything is being `toString`ed, array values will be comma separated and objects will become the good old `[object Object]`.

I think, `any` is the way to go here, as everything should be accepted at this place.

gregor-mueller: Here is a quick TS playground to illustrate my issue:

https://www.typescriptlang.org/play/?#code/JYWwDg9gTgLgBAJQKYEMDG8BmUIjgcilQ3wChTgA7GJKTdJRNAWSUoFcBJGkAFQE8wjAN6k4cTBAgAuOAGcYUKgHMA9KrgArdgrgpKcJAA8U4ADZJSAX3LHIsODEGMAIihgoAgjEXAARuw0cnAAvHCi4gDaANJI-HBUcAAGACbuKAC0ACTCCkqUylZJALqyHCB+tHAAPvK+BQDc1k0U1LT0aIysHN6+AUGGRjSUKcFuHr1K-UhyADRM3Vw8AkLhYuJwaQW0APyyflIW+k0bCWgQlHuIxDAAdMjoMAByEClIJxswwDAWV3kqH3E6jgt1B1nImHYlAwwAucAAwrhIJQ2DAABRgHBgOSyRaTfyBGYASjWpyIMHYUAMHDMZjgwJSEBmcEoEHgIHcNCgcAAFrRLDYIVCYXDEeALqjeDN4GiSRENuTKQYADwAPnW4mVYuRqIkUhC+D8KCg+E26QykggZo8GSNUBCwnoZjkSCs1syRoAXiEAEQAVh99NV9I0MB5wGCEYkVCQ8xQkfgch5EHYZhScEqGrgWqREuo7otUgLdodTpdbrSNq9voDQZDjnDkeC53MKCoKgb7mSlqSCWCIAjchUcYT8mTqfTmdOytU6sF50ouggfk0SAwUoUAEZcWx2PjpsEwvKgQAqOAvGjzZMAdwbjFoOCgwUqZggt7k15QtLfd70GHYX6GFAj7zGGXZJD2fZwAOchDgUACEcAnqoWb4JWmSWvgsiKOwsb1kmKZphmjCYDGqHobaxpYRIX4uvMwIEROejOlalTRiiejBPo8Rsny3IAG5frhnHXI8cDXsAtJwOcwFrj88SYhA-HAG8hj8bQ-Bhh28Z6Mk-wFL2MBWmGrgAPLMCyrwzOR5petR+AACwAExkPOFxLiuckbjATk7j0PhTISh6khsqhnsAmC-hRnKBQMpHwFQRm-kkeIBQSQSGc48xJMuq7rtKTm9lGglmCp7hIOmMlEBgZjxPo6YEdewQmUB0BQKBPLgZBUYwXByhIShpxoeamGyA5sw2VWKCevZLnWEAA

This relates to the others' issue with the 'broken inference': See, how in line 30, the `objectTest1` does not show the correct error, as `objectTest2` in line 37.

PR #54691, opened 2025-08-15
fix(Typography): Paragraph ellipsis tooltip becomes inconsistent when used inside Select options.

### LINKED ISSUE #54677: Typography.Paragraph's ellipsis tooltip becomes inconsistent when used inside Select options.
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/sandbox/adoring-poincare-qvl5gz?file=%2Fpackage.json%3A14%2C22-14%2C28)

### Steps to reproduce

1. Create a Select with options containing Typography.Paragraph with ellipsis tooltip
2. In the example of refreshing codesandbox, click Select, and hover the mouse over the omitted option, the tooltip will not be displayed
3. After that, when I switched options several times, turning on and off Select, there would occasionally be a situation where the Tooltip couldn't be displayed when the mouse hovered over it, but I couldn't find a pattern

### What is expected?

The Option exceeds the option limit and can be omitted to display the Tooltip normally

When refreshing the examples in codesandbox for the first time, open Select. When the mouse hovers over options that exceed the omitted ones, there is no Tooltip. Meanwhile, during multiple switches to open, the tooltip may occasionally fail to display

| Environment | Info |
| --- | --- |
| antd | 5.27.0 |
| React | 19.0.0 |
| System | MacOS |
| Browser | Google |

meet-student:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

> - close #54677

### 💡 Background and Solution

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     fix: Typography.Paragraph's ellipsis tooltip becomes inconsistent when used inside Select options.      |
| 🇨🇳 Chinese |    fix: Typography.Paragraph's ellipsis tooltip becomes inconsistent when used inside Select options.       |


components/typography/Base/util.ts
```diff
@@ -41,7 +42,10 @@ export function isEleEllipsis(ele: HTMLElement): boolean {
     childRect.right > rect.right ||
     // Vertical out of range
     rect.top > childRect.top ||
-    childRect.bottom > rect.bottom
+    childRect.bottom > rect.bottom ||
+    // Overflow
+    ele.scrollWidth > ele.clientWidth ||
```
zhouxinyong: These two lines could be placed at the beginning of the isEleEllipsis function. That would be better, right? @Meet-student

meet-student: > These two lines could be placed at the beginning of the isEleEllipsis function. That would be better, right?

```js
    // Overflow
    ele.scrollWidth > ele.clientWidth ||
    ele.scrollHeight > ele.clientHeight ||
    // Horizontal out of range
    rect.left > childRect.left ||
    childRect.right > rect.right ||
    // Vertical out of range
    rect.top > childRect.top ||
    childRect.bottom > rect.bottom
```

Like this? @zhouxinyong

zhouxinyong: just like this, @Meet-student 
```ts
export function isEleEllipsis(ele: HTMLElement): boolean { 
  if (ele.scrollWidth > ele.clientWidth || ele.scrollHeight > ele.clientHeight) {
    return true
  }
  // old code
}
```

meet-student: @zhouxinyong  why?

zhouxinyong: First, judge this newly added condition. If this condition is already satisfied, there is no need to execute the following judgment, nor is it necessary to execute this: const rect = ele.getBoundingClientRect(); const childRect = childDiv.getBoundingClientRect();
I think this is better,  @Meet-student

meet-student: Nice. You're right, thank you for your suggestion @zhouxinyong

components/typography/Base/util.ts
```diff
@@ -17,8 +17,13 @@ export function getNode(dom: React.ReactNode, defaultNode: React.ReactNode, need
  * ref:
  * - https://github.com/ant-design/ant-design/issues/50143
  * - https://github.com/ant-design/ant-design/issues/50414
+ * - https://github.com/ant-design/ant-design/issues/54677
  */
 export function isEleEllipsis(ele: HTMLElement): boolean {
+  // Overflow
+  if (ele.scrollWidth > ele.clientWidth || ele.scrollHeight > ele.clientHeight) {
```
zombieJ, line 24: https://github.com/ant-design/ant-design/commit/b9d59339758cd8d88e35ceb69cf8cff8f31961fb#diff-e0f493a5b835b921d62f8ac43a2be141090ef6bcf073c437ea2228d07e926248L284

过去就是这么实现的

PR #54842, opened 2025-09-02
fix(locale): added missing Italian translations for QRCode & ColorPicker components

nikzanda:
…ker components

### 🤔 This is a ...

- [ ] 🆕 New feature
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [x] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

### 💡 Background and Solution

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🇮🇹 added missing Italian translations for QRCode & ColorPicker components |
| 🇨🇳 Chinese | 🇮🇹 为 QRCode 和 ColorPicker 组件添加了缺失的意大利语翻译 |

### DIFF
--- a/components/locale/it_IT.ts
+++ b/components/locale/it_IT.ts
@@ -134,6 +134,17 @@ const localeValues: Locale = {
   Image: {
     preview: 'Anteprima',
   },
+  QRCode: {
+    expired: 'Codice QR scaduto',
+    refresh: 'Ricarica',
+    scanned: 'Scansionato',
+  },
+  ColorPicker: {
+    presetEmpty: 'Vuoto',
+    transparent: 'Trasparente',
+    singleColor: 'Singolo',
+    gradientColor: 'Gradiente',
+  },
 };
 
 export default localeValues;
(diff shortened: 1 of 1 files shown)


components/locale/it_IT.ts
```diff
@@ -134,6 +134,17 @@ const localeValues: Locale = {
   Image: {
     preview: 'Anteprima',
   },
+  QRCode: {
+    expired: 'Codice QR scaduto',
+    refresh: 'Ricarica',
```
thinkasany, line 139: I also learned a word `aggiornare` , but I'm not good at the language and not sure if it would be more appropriate. What do you think?

thinkasany (approved):

components/locale/it_IT.ts
nikzanda: As a native italian speaker, I would use `ricarica` for refresh and `aggiorna` for update.

thinkasany: Thanks for the explanation and contribution. I'm coming from another language so I was a little confused.

nikzanda: No worries! 😄

PR #54933, opened 2025-09-09
fix(wave): make wave effect work with tailwind disabled variant

Jiyur:
### 🤔 This is a ...

- [X] 🐞 Bug fix

### 🔗 Related Issues

None

### 💡 Background and Solution

> - When using Tailwind's `disabled:` variant to style Ant Design components that feature wave effects (such as buttons), the wave animation fails to function properly.
> - The issue stems from the current implementation of the wave component, which performs a string-based check for the word "disabled" within the container's className. This approach incorrectly identifies Tailwind's disabled: variants as disabled states.
> - Example of problematic usage: 
```
<Button className="disabled:text-red-500">Hello</Button>
```

**Solution:** Remove the className-based disabled state detection from the wave component implementation. Instead, rely on the element's actual disabled attribute or aria states to determine whether wave effects should be applied.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   fix(wave): make wave effect work with tailwind disabled variant    |
| 🇨🇳 Chinese |  修复（波浪）：使波浪效果与顺风禁用变体一起工作         |

### DIFF
--- a/components/_util/wave/index.ts
+++ b/components/_util/wave/index.ts
@@ -44,7 +44,8 @@ const Wave: React.FC<WaveProps> = (props) => {
         !node.getAttribute ||
         node.getAttribute('disabled') ||
         (node as HTMLInputElement).disabled ||
-        node.className.includes('disabled') ||
+        (node.className.includes('disabled') && !node.className.includes('disabled:')) ||
+        node.getAttribute('aria-disabled') === 'true' ||
         node.className.includes('-leave')
       ) {
         return;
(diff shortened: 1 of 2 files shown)


components/_util/wave/index.ts
```diff
@@ -44,7 +44,6 @@ const Wave: React.FC<WaveProps> = (props) => {
         !node.getAttribute ||
         node.getAttribute('disabled') ||
         (node as HTMLInputElement).disabled ||
-        node.className.includes('disabled') ||
```
zombieJ, line 47: do not remove directly. Shoud be inlcude `disabled` but not `disabled:`.

Jiyur: @zombieJ  Thanks you very much for the suggestion. I have updated the PR, please have a check again

afc163 (approved):

PR #55075, opened 2025-09-18
fix(list): adjust bordered list overflow styling

Jiyur:
### 🤔 This is a ...

- [X] 💄 Component style improvement

### 🔗 Related Issues

None

### 💡 Background and Solution

> - List with `bordered` has overflow issue styling with it's `header` and `footer`

**Solution:**
Apply overflow hidden to `bordered` List to avoid this issue

**Before**
image
image

**After**
image
image

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   fix(list): adjust bordered list overflow styling   |
| 🇨🇳 Chinese |  修复（列表）：调整有边框的列表溢出样式     |

### DIFF
--- a/components/list/style/index.ts
+++ b/components/list/style/index.ts
@@ -94,10 +94,10 @@ const genBorderedStyle = (token: ListToken): CSSObject => {
   } = token;
 
   // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the top corners.
-  const topCornerBorderRadius = `${unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal())}`;
+  const topCornerBorderRadius = unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal());
 
   // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the bottom corners.
-  const bottomCornerBorderRadius = `${unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal())}`;
+  const bottomCornerBorderRadius = unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal());
 
   return {
     [listBorderedCls]: {
--- a/components/list/style/index.ts
+++ b/components/list/style/index.ts
@@ -94,9 +94,20 @@ const genBorderedStyle = (token: ListToken): CSSObject => {
   } = token;
   return {
     [listBorderedCls]: {
-      overflow: 'hidden',
       border: `${unit(token.lineWidth)} ${token.lineType} ${token.colorBorder}`,
       borderRadius: borderRadiusLG,
+      // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the top corners.
+      [`${componentCls}-header`]: {
+        borderTopLeftRadius: token.calc(borderRadiusLG).sub(token.lineWidth).equal(),
+        borderTopRightRadius: token.calc(borderRadiusLG).sub(token.lineWidth).equal(),
+      },
+
+      // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the bottom corners.
+      [`${componentCls}-footer`]: {
+        borderBottomLeftRadius: token.calc(borderRadiusLG).sub(token.lineWidth).equal(),
+        borderBottomRightRadius: token.calc(borderRadiusLG).sub(token.lineWidth).equal(),
+      },
+
       [`${componentCls}-header,${componentCls}-footer,${componentCls}-item`]: {
         paddingInline: paddingLG,
       },
--- a/components/list/style/index.ts
+++ b/components/list/style/index.ts
@@ -92,20 +92,23 @@ const genBorderedStyle = (token: ListToken): CSSObject => {
     marginLG,
     borderRadiusLG,
   } = token;
+
+  // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the top corners.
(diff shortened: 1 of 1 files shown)


components/list/style/index.ts
```diff
@@ -94,6 +94,7 @@ const genBorderedStyle = (token: ListToken): CSSObject => {
   } = token;
   return {
     [listBorderedCls]: {
+      overflow: 'hidden',
```
afc163: 尽量不加 overflow

Jiyur: Thank for the suggestion, I tried new approach and it worked well (by apply border radius to both list header and footer). Please help me review it

components/list/style/index.ts
```diff
@@ -91,11 +91,22 @@ const genBorderedStyle = (token: ListToken): CSSObject => {
...
     borderRadiusLG,
+    calc,
   } = token;
   return {
     [listBorderedCls]: {
       border: `${unit(token.lineWidth)} ${token.lineType} ${token.colorBorder}`,
       borderRadius: borderRadiusLG,
+      [`${componentCls}-header`]: {
+        borderTopLeftRadius: calc(borderRadiusLG).sub(1).equal(),
+        borderTopRightRadius: calc(borderRadiusLG).sub(1).equal(),
```
afc163:
```suggestion
        borderTopLeftRadius: borderRadiusLG,
        borderTopRightRadius: borderRadiusLG,
```

Jiyur: @afc163 If I apply original `borderRadiusLG`, it still has bug on the UI like this 

With original `borderRadiusLG` (still have white space on the top left and top right not covered by pink background - **please zoom in too see the issue**):
image
=> I think this happened because of the list container `border` width 

With subtract by 1 `borderRadiusLG`:
image

Do you think we need to keep the original `borderRadiusLG` or subtract it by 1 (list container border with) in this case?

components/list/style/index.ts
```diff
@@ -96,6 +96,16 @@ const genBorderedStyle = (token: ListToken): CSSObject => {
...
       borderRadius: borderRadiusLG,
+      [`${componentCls}-header`]: {
+        borderTopLeftRadius: borderRadiusLG,
+        borderTopRightRadius: borderRadiusLG,
+      },
+
+      [`${componentCls}-footer`]: {
+        borderBottomLeftRadius: borderRadiusLG,
+        borderBottomRightRadius: borderRadiusLG,
+      },
```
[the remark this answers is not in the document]
Jiyur: @afc163  This is the bug as I mentioned above, happened when apply the original `borderRadiusLG` without subtract it by border `width` of outer element

components/list/style/index.ts
```diff
@@ -92,10 +92,25 @@ const genBorderedStyle = (token: ListToken): CSSObject => {
     marginLG,
     borderRadiusLG,
   } = token;
+
+  // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the top corners.
+  const topCornerBorderRadius = `${unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal())}`;
+
+  // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the bottom corners.
+  const bottomCornerBorderRadius = `${unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal())}`;
```
afc163: 看上去是一样的？

Jiyur: > 看上去是一样的？

Although the logic is the same, I chose to separate the calculation into two variables—one for the header and one for the footer—to improve readability and maintainability. In my opinion, this approach also adds flexibility in case the header and footer require different border radius values in the future.

Jiyur: > > 看上去是一样的？
> 
> Although the logic is the same, I chose to separate the calculation into two variables—one for the header and one for the footer—to improve readability and maintainability. In my opinion, this approach also adds flexibility in case the header and footer require different border radius values in the future.

 if both header and foot always identical, i can merged single variable then

components/list/style/index.ts
```diff
@@ -92,10 +92,25 @@ const genBorderedStyle = (token: ListToken): CSSObject => {
     marginLG,
     borderRadiusLG,
   } = token;
+
+  // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the top corners.
+  const topCornerBorderRadius = `${unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal())}`;
```
li-jia-nan:
```suggestion
  const topCornerBorderRadius = unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal());
```

li-jia-nan (approved):

components/list/style/index.ts
```diff
@@ -92,10 +92,25 @@ const genBorderedStyle = (token: ListToken): CSSObject => {
     marginLG,
     borderRadiusLG,
   } = token;
+
+  // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the top corners.
+  const topCornerBorderRadius = unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal());
+
+  // Inner radius = Outer radius - border width; this prevents white gaps from appearing at the bottom corners.
+  const bottomCornerBorderRadius = unit(token.calc(borderRadiusLG).sub(token.lineWidth).equal());
```
afc163, line 100: Please use one variable.

li-jia-nan: - #55241

PR #55110, opened 2025-09-21
chore: turn on correctness.useHookAtTopLevel

thinkasany:
### 🤔 This is a ...
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

### 💡 Background and Solution

https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level

image

```
components/form/Form.tsx:97:5 lint/correctness/useHookAtTopLevel ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━

  ✖ This hook is being called conditionally, but all hooks must be called in the exact same order in every component render.

    95 │   if (process.env.NODE_ENV !== 'production') {
    96 │     // eslint-disable-next-line react-hooks/rules-of-hooks
  > 97 │     useFormWarning(props);
       │     ^^^^^^^^^^^^^^
    98 │   }
    99 │ 

  ℹ For React to preserve state between calls, hooks needs to be called unconditionally and always in the same order.

  ℹ See https://reactjs.org/docs/hooks-rules.html#only-call-hooks-at-the-top-level
```

wanring hooks 内部用了 ```process.env.NODE_ENV !== 'production'```, 所以应该问题不大
image

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |      -     |
| 🇨🇳 Chinese |        -   |

### DIFF
--- a/components/form/Form.tsx
+++ b/components/form/Form.tsx
@@ -92,10 +92,12 @@ const InternalForm: React.ForwardRefRenderFunction<FormRef, FormProps> = (props,
 
   const contextValidateMessages = React.useContext(ValidateMessagesContext);
 
+  /* eslint-disable react-hooks/rules-of-hooks */
   if (process.env.NODE_ENV !== 'production') {
-    // eslint-disable-next-line react-hooks/rules-of-hooks
+    // biome-ignore lint/correctness/useHookAtTopLevel: Development-only warning hook called conditionally
     useFormWarning(props);
   }
+  /* eslint-enable */
 
   const mergedRequiredMark = React.useMemo(() => {
     if (requiredMark !== undefined) {
--- a/biome.json
+++ b/biome.json
@@ -52,7 +52,6 @@
       "correctness": {
         "useUniqueElementIds": "off",
         "useExhaustiveDependencies": "off",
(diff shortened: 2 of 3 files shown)


components/form/Form.tsx
```diff
@@ -92,10 +92,8 @@ const InternalForm: React.ForwardRefRenderFunction<FormRef, FormProps> = (props,
 
   const contextValidateMessages = React.useContext(ValidateMessagesContext);
 
-  if (process.env.NODE_ENV !== 'production') {
```
li-jia-nan, line 95: 实际上在 `devUseWarning` 里面已经判断过 env 了，外面这个判断我记得是专门加上的，是为了让 bundle 工具在构建阶段跳过这块代码

thinkasany: 原来如此，那我加个 biome 注释跳过这个吧，这个规则还是不要关闭比较好

thinkasany:
@li-jia-nan  佬，求个 +1

image

meet-student (approved):

PR #55192, opened 2025-09-29
feat: add numeric input only support for OTP component

ahmedsamirdev:
### 🤔 This is a ...
- [x] 🆕 New feature
- [x] 📽️ Demo improvement

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Input.OTP: add `numericOnly` to restrict input to digits and enable mobile-friendly numeric keyboards. |
| 🇨🇳 Chinese | Input.OTP：新增 `numericOnly` 属性，限制仅输入数字，并优化移动端数字键盘体验。 |


meet-student:
The ci is incorrect and needs to be checked

meet-student:
The target branch should be feature

ahmedsamirdev:
I’ve made the required changes to the base branch and updated the CI test.
The initial CI failures appear to be unrelated to this PR.

li-jia-nan:
@zombieJ Do you think this needs to be supported?

components/input/index.en-US.md
```diff
@@ -135,6 +135,7 @@ Added in `5.16.0`.
 | defaultValue | Default value | string | - |  |
 | disabled | Whether the input is disabled | boolean | false |  |
 | formatter | Format display, blank fields will be filled with ` ` | (value: string) => string | - |  |
+| numericOnly | Whether to allow only numeric input | boolean | false | 5.x.x |
```
zombieJ, line 138: currently we already have `formatter` props. So we can just reuse the API:

```tsx
<OTP formatter="numberic" />
```

ahmedsamirdev: This approach can lead to unexpected behavior, such as silent input filtering, typed characters are ignored, pasted text is stripped without feedback, screen readers provide no announcement, and the wrong mobile keyboard may appear.

As per discussion above, we currently don’t have a numbers-only option.

Also, using <Input.OTP type="number" /> would require overriding styles to match the OTP component’s appearance.

zombieJ: This is a API design. It's nothing different with `type=xxx` in coding.

afc163:
Thanks for the contribution. The proposed `numericOnly` API remained blocked at the API-design stage: the existing formatting path can already filter input, and the review did not establish a sufficiently distinct contract for another boolean prop. There has been no implementation follow-up since the final maintainer feedback in October 2025, and the branch now conflicts with `feature`.

Closing this inactive implementation. A fresh proposal is welcome if it first defines the mobile-keyboard and accessibility behavior that cannot be represented by the existing API, together with focused tests.

PR #55199, opened 2025-09-30
refactor: semantic of message and notification

thinkasany:
### 🤔 This is a ...

- [ ] ❓ Other (about what?)

### 🔗 Related Issues

### 💡 Background and Solution

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    -      |
| 🇨🇳 Chinese |       -    |

### DIFF
--- a/components/message/useMessage.tsx
+++ b/components/message/useMessage.tsx
@@ -9,8 +9,8 @@ import type {
 } from '@rc-component/notification';
 import classNames from 'classnames';
 
+import useMergeSemantic, { mergeClassNames, mergeStyles } from '../_util/hooks/useMergeSemantic';
 import { devUseWarning } from '../_util/warning';
-import useMergeSemantic from '../_util/hooks/useMergeSemantic';
 import { ConfigContext } from '../config-provider';
 import { useComponentConfig } from '../config-provider/context';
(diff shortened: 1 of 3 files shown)


components/message/useMessage.tsx
```diff
@@ -219,21 +219,14 @@ export function useInternalMessage(
         resolveFunctionStyle(value, config),
```
li-jia-nan: 还有两个可以优化的点：

这个 resolveFunctionStyle 方法没有依赖到组件内部的状态，可以放在外面作为纯函数，另外可以简化一下：

```ts
const resolveFunctionStyle = <T extends AnyObject>(
  value: T | ((config: any) => T),
  info: { props: AnyObject },
) => {
  return typeof value === 'function' ? value(info) : value;
};
```

还有调用的地方没必要用 map，徒增了复杂度，直接用就行：

```ts
const semanticClassNames = resolveFunctionStyle(configClassNames, { props: config });

const semanticStyles = resolveFunctionStyle(styles, { props: config });
```

thinkasany:
bun 又抽了

image

zombieJ (approved):

Issue #53043, opened 2025-03-04
Support handling drag and drop events on Tree.DirectoryTree that came from outside the browser


Cameron-Asdf:
### What problem does this feature solve?

I want be able to handle files being dropped onto specific parts of a Tree.DirectoryTree from the system file browser outside the browser.
Currently, it seems like the dragging-related events (`onDragEnter`, `onDragOver`, `onDrop`...) don't fire for these drags.

The end user experience here is that we are using DirectoryTree to visualize datasets that are calculated from uploaded files and users want to upload more files by just dragging them to that list.

### What does the proposed API look like?

Either onDragEnter, onDragOver and onDrop should be able to behave like their regular HTML equivalents or there should be some way to set the HTML versions of these, ideally per item in the tree. (to be able to tell which one a file was dropped on)

MadCcc:
Only support drag node from Tree now.
Maybe we could support external node in `onDrop` event.

PR #55884, opened 2025-11-25
feat(DirectoryTree): Support handling drag and drop events from outside the browser

caronchen:

### 🤔 This is a ...

- [X] 🆕 New feature

### 🔗 Related Issues

https://github.com/ant-design/ant-design/issues/53043

### 💡 Background and Solution

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |   Support dragging files or directories from the file system into the directory tree.   |
| 🇨🇳 Chinese |  支持从文件系统中拖拽文件或目录到目录树中。  |


caronchen:
#53043

afc163:
Thanks for the contribution. Closing because the implementation coerces numeric Tree keys to strings, so the callback cannot faithfully return the public `Key` value. This needs a type-preserving implementation.

PR #56417, opened 2025-12-30
[performance] optimize Masonry virtualization 

Alfredoeb9:
> * Implement height estimation to support large datasets (e.g., 100k items) without freezing.
> * Fix initial render freeze by limiting the number of unmeasured items.
> * Fix item overlap issue caused by missing position data on first render.
> * Fix CSS width calculation syntax error.
> * Add `virtual.test.tsx` to verify scrolling and virtualization behavior.
> * Updated Dynamic sub component as setState was being called repeatedly
> * Fixed missing Top style in Masonry item layout calculation to ensure correct position
> * Properly restore ResizeObserver in virtual.test to prevent mock leak into other tests
> * Update demo snapshot to reflect the layout fixes

### 🤔 This is a ...

- [ ] 🆕 New feature
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [x ] ⚡️ Performance optimization
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x ] 🎨 Code style optimization
- [ x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

> - Fixes issue: #56370

### 💡 Background and Solution

> - It was found after a large number of elements to be rendered on screen using the Masonry component, performance would greatly decline and at times crash the application. 
> - Solution was to find all properties of the elements wanted and allowing only a defined set of elements to render based on the window client height.

### 📝 Change Log

> * Implement height estimation to support large datasets (e.g., 100k items) without freezing.
> * Fix initial render freeze by limiting the number of unmeasured items.
> * Fix item overlap issue caused by missing position data on first render.
> * Fix CSS width calculation syntax error.
> * Add `virtual.test.tsx` to verify scrolling and virtualization behavior.
> * Updated Dynamic sub component as setState was being called repeatedly
> * Fixed missing Top style in Masonry item layout calculation to ensure correct position
> * Properly restore ResizeObserver in virtual.test to prevent mock leak into other tests
> * Update demo snapshot to reflect the layout fixes

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |  ✅ Optimized Masonry component to handle large amount of elements |
| 🇨🇳 Chinese |   ✅ 优化 Masonry 组件以支持大量元素的渲染  |

### DIFF
--- a/components/masonry/Masonry.tsx
+++ b/components/masonry/Masonry.tsx
@@ -1,4 +1,5 @@
 import * as React from 'react';
+import { useEffect, useState } from 'react';
 import type { CSSProperties } from 'react';
 import { CSSMotionList } from '@rc-component/motion';
 import ResizeObserver from '@rc-component/resize-observer';
@@ -11,6 +12,7 @@ import { useMergeSemantic } from '../_util/hooks';
 import type { SemanticClassNamesType, SemanticStylesType } from '../_util/hooks';
 import { responsiveArray } from '../_util/responsiveObserver';
 import type { Breakpoint } from '../_util/responsiveObserver';
+import throttleByAnimationFrame from '../_util/throttleByAnimationFrame';
 import { useComponentConfig } from '../config-provider/context';
 import useCSSVarCls from '../config-provider/hooks/useCSSVarCls';
 import type { RowProps } from '../grid';
@@ -75,8 +77,6 @@ export interface MasonryRef {
   nativeElement: HTMLDivElement;
 }
 
-type ItemColumnsType = [item: MasonryItemType, column: number];
-
 const Masonry = React.forwardRef<MasonryRef, MasonryProps>((props, ref) => {
   const {
     rootClassName,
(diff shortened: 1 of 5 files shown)


Alfredoeb9:
@thinkasany can i get a review please

components/masonry/Masonry.tsx
```diff
@@ -123,6 +123,56 @@ const Masonry = React.forwardRef<MasonryRef, MasonryProps>((props, ref) => {
...
+  useEffect(() => {
+    const container = containerRef.current;
+
+    if (!container) return;
+
+    // find scrollable parent
+    let scrollNode: HTMLElement | Window = window;
+    let parent = container.parentElement;
+
+    while (parent) {
```
zombieJ, line 138: Maybe we can just let scroll ability with the root div instead of parent loop. It will meet too many edge case for this. And `getComputedStyle` is also a perf cost.

components/masonry/Masonry.tsx
```diff
@@ -167,66 +217,120 @@ const Masonry = React.forwardRef<MasonryRef, MasonryProps>((props, ref) => {
...
+    });
+  }, [mergedItems, itemPositions, allItemHeights]);
 
-  React.useEffect(() => {
-    collectItemSize();
-  }, [mergedItems, columnCount]);
+  // ====================== Virtual ======================
+  const visibleItems = React.useMemo(() => {
+    if (!viewState.clientHeight) {
+      return itemWithPositions.slice(0, 50);
```
zombieJ, line 297: Should not hard code here. If data is not ready, maybe just not render.

components/masonry/Masonry.tsx
```diff
@@ -167,66 +217,120 @@ const Masonry = React.forwardRef<MasonryRef, MasonryProps>((props, ref) => {
...
+
+      mergedItems.forEach((item, index) => {
+        const itemKey = item.key ?? index;
+        const itemEle = getItemRef(itemKey);
+        const rect = itemEle?.getBoundingClientRect();
+
+        if (rect) {
+          nextItemsHeight.push([itemKey, rect.height, item.column]);
+        } else {
+          const prevHeight = prevHeightMap.get(itemKey) || averageHeight;
```
zombieJ, line 262: averageHeight seems not a good design. If item has large diff of size, this is not stable.

afc163:
Closing for no further response.

PR #56524, opened 2026-01-07
fix(Dropdown) overlayNode prop pass

Damon4:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

### 💡 Background and Solution

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |      fix(Dropdown) overlayNode prop pass     |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/dropdown/dropdown.tsx
+++ b/components/dropdown/dropdown.tsx
@@ -73,6 +73,8 @@ export interface DropdownProps {
   rootClassName?: string;
   transitionName?: string;
   placement?: Placement;
+  /** @deprecated please use `menu.items` instead.*/
+  overlayNode?: React.ReactNode;
   /** @deprecated please use `classNames.root` instead.*/
   overlayClassName?: string;
   /** @deprecated please use `styles.root` instead.*/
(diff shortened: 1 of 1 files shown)


afc163:
Should not add a deprecated prop.

Damon4:
> Should not add a deprecated prop.

I generally agree, but it can still work, and besides, migrating to a new API is very expensive.

thinkasany:
> > Should not add a deprecated prop.
> 
> I generally agree, but it can still work, and besides, migrating to a new API is very expensive.

However, compatibility with deprecated APIs will increase maintenance costs.

Damon4:
well, now it's just going to be a React.Children.only error, that's all, the solution is not so complicated and will allow to keep the solution still working, albeit temporarily.

thinkasany:
I'm still sorry, but I feel that this deprecated feature is indeed not a good idea to adopt, so I'll turn it off for now.

PR #56930, opened 2026-02-10
feat(ConfigProvider, Select): support showSearch, allowClear, clearIcon, loadingIcon, menuItemSelectedIcon, removeIcon and suffixIcon config

Pareder:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [x] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

Closes #55428

### 💡 Background and Solution

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | feat(ConfigProvider, Select): support showSearch, allowClear, clearIcon, loadingIcon, menuItemSelectedIcon, removeIcon and suffixIcon config |
| 🇨🇳 Chinese | feat(ConfigProvider, Select): 支持 showSearch、allowClear、clearIcon、loadingIcon、menuItemSelectedIcon、removeIcon 和 suffixIcon 配置 |


afc163 (dismissed):

components/select/index.tsx
```diff
@@ -289,6 +290,7 @@ const InternalSelect = <
   // ===================== Icons =====================
   const { suffixIcon, itemIcon, removeIcon, clearIcon } = useIcons({
     ...rest,
+    suffixIcon: rest.suffixIcon === undefined ? contextSuffixIcon : rest.suffixIcon,
```
afc163:
```suggestion
    suffixIcon: rest.suffixIcon ?? contextSuffixIcon,
```

Pareder: In this case `suffixIcon={null}` will not work.

thinkasany: support in https://github.com/ant-design/ant-design/pull/56725/changes#diff-7d03523d6691d33926ab749f915504adb7083435a36ed9584b19ad620f13fe94R77-R89

Pareder: @thinkasany But that PR did not add support of context icons for `Select` component, only for `Cascader`. I can modify and reopen this PR.

thinkasany: I'm so sorry, I missed it. Thanks for your contribution.

Pareder:
@thinkasany Done.
- added new `Select` props:
    - `loadingIcon`
    - `showSearch.searchIcon`
- extended `ConfigProvider` Select props:
    - `showSearch` (fixed type)
    - `allowClear`
    - `clearIcon`
    - `loadingIcon`
    - `menuItemSelectedIcon`
    - `removeIcon`
    - `suffixIcon`
 - added default icon values in `Select` documentation

components/cascader/index.tsx
```diff
@@ -385,7 +385,8 @@ const Cascader = React.forwardRef<CascaderRef, CascaderProps<any>>((props, ref)
     contextClearIcon,
     removeIcon,
     contextRemoveIcon,
-    loadingIcon: mergedLoadingIcon,
+    loadingIcon,
+    contextLoadingIcon,
```
zombieJ: It's little strange that `useIcons` already do the merge logic but `useSelectIcons` still need handle this. Maybe `useSelectIcons` should simplify its logic.

Pareder: @zombieJ You are right. Reverted `Cascader` changes at all.

Pareder:
Pipeline failed due to the latest changes in `feature` branch.

components/select/useIcons.tsx
```diff
@@ -89,15 +95,9 @@ export default function useIcons({
...
-    if (menuItemSelectedIcon !== undefined) {
-      mergedItemIcon = menuItemSelectedIcon;
-    } else if (multiple) {
-      mergedItemIcon = <CheckOutlined />;
-    } else {
-      mergedItemIcon = null;
-    }
-
+    const mergedItemIcon = multiple
+      ? fallbackProp(menuItemSelectedIcon, contextMenuItemSelectedIcon, <CheckOutlined />)
```
zombieJ: 这个逻辑和之前的不太一样，之前是如果有配置 menuItemSelectedIcon 就使用，现在变成了只有 multiple 才有了

Pareder: @zombieJ I think that it makes sense to render this icon only for multiple mode. But if you want I can roll back the previous logic like this:
```
const mergedItemIcon = fallbackProp(menuItemSelectedIcon, contextMenuItemSelectedIcon, multiple ? <CheckOutlined /> : null)
```

Pareder: Done.

Pareder:
Could you please reopen this PR? I think Github Actions closed it accidentally.

Pareder:
Markdown files were modified automatically by `lint-staged` command after merging with `feature` branch.

components/select/index.tsx
```diff
@@ -279,6 +290,16 @@ const InternalSelect = <
     showSuffixIcon,
     prefixCls,
     componentName: 'Select',
+    clearIcon: typeof allowClear === 'object' ? allowClear.clearIcon : rest.clearIcon,
+    searchIcon: typeof rest.showSearch === 'object' ? rest.showSearch.searchIcon : undefined,
```
zombieJ: 感觉这几个都可以抽成一个方法专门用来将 boolean 转从 object

Pareder: @zombieJ Do you mean that this function should make an object from `allowClear` or `showSearch` if it is a boolean?
```
function getObjectWithFallbackValue<K extends string, V extends { [key in K]?: React.ReactNode }>(
  value: boolean | V | undefined,
  key: K,
  fallback?: React.ReactNode,
) {
  if (value === false) {
    return;
  }

  if (value === true) {
    return { [key]: fallback };
  }

  return value;
}

const mergedAllowClear = getObjectWithFallbackValue(allowClear, 'clearIcon', rest.clearIcon);
const mergedShowSearch = getObjectWithFallbackValue(rest.showSearch, 'searchIcon');
```

zombieJ: yes. A simple convert function is better. Maybe named as `normalizeIcon(...)` since `typeof null` is also an object. If all inline will be lots of dulicated code

Pareder: Done.

Pareder:
Could you please reopen this PR? Markdown files were modified automatically by lint-staged command. I reverted those changes but need to reopen this PR to apply changes.

components/select/index.tsx
Pareder: @zombieJ Could you please review?

components/auto-complete/__tests__/__snapshots__/demo-extend.test.ts.snap
```diff
@@ -525,7 +525,11 @@ Array [
 ]
 `;
 
-exports[`renders components/auto-complete/demo/allowClear.tsx extend context correctly 2`] = `[]`;
+exports[`renders components/auto-complete/demo/allowClear.tsx extend context correctly 2`] = `
+[
+  "Warning: [antd: Select] \`clearIcon\` is deprecated. Please use \`allowClear={{ clearIcon: React.ReactNode }}\` instead.",
+]
```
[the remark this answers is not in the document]
thinkasany: I think the solution here should be to adjust the usage of the demo to make it correct and use the new syntax, rather than changing snap and retaining the incorrect usage.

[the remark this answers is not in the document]
Pareder: @thinkasany Fixed. The problem was with passing `allowClear.clearIcon` to `useIcons` hook as `clearIcon` prop and it was checked for deprecation here https://github.com/ant-design/ant-design/blob/master/components/select/useIcons.tsx#L45.

Pareder:
Could you please review?

Pareder:
@zombieJ @thinkasany Please review when you have time.

components/select/index.tsx
```diff
@@ -403,7 +430,7 @@ const InternalSelect = <
       virtual={virtual}
       classNames={mergedClassNames}
       styles={mergedStyles}
-      showSearch={showSearch}
+      showSearch={contextShowSearch}
```
thinkasany: It is likely `mergedShowSearch`, which merges the parameters from the context and the component.

thinkasany: Naming it that way would better align with the established conventions; otherwise, it looks as though this section is intended solely for global configuration.

components/_util/normalizeIcon.ts
```diff
@@ -0,0 +1,12 @@
+import fallbackProp from './fallbackProp';
+
+export default function normalizeIcon<K extends string, V extends { [key in K]?: React.ReactNode }>(
+  value: boolean | V | undefined,
+  key: K,
+  fallback?: React.ReactNode,
+) {
+  if (value === false) return;
```
zombieJ: It's better to return `null` instead of empty.

And avoid inline if since this will always match the test coverage even the code may not reached.

zombieJ (dismissed):

components/select/index.tsx
zombieJ: agree，seems merged may better

thinkasany:
@Pareder Thanks for your contributions!  Once we've addressed the last two comments, we'll be ready to merge this PR.

components/select/index.tsx
Pareder: The thing is that `showSearch` from component props will override the value from context in the next line (`{...selectProps}`). And it was made before my PR, I've just renamed the variable from context. But I created a separate variable `const mergedShowSearch = showSearch ?? contextShowSearch` to have straightforward value.

components/_util/normalizeIcon.ts
Pareder: Fixed the code to return `null` in case of `false` value.
What do you mean by avoiding inline if?

components/_util/normalizeIcon.ts
```diff
@@ -0,0 +1,12 @@
...
+
+export default function normalizeIcon<K extends string, V extends { [key in K]?: React.ReactNode }>(
+  value: boolean | V | undefined,
+  key: K,
+  fallback?: React.ReactNode,
+) {
+  if (value === false) return null;
+  if (value === true) return fallback;
+
+  return fallbackProp<React.ReactNode>(value?.[key], fallback);
```
thinkasany: Refactored code:
```
export default function normalizeIcon<K extends string, V extends { [key in K]?: React.ReactNode }>(
  value: boolean | V | undefined,
  key: K,
  fallback?: React.ReactNode,
) {
  if (value === false) return null;  // Explicitly return null when value is false

  if (value === true) return fallback;  // Return fallback when value is true

  // More explicit if statement, avoiding inline if
  if (value && value[key] !== undefined) {
    return value[key];  // Return value[key] if it's explicitly defined
  }

  return fallback;  // Return fallback when no value is found
}
```
Explanation of changes:

Handling value === false:

This remains unchanged: if value is false, it explicitly returns null, representing "no value."

Handling value === true:

If value is true, it returns fallback.

Avoiding inline if:

I replaced the inline if (optional chaining value?.[key]) with an explicit if statement to check if value exists and if value[key] is defined. This makes the logic clearer and more readable.

This also ensures that test coverage tools can accurately track the execution paths.

Returning fallback:

If none of the conditions are met, the function will return the fallback value, ensuring that when no matching value is found, the default value is returned.

Conclusion:

By using explicit if statements, we avoid the potential issues with inline if and improve the clarity of the logic.

Returning null to represent "no value" rather than empty objects or arrays aligns with best practices and makes the code intent clearer.

Pareder: @thinkasany Done.

thinkasany (approved):

PR #56983, opened 2026-02-13
feat(Tag): allow preset colors for CheckableTag (checkable mode)

### LINKED ISSUE #11889: tags: support preset colors for checkable mode

We can't use checkable mode with colors. But sometimes, tags with colors could be checkable. Say we try to tag a job as emergent.   

No need to add an API.

nicepopo86-lang:
Fixes ant-design/ant-design#11889.

Problem:
- Tag.CheckableTag (checkable mode) couldn't use preset colors (e.g. red/green) while normal Tag supports them.

Solution:
- Add `color` prop support to `CheckableTag` (leveraging existing `useColor` logic).
- Extend preset color styles to cover checkable state (hover/checked/active) so preset colors render correctly.

Notes:
- Keeps API surface minimal (adds `color` prop; aligns with Tag color behavior).

### DIFF
--- a/components/tag/CheckableTag.tsx
+++ b/components/tag/CheckableTag.tsx
@@ -1,14 +1,20 @@
 import * as React from 'react';
 import { clsx } from 'clsx';
 
+import type { PresetColorType, PresetStatusColorType } from '../_util/colors';
+import type { LiteralUnion } from '../_util/type';
+import { useComponentConfig } from '../config-provider/context';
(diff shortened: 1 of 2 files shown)


afc163:
feature branch

afc163:
Thanks for the contribution. Closing because this new Tag color API targets `master`, lacks the required API documentation and tests, and reverses expected style precedence. The linked issue is also already closed.

Issue #56947, opened 2026-02-11
Firefox 下 Select 选中后文字向上跳动


afc163:
![Image](https://github.com/user-attachments/assets/1a50c35e-dc57-4536-a667-5813540148fb)

zen 1.18.5b (Firefox 147.0.3) (aarch64)

spider-yamet:
@afc163 @li-jia-nan @zombieJ May I pick up this issue?

PR #57030, opened 2026-02-20
fix(select): calculate font-height from line-height × font-size to prevent text jump in Firefox

pierreeurope:
This fixes the text jumping issue in Firefox when selecting an option in the Select component (issue #56947).

The problem was that the vertical padding calculation used the global `fontHeight` token, but the actual text rendered using the component's `fontSize`. In Firefox, this mismatch in dimensions causes the text to visibly jump up after selection.

The fix calculates `font-height` from `line-height × font-size` instead of using the global token. This ensures the padding matches the actual rendered text height across all browsers.

Changes:
- Updated default size to use `calc(line-height * font-size)` for font-height
- Updated large size variant with the same calculation

Tested locally in Firefox - text no longer jumps on selection.

### DIFF
--- a/components/select/style/select-input.ts
+++ b/components/select/style/select-input.ts
@@ -266,8 +266,7 @@ const genSelectInputStyle: GenerateStyle<SelectToken, CSSObject> = (token) => {
         [`&-single:not(${componentCls}-customize)`]: {
           [`${componentCls}-input`]: {
             position: 'absolute',
-            insetInline: 0,
-            insetBlock: `calc(${varRef('padding-vertical')} * -1)`,
+            inset: 0,
             lineHeight: `calc(${varRef('font-height')} + ${varRef('padding-vertical')} * 2)`,
           },
 


afc163:
图片

components/select/style/select-input.ts
afc163:
```suggestion
inset: 0,
```

Try this one.

components/select/style/select-input.ts
```diff
@@ -104,7 +104,7 @@ const genSelectInputStyle: GenerateStyle<SelectToken, CSSObject> = (token) => {
         // Font
         [varName('font-size')]: token.fontSize,
         [varName('line-height')]: token.lineHeight,
-        [varName('font-height')]: fontHeight,
+        [varName('font-height')]: `calc(${varRef('line-height')} * ${varRef('font-size')})`,
```
afc163:
```suggestion
        [varName('font-height')]: fontHeight,
```

components/select/style/select-input.ts
```diff
@@ -232,7 +232,6 @@ const genSelectInputStyle: GenerateStyle<SelectToken, CSSObject> = (token) => {
           [varName('height')]: token.controlHeightLG,
           [varName('font-size')]: token.fontSizeLG,
           [varName('line-height')]: token.lineHeightLG,
-          [varName('font-height')]: token.fontHeightLG,
```
afc163, line 235: These changes could be reverted now.

afc163 (approved):

PR #57032, opened 2026-02-21
fix: keep Input onChange event target compatible with customInput integrations

### LINKED ISSUE #46999: Antd Input: To use with react-number-format customInput feature.

### Reproduction link

### Steps to reproduce

### What is expected?

| Environment | Info |
| --- | --- |
| antd | 5.13.3 |
| React | 18 |
| System | Windows 11 |

---

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Backers (Total: $25.00)

### Submitted pull Requests
---

### Tips

</details>

larryjiang-star:
## Summary
- normalize `Input` `onChange` events to always expose the real DOM input as `event.target/currentTarget`
- this avoids passing through the cloned target from `@rc-component/input`, which can break libraries that depend on the live input element (e.g. `react-number-format` with `customInput={Input}`)
- add regression test to ensure Ant Design `Input` forwards the real input target

Fixes #46999

## Why
`@rc-component/input` clones event targets in some change paths. That clone is useful for value normalization, but downstream integrations expecting the real input element can lose caret/focus behavior and see cursor jumps.

By normalizing in Ant Design `Input` wrapper, we keep compatibility while keeping the rest of rc-input behavior untouched.

## Before / After
- **Before**: `onChange` could receive a cloned target (not the mounted input), which breaks integrations relying on the live element.
- **After**: `onChange` always receives the actual input element as `target/currentTarget`.

## Test
- Added: `should pass real input as event target on change for custom input integrations`

### Local run notes
I attempted to run the targeted suite, but this environment does not yet have repo dependencies installed (`tsx: command not found` during `npm run test` pretest).

Command attempted:
```bash
npm run test -- components/input/__tests__/index.test.tsx -t "custom input integrations"
```

Expected verification on CI for full dependency environment.

### DIFF
--- a/components/input/Input.tsx
+++ b/components/input/Input.tsx
@@ -226,6 +226,19 @@ const Input = forwardRef<InputRef, InputProps>((props, ref) => {
 
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     removePasswordTimeout();
+
+    const input = inputRef.current?.input;
+    if (input && e.target !== input) {
+      e = Object.create(e, {
+        target: {
+          value: input,
+        },
+        currentTarget: {
+          value: input,
+        },
+      });
+    }
+
     onChange?.(e);
   };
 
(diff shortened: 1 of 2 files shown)


afc163:

Issue #52551, opened 2025-01-23
AutoComplete customize input size


IndProgo:
### Reproduction link

[![Edit on CodeSandbox](https://codesandbox.io/static/img/play-codesandbox.svg)](https://codesandbox.io/p/sandbox/customize-input-component-antd-5-23-2-forked-grw9hy?file=%2Fdemo.tsx)

### Steps to reproduce

Switch between small, middle and large

### What is expected?

AutoComplete wrapper should size to custom Input's height

### What is actually happening?

* When size prop is set only on the custom Input, AutoComplete wrapper itself keeps fixed height. 
* When size prop is set on both Input and AutoComplete,then it works almost as it should, except at large size AutoComplete height is larger the Input's, also it shows the warning in console: Warning: [antd: AutoComplete] You need to control style self instead of setting `size` when using customize input. 

| Environment | Info |
| --- | --- |
| antd | 5.23.2 |
| React | 18 |
| System | Windows |
| Browser | Firefox |

zombieJ:
Seems can add `height: auto` when using customize input.

IndProgo:
`height: auto` hack won't work anymore in 5.26.6 for this

As inner Select component doesn't receive the size prop and doesn't read it from the custom Input props, 
the updated `.ant-select-selection-wrap::after` rule doesn't change its line-height

https://github.com/ant-design/ant-design/blob/7f47601b8836d51a13eed6dd22ba1e8ee16adeee/components/select/style/single.ts#L31-L33

Need to reset the `line-height` additionally to `height: auto`

```css
.autocomplete-with-input {
    height: auto;
    .ant-select-selection-wrap::after {
        line-height: 1 !important;
    }
}
```

https://codesandbox.io/p/devbox/customize-input-component-antd-5-23-2-forked-x7n28r?file=%2Fdemo.tsx&workspaceId=ws_91R53CiJCYtGvKdAxJyPAd

PR #57107, opened 2026-03-01
fix(select): fix height mismatch in customize mode with sized custom input

### LINKED ISSUE #57065: AutoComplete demo input height bug
Image

https://preview-57030-ant-design.surge.sh/components/auto-complete-cn#auto-complete-demo-certain-category

左右高度不对齐。

Nicknames96:
When AutoComplete wraps a custom Input component with a size prop (e.g. Input.Search size="large"), the internal Select wrapper can cause height misalignment between parts of the custom input.

The customize mode strips border and padding but still inherits the Select default sizing constraints. The content::before pseudo-element (used as a height spacer in normal mode) is unnecessary in customize mode and can interfere with the custom input layout.

Changes:
- Add height: auto so the wrapper defers to the custom input height
- Hide content ::before spacer in customize mode

Closes #57065
Related: #52551

### DIFF
--- a/components/select/style/select-input-customize.ts
+++ b/components/select/style/select-input-customize.ts
@@ -12,6 +12,7 @@ const genSelectInputCustomizeStyle: GenerateStyle<SelectToken, CSSObject> = (tok
       padding: 0,
       fontSize: 'inherit',
       lineHeight: 'inherit',
+      height: 'auto',
 
       [`${componentCls}-placeholder`]: {
         display: 'none',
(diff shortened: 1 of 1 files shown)


QDyanbing:
image

It seems the issue is still reproducible in Firefox.

You can check it by clicking the Preview button and inspecting the component there. The problem still exists and does not appear to be fully fixed.

Also, based on my testing, this issue specifically occurs on the lg size input (the internal input element), not other sizes.

Could you please double-check this in Firefox?

Nicknames96:
Good catch, thanks for testing in Firefox. The issue is that the physical shorthand `padding: 0` does not override the logical `paddingBlock` property set by the base styles — Firefox treats them as independent property groups. For lg size the computed paddingBlock is about 7px, which causes the visible mismatch.

Pushed a fix that explicitly resets `paddingBlock` and `paddingInline` to 0. Should resolve it for all sizes in Firefox now.

QDyanbing:
image

still exists

Nicknames96:
Closing this — no longer pursuing this change. Thanks for your time.

PR #57133, opened 2026-03-03
fix(transfer): handle non-string render results when rendering items

WustLCQ:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix

### 🔗 Related Issues

### 💡 Background and Solution

__Background:__ When using the Transfer component with `showSearch` enabled and the `render` prop returning JSX elements, a JavaScript error occurs during search operations. The root cause is that the `renderItem` function incorrectly treats JSX elements as strings when extracting text for search filtering.

__Problem:__

- Transfer component crashes when `render` prop returns JSX and `showSearch` is enabled
- The original code forcefully casts JSX elements to strings in `renderItem` function
- This causes type errors and breaks the search functionality

__Solution:__

- Added proper type checking for render function return values

- Implemented `getTextFromReactElement` helper function to extract text content from React elements

- Enhanced `renderItem` function to handle different return types:

  - Plain objects (existing behavior)
  - String values (existing behavior)
  - React elements (new: extract text content)
  - Fallback to empty string for other types

[Minimal Reproducible Example](https://codesandbox.io/p/sandbox/ji-ben-yong-fa-antd-5-26-5-forked-pnvz69)

image

### 📝 Change Log

🇨🇳 Chinese
| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English  |Enhanced transfer section rendering so JSX results are handled gracefully and we now fall back to item metadata for filter text.|
| 🇨🇳 Chinese |改进了穿梭框的渲染逻辑，兼容 JSX 返回值并在需要时回退到条目的元信息以保证过滤文本可用。|

### DIFF
--- a/components/transfer/Section.tsx
+++ b/components/transfer/Section.tsx
@@ -41,14 +41,10 @@ function getTextFromRenderResult<RecordType extends KeyWiseTransferItem>(
   renderResult: RenderResult,
   item: RecordType,
 ): string {
-  if (typeof renderResult === 'string' || typeof renderResult === 'number') {
-    return String(renderResult);
-  }
-  if (typeof item.title === 'string' || typeof item.title === 'number') {
-    return String(item.title);
-  }
-  if (typeof item.key === 'string' || typeof item.key === 'number') {
-    return String(item.key);
+  for (const v of [renderResult, item.title, item.key]) {
+    if (typeof v === 'string' || typeof v === 'number') {
+      return String(v);
+    }
   }
   return '';
 }
@@ -200,16 +196,10 @@ const TransferSection = <RecordType extends KeyWiseTransferItem>(
   const renderItem = (item: RecordType): RenderedItem<RecordType> => {
     const renderResult = render(item);
     const isRenderResultPlain = isRenderResultPlainObject(renderResult);
-    let renderedEl: React.ReactNode;
-    let renderedText: string;
-
-    if (isRenderResultPlain) {
-      renderedEl = renderResult.label;
-      renderedText = renderResult.value;
-    } else {
-      renderedEl = renderResult;
-      renderedText = getTextFromRenderResult(renderResult, item);
-    }
+    const renderedEl = isRenderResultPlain ? renderResult.label : renderResult;
+    const renderedText = isRenderResultPlain
+      ? renderResult.value
+      : getTextFromRenderResult(renderResult, item);
 
     return {
       item,
--- a/components/transfer/Section.tsx
+++ b/components/transfer/Section.tsx
@@ -37,6 +37,22 @@ function getEnabledItemKeys<RecordType extends KeyWiseTransferItem>(items: Recor
   return items.filter((data) => !data.disabled).map((data) => data.key);
 }
 
+function getTextFromRenderResult<RecordType extends KeyWiseTransferItem>(
+  renderResult: RenderResult,
+  item: RecordType,
+): string {
+  if (typeof renderResult === 'string' || typeof renderResult === 'number') {
+    return String(renderResult);
+  }
+  if (typeof item.title === 'string' || typeof item.title === 'number') {
+    return String(item.title);
+  }
+  if (typeof item.key === 'string' || typeof item.key === 'number') {
+    return String(item.key);
+  }
+  return '';
+}
+
 const isValidIcon = (icon: React.ReactNode) => icon !== undefined;
 
 export interface RenderedItem<RecordType> {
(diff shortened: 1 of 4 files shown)


afc163:
Thanks for the fix\! This solves the immediate crash issue when using showSearch with render prop returning JSX elements.

## Review Summary

The PR correctly handles different render prop return types:
- Plain object: uses label for display, value for search text
- String/Number: uses the value directly  
- React Element: falls back to item.title or item.key for search text

## A Potential Improvement

The PR description mentions implementing a getTextFromReactElement helper to extract text from JSX, but it was not included in the final diff. Consider extracting text from React elements instead of just falling back to title/key:

```js
const getTextFromReactElement = (element: React.ReactElement): string => {
  const extractText = (node: React.ReactNode): string => {
    if (typeof node === 'string' || typeof node === 'number') {
      return String(node);
    }
    if (Array.isArray(node)) {
      return node.map(extractText).join('');
    }
    if (node?.props?.children) {
      return extractText(node.props.children);
    }
    return '';
  };
  return extractText(element.props.children);
};
```

This would allow users to search within the actual rendered JSX content rather than just falling back to metadata. The current fallback to title/key is reasonable but limits search functionality when using JSX.

## Minor Nitpick

As CodeRabbit mentioned, the fallback text logic could be extracted into a helper function for better readability.

Overall the fix works well and prevents the crash. The text extraction from JSX would be a nice enhancement but not required for merging.

components/transfer/Section.tsx
```diff
@@ -184,10 +184,30 @@ const TransferSection = <RecordType extends KeyWiseTransferItem>(
...
+
+      if (typeof renderResult === 'string' || typeof renderResult === 'number') {
+        renderedText = String(renderResult);
+      } else if (typeof item.title === 'string' || typeof item.title === 'number') {
+        renderedText = String(item.title);
+      } else if (typeof item.key === 'string' || typeof item.key === 'number') {
+        renderedText = String(item.key);
+      } else {
+        renderedText = '';
+      }
```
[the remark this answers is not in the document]
WustLCQ: Done!

components/transfer/Section.tsx
```diff
@@ -37,6 +37,22 @@ function getEnabledItemKeys<RecordType extends KeyWiseTransferItem>(items: Recor
...
+    return String(renderResult);
+  }
+  if (typeof item.title === 'string' || typeof item.title === 'number') {
+    return String(item.title);
+  }
+  if (typeof item.key === 'string' || typeof item.key === 'number') {
+    return String(item.key);
+  }
+  return '';
+}
```
afc163, line 50:
```js
function getTextFromRenderResult<RecordType extends KeyWiseTransferItem>(
  renderResult: RenderResult,
  item: RecordType,
): string {
  for (const v of [renderResult, item.title, item.key]) {
    if (typeof v === 'string' || typeof v === 'number') {
      return String(v);
    }
  }
  return '';
}
```

components/transfer/Section.tsx
```diff
@@ -184,10 +200,21 @@ const TransferSection = <RecordType extends KeyWiseTransferItem>(
...
+    let renderedEl: React.ReactNode;
+    let renderedText: string;
+
+    if (isRenderResultPlain) {
+      renderedEl = renderResult.label;
+      renderedText = renderResult.value;
+    } else {
+      renderedEl = renderResult;
+      renderedText = getTextFromRenderResult(renderResult, item);
+    }
```
afc163:
```suggestion
const renderedEl = isRenderResultPlain ? renderResult.label : renderResult;
const renderedText = isRenderResultPlain
  ? renderResult.value
  : getTextFromRenderResult(renderResult, item);
```

components/transfer/Section.tsx
WustLCQ: Thanks for the suggestion. I've refactored this part as you suggested.

components/transfer/Section.tsx
WustLCQ: Done. Thanks!

afc163:
ci broken

thinkasany (dismissed):

afc163:
图片

Eliot00:
我也遇到了这个问题，后来发现`render`返回一个对象，value是字符串就可以搜索了，可以在文档特别说明下

> 每行数据渲染函数，该函数的入参为 dataSource 中的项，返回值为 ReactElement。或者返回一个普通对象，其中 label 字段为 ReactElement，value 字段为 title

afc163 (approved):

PR #57142, opened 2026-03-03
fix(Splitter): set partially defined panel sizes correctly.

js0753:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 💡 Background and Solution

When setting the panel sizes in a controlled way and updating them by setting some of the `size` values while other panels' `size` are set to undefined, the panels which have a defined size in the updated array do not get set correctly. Instead, the panels with undefined sizes use their inner values, and this leads to either:

a) The total of all sizes becoming greater than 100%, in which case all panels have their sizes decreased in a scaled manner.
b) The total of all sizes becoming less than 100%, in which case all panels get filled with extra size in a scaled manner.

The proposed solution, is to only use innerSizes when no panel gets a value for the `size` prop, this ensures that the panel that has it's size explicitly set, ends up with the correct size. The other panels having their size recalculated is understandable and intuitive. If no panel has received a `size` prop, then use the `innerSizes`.

Example : 
```
import React, { useState } from 'react';

import { Flex, Splitter, Typography, Button } from 'antd';

export const Desc: React.FC<Readonly<{ text?: string | number }>> = (props) => (
  <Flex justify="center" align="center" style={{ height: '100%' }}>
    <Typography.Title
      type="secondary"
      level={5}
      style={{ whiteSpace: 'nowrap' }}
    >
      {props.text}
    </Typography.Title>
  </Flex>
);

const App: React.FC = () => {
  const [sizes, setSizes] = useState(['50%', undefined]);
  return (
    <>
      <Button
        style={{ 'margin-bottom': 4 }}
        onClick={() => setSizes(['50%', undefined])}
      >
        Reset Panels
      </Button>
      <Splitter
        style={{ height: 200, boxShadow: '0 0 10px rgba(0, 0, 0, 0.1)' }}
        onResize={setSizes}
      >
        <Splitter.Panel resizable={true} size={sizes[0]}>
          <Desc text="First" />
        </Splitter.Panel>
        <Splitter.Panel size={sizes[1]}>
          <Desc text="Second" />
        </Splitter.Panel>
      </Splitter>
    </>
  );
};

export default App;
```

Before the change, this would not have reset the first panel to be at "50%" size, despite the first panel being explicitly set to be "50%".
It may be have been easy to work around this by setting the value for the second panel to be (100-x) %, but that's because this example has only two panels. Imagine five, or even ten panels, having to calculate or define a value for each panel would add a lot of complexity and would frankly be redundant given this calculation is already done in the `autoPtgSizes` function in `ant-design/components/splitter/hooks/sizeUtil.ts`, it was just not being used correctly due to `innerSizes` being used for the panels that had undefined sizes.

### 📝 Change Log

- Update the `sizes` calculation in `useSizes` to only use `innerSizes` when no Panel has received a value for the `size` prop.

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Splitter : Update the sizes calculation in `useSizes` to only use `innerSizes` when no Panel has received a value for the size prop. |
| 🇨🇳 Chinese | Splitter : 更新 `useSizes` 中的尺寸计算，使其仅在没有 Panel 收到 size 属性值时才使用 `innerSizes`。|


components/splitter/hooks/useSizes.ts
```diff
@@ -29,13 +29,9 @@ export default function useSizes(items: PanelProps[], containerSize?: number) {
...
-    const mergedSizes: PanelProps['size'][] = [];
-
-    for (let i = 0; i < itemsCount; i += 1) {
-      mergedSizes[i] = propSizes[i] ?? innerSizes[i];
-    }
-
-    return mergedSizes;
+    // If any panel has a `size` passed as a prop, use `propSizes` and calculate all other panel sizes that don't have a `size` defined by the prop.
+    // If no panel has received a value for the `size` prop, use `innerSizes`.
+    return propSizes.some((propSize) => propSize != null) ? propSizes : innerSizes;
```
afc163:
```suggestion
    return propSizes.some((propSize) => propSize !== null) ? propSizes : innerSizes;
```

afc163: Could you add some test cases for this?

js0753: Thanks @afc163 for taking a look! The loose equality check ( `!=` ) here is to ensure we catch both `null` and `undefined`, so equivalent too : 

```
propSize !== null && propSize !== undefined
```

Do you think we can use `!=` in this case?

afc163: We don't use `!=` in any situation.

js0753: Thanks @afc163 , I have added a test in the new commit.

js0753: Thanks @afc163 , I have updated to use `!==` in the new commit.

afc163 (approved):

js0753:
Thanks @afc163 for approving! There seem to be two failing checks, but I am not sure how to resolve those. I did not even touch `color-picker-controlled.dark.png` so I am not sure how there is a visual diff there. How can I fix this and merge the PR ? 

Thanks.

thinkasany:
> Thanks @afc163 for approving! There seem to be two failing checks, but I am not sure how to resolve those. I did not even touch `color-picker-controlled.dark.png` so I am not sure how there is a visual diff there. How can I fix this and merge the PR ?
> 
> Thanks.

That's indeed unrelated to your PR and can be ignored.

meet-student (approved):

PR #57216, opened 2026-03-08
fix(Input): normalize onChange event.target to real mounted element

### LINKED ISSUE #46999: Antd Input: To use with react-number-format customInput feature.

AyushPatel94:
## Summary

  Fixes #46999 — `Input` is incompatible with `react-number-format`'s `customInput` feature since v5.13.3.

  ## Root Cause

  `@rc-component/input`'s `resolveOnChange` calls `cloneEvent()` which does:

  ```ts
  const currentTarget = target.cloneNode(true); // detached clone
  const newEvent = Object.create(event, {
    target: { value: currentTarget },
    currentTarget: { value: currentTarget },
  });

  This means e.target in every onChange handler is a detached DOM node, not the mounted <input>. Libraries like
  react-number-format that use customInput call document.contains(e.target) and perform identity checks that fail
  silently against a detached clone, causing cursor jumps and broken input behaviour.

  Fix

  In handleChange, detect when e.target !== realInput and wrap the event in a thin Object.create override restoring
  target/currentTarget to the real element.

  Why not clone nativeEvent? PR #57032 tried that but reviewers noted it causes "Illegal invocation" because
  Object.create(nativeEvent) is not a real Event instance. This fix leaves nativeEvent untouched (inherited via
  prototype chain from the original synthetic event).

  Test

  Added unit test asserting e.target === container.querySelector('input') and document.contains(e.target) === true after
   fireEvent.change.


<!-- Issuehunt content -->

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#46999: Antd Input: To use with react-number-format customInput feature.](https://oss.issuehunt.io/repos/34526884/issues/46999)
---
</details>
<!-- /Issuehunt content-->

### DIFF
--- a/components/input/Input.tsx
+++ b/components/input/Input.tsx
@@ -226,7 +226,27 @@ const Input = forwardRef<InputRef, InputProps>((props, ref) => {
 
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     removePasswordTimeout();
-    onChange?.(e);
+    if (onChange) {
+      const realInput = inputRef.current?.input;
+      // @rc-component/input's resolveOnChange passes a cloneEvent()-created object
+      // whose target/currentTarget point to a detached DOM clone (not the mounted
+      // element). Third-party libraries such as react-number-format that use
+      // customInput perform DOM operations (e.g. document.contains, identity
+      // checks) against e.target and fail silently when it is detached.
+      //
+      // We normalise the event by re-pointing target/currentTarget to the real
+      // input. We deliberately do NOT clone nativeEvent to avoid breaking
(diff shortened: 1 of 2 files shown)


afc163:
This issue has been fixed in antd v6. See https://github.com/ant-design/ant-design/issues/46999#issuecomment-4507864963 for details.

PR #57241, opened 2026-03-10
feat: use @ant-design/static-style-extract for style extraction

zombieJ:
## PR Type
Feature / Enhancement

## What is the new feature?
Replace manual style extraction logic with `@ant-design/static-style-extract` package

## Why is this change needed?
- Simplify build-style script by removing ~130 lines of boilerplate code
- Use dedicated package for static style extraction instead of manual implementation
- Add dynamic import support to allow package to resolve antd module correctly

## How does it work?
- Use `@ant-design/static-style-extract` package's `extractStyle` function
- Hijack module resolution to allow 'antd' imports to resolve to local components
- Use dynamic import to load the extract package at runtime

## Related issues
None

## Checklist
- [ ] Test passed
- [ ] Demo updated
- [ ] TypeScript definition updated

## CLA
- [X] I have signed the CLA


li-jia-nan (dismissed):

afc163:
visual diff 老是挂，最好能解决一下。

thinkasany (approved):

PR #57332, opened 2026-03-17
feat(Flex): add childrenFlex prop to apply flex CSS to direct children

### LINKED ISSUE #56860: Flex component new API property

Right now the flex property of the Flex component sets the flex container's CSS property. This is only useful if the Flex component is nested within another Flex component. It would be more logical if there would be a property that applied the given rule to its children. For instance, if there is a horizontal Flex with Card components inside, a childrenFlex={1} property of the Flex component could apply this to the Cards inside it, so that all Cards would occupy the same space, independently of the number of cards.

Add a childrenFlex property which would apply the passed value to the flex CSS rule of its children.

fury0928:
### 🤔 This is a ...

- [x] 🆕 New feature
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

close #56860

### 💡 Background and Solution

The Flex component's existing `flex` prop sets the flex CSS property on the container itself, which is only useful when the Flex component is nested inside another flex container. There was no way to uniformly apply a flex value to all direct children.

This PR adds a `childrenFlex` prop that applies the given flex CSS shorthand to all direct children of the Flex container using a CSS variable + child selector approach (`> *`), avoiding direct style injection on child elements (per maintainer feedback from zombieJ).

**API:**

```tsx
// All children will have flex: 1, occupying equal space
<Flex childrenFlex={1}>
  <Card>Card 1</Card>
  <Card>Card 2</Card>
  <Card>Card 3</Card>
</Flex>

// Supports full flex shorthand
<Flex childrenFlex="1 0 auto">
  <div>Item 1</div>
  <div>Item 2</div>
</Flex>
```

**Implementation:**
1. Added `childrenFlex` prop to `FlexProps` interface (type: `React.CSSProperties['flex']`)
2. When set, adds `ant-flex-children-flex` class and sets `--ant-flex-children-flex` CSS variable on the container
3. CSS rule `&-children-flex > * { flex: var(--ant-flex-children-flex) }` applies the value to all direct children
4. Added 3 unit tests covering numeric values, string values, and absence behavior

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🆕 Add `childrenFlex` prop to Flex component for applying flex CSS to all direct children. |
| 🇨🇳 Chinese | 🆕 为 Flex 组件新增 `childrenFlex` 属性，支持对所有直接子元素统一设置 flex CSS 简写属性。 |

### DIFF
--- a/components/flex/__tests__/index.test.tsx
+++ b/components/flex/__tests__/index.test.tsx
@@ -106,6 +106,42 @@ describe('Flex', () => {
     });
   });
 
+  // ============================= childrenFlex =============================
+  describe('Props: childrenFlex', () => {
+    it('should add children-flex class and CSS variable', () => {
+      const { container } = render(
+        <Flex childrenFlex={1}>
+          <div>child1</div>
+          <div>child2</div>
+        </Flex>,
+      );
+      const flexEl = container.querySelector('.ant-flex');
+      expect(flexEl).toHaveClass('ant-flex-children-flex');
+      expect(flexEl).toHaveStyle({ '--ant-flex-children-flex': '1' });
+    });
+
+    it('should support string value', () => {
+      const { container } = render(
+        <Flex childrenFlex="1 0 auto">
+          <div>child1</div>
+          <div>child2</div>
+        </Flex>,
+      );
+      const flexEl = container.querySelector('.ant-flex');
+      expect(flexEl).toHaveClass('ant-flex-children-flex');
+      expect(flexEl).toHaveStyle({ '--ant-flex-children-flex': '1 0 auto' });
+    });
+
+    it('should not add class when childrenFlex is not set', () => {
+      const { container } = render(
+        <Flex>
+          <div>child1</div>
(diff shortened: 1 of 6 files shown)


fury0928:
Hi @thinkasany nice to meet you
Would you review my PR when you have a moment?
Thank you

eureka0928:
Hi @thinkasany @li-jia-nan @meet-student @afc163 would you review this?
This was open for one month
Thank you

afc163 (changes requested):
This is a new feature and should target `feature`, not `master`. Please also add the API version metadata in the correct table order and rebase.

PR #57343, opened 2026-03-18
fix: improve Popover and Popconfirm keyboard focus

claytonlin1110:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

fix #57333
### 💡 Background and Solution
- Problem: Opening Popover or Popconfirm did not move keyboard focus into the overlay, so keyboard/screen-reader users could not reach interactive content (e.g. OK/Cancel in Popconfirm). This fails WCAG 2.4.3 (Focus Order) and 1.3.1.
-  Solution: In Tooltip, when used by Popover/Popconfirm with click trigger, we now:
- Scope: Only applies when data-popover-inject is set and trigger is click (Popconfirm default and Popover with trigger="click"). Hover-only tooltips are unchanged.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | - ⌨️ Improve Popover and Popconfirm accessibility by moving focus into the overlay when opened via click and restoring focus to the trigger on close (WCAG 2.4.3) |
| 🇨🇳 Chinese | - ⌨️ 改进 Popover、Popconfirm 可访问性：点击打开时焦点移入浮层，关闭时焦点回到触发元素（符合 WCAG 2.4.3）|


claytonlin1110:
@afc163 @meet-student Please review

aojunhao123:
From my perspective, for accessibility-related changes like this, it’s better to first go through the full context before jumping into implementation, rather than applying a localized fix based on a single issue.

Also, this doesn’t seem like something that can be fully addressed by simply adding a patch on the antd side. The actual interaction scenarios and edge cases involved are likely more complex than what’s described in the original issue.

It might be better to first align on the overall design and scope, and then move forward with the implementation. That would also help reduce maintenance costs in the long run.

claytonlin1110:
@aojunhao123 
The issue #57333 is specifically about keyboard focus order for Popover/Popconfirm when opened via keyboard/click: focus should move into the overlay, allow Tab navigation within it, and restore focus to the trigger on close. The current behavior leaves focus on the trigger, so Tab skips the overlay entirely.

What I implemented is intentionally scoped to Popover/Popconfirm click-triggered overlays (via the data-popover-inject path) and does not change hover-only Tooltip behavior. It’s meant as a minimal improvement for the reported WCAG failure (2.4.3 Focus Order).

claytonlin1110:
@aojunhao123 
If you’re open to it, I can update the PR description to explicitly state this scoped behavior and add a short "Design" section (when/where/restore rules). If we want a more comprehensive approach, we should do a design pass across overlay components and possibly coordinate with rc-trigger for a shared focus-management hook/API.

claytonlin1110:
@aojunhao123 any feedback for this?

aojunhao123:
Sorry for the late reply. Work has been quite busy these past few days, so my response on this issue may take a bit longer. Focus management is not as straightforward as described in the [original issue](https://github.com/ant-design/ant-design/pull/57333), and I need some time to think through a complete solution carefully.

claytonlin1110:
> Sorry for the late reply. Work has been quite busy these past few days, so my response on this issue may take a bit longer. Focus management is not as straightforward as described in the [original issue](https://github.com/ant-design/ant-design/pull/57333), and I need some time to think through a complete solution carefully.

Thanks for the update, and no worries about the delay.
To make next steps easier once you’re ready, I can prepare a short design note covering:

- target behavior on open/close (including restore-focus rules)
- applicable components/triggers
- key edge cases (nested overlays, programmatic open, container/portal cases)
- and a minimal phased rollout plan
-
Happy to follow whichever direction you think is best.

claytonlin1110:
@aojunhao123 any feedback?

claytonlin1110:
@aojunhao123

claytonlin1110:
@meet-student would you mind review this? thanks.

meet-student:
> @meet-student would you mind review this? thanks.

Hi, hello. My suggestion is that you need to read the @react-component component and modify the underlying dependencies of @react-component in the issue to solve this issuse in the best way. In antd, only this dependency is upgraded and api calls are made. This plan is exactly what we expect. If you need any help, you can contact me alone

claytonlin1110:
> > @meet-student would you mind review this? thanks.
> 
> Hi, hello. My suggestion is that you need to read the @react-component component and modify the underlying dependencies of @react-component in the issue to solve this issuse in the best way. In antd, only this dependency is upgraded and api calls are made. This plan is exactly what we expect. If you need any help, you can contact me alone

Thanks for the suggestion. I agree upstream could be ideal, but here the requirement is specific to Popover/Popconfirm interaction semantics (move focus into the overlay on open and restore on close). @rc-component/tooltip is a generic primitive used broadly, so adding Popover-specific focus logic there would either need a new contract/API or brittle heuristics, which risks regressions elsewhere.

meet-student:
> > > @meet-student would you mind review this? thanks.您能帮忙审阅一下吗？谢谢。
> > 
> > 
> > Hi, hello. My suggestion is that you need to read the @react-component component and modify the underlying dependencies of @react-component in the issue to solve this issuse in the best way. In antd, only this dependency is upgraded and api calls are made. This plan is exactly what we expect. If you need any help, you can contact me alone您好。我的建议是，您需要仔细阅读组件文档，并修改问题中提到的底层依赖项，以最佳方式解决此问题。在 Antd 中，仅升级了此依赖项并进行了 API 调用。这个方案正是我们所期望的。如果您需要任何帮助，可以随时联系我。
> 
> Thanks for the suggestion. I agree upstream could be ideal, but here the requirement is specific to Popover/Popconfirm interaction semantics (move focus into the overlay on open and restore on close). @rc-component/tooltip is a generic primitive used broadly, so adding Popover-specific focus logic there would either need a new contract/API or brittle heuristics, which risks regressions elsewhere.感谢您的建议。我同意上游方案可能更理想，但这里的需求是针对 Popover/Popconfirm 交互语义的（打开时将焦点移至覆盖层，关闭时恢复焦点）。@rc-component/tooltip 是一个广泛使用的通用原语，因此在其中添加 Popover 特有的焦点逻辑要么需要新的接口/API，要么需要脆弱的启发式方法，这可能会导致其他地方出现回归问题。



> > > @meet-student would you mind review this? thanks.
> > 
> > 
> > Hi, hello. My suggestion is that you need to read the @react-component component and modify the underlying dependencies of @react-component in the issue to solve this issuse in the best way. In antd, only this dependency is upgraded and api calls are made. This plan is exactly what we expect. If you need any help, you can contact me alone
> 
> Thanks for the suggestion. I agree upstream could be ideal, but here the requirement is specific to Popover/Popconfirm interaction semantics (move focus into the overlay on open and restore on close). @rc-component/tooltip is a generic primitive used broadly, so adding Popover-specific focus logic there would either need a new contract/API or brittle heuristics, which risks regressions elsewhere.

@rc-component/trigger

claytonlin1110:
@meet-student I created PR on trigger repo.

afc163:
Confirmed that the requested upstream work is now tracked in react-component/trigger#613 and was updated again on July 21. Keeping this PR open rather than treating it as abandoned. The next action is to finish and merge the upstream focus-management contract, then rebase this PR, upgrade the dependency, and reduce the Ant Design change to the Popover/Popconfirm integration plus focused accessibility coverage.

nrps9909 (changes requested):
The new focus transfer can leave focus on the trigger even when the Popover contains a valid visible control, because `getFirstFocusable` returns the first selector match without checking whether it is actually focusable.

I reproduced this on exact head `b2df3142e6b7738c2916125632688dab6ae87e4c`: a click-triggered Popover whose content is `<button hidden>Hidden</button><button>Visible</button>` opens normally, but after the scheduled focus frame the visible button is not focused and `document.activeElement` remains the trigger. The hidden button matches `button:not([disabled])`; calling `focus()` on it has no effect, and the implementation never tries the next candidate. The same regression also reproduces after applying the PR’s one-file patch to current master `e315024afad5ff69075c8ccf1ab34b99a9ef745a`.

The repository already has the shared `getFocusNodeList` helper in `@rc-component/util/lib/Dom/focus`; it filters candidates through the canonical visibility, disabled, editable, anchor, and tabindex rules. Please use that (or equivalent real focusability filtering) and commit focus-management tests covering at least hidden-first/visible-second content, the no-focusable fallback, cancellation during a rapid close, and trigger restoration. The current GitHub diff changes only `components/tooltip/index.tsx`, so none of the claimed focus behavior has a regression test yet. I removed my temporary test and both worktrees were clean after verification.

PR #57345, opened 2026-03-18
type: autoAdjustOverflow supports object config

Damon4:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

### 💡 Background and Solution

`autoAdjustOverflow` already supports object-based overflow configuration at runtime, but the public typings and docs were narrower than the supported shape.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Tooltip, Popover, and Dropdown `autoAdjustOverflow` typing and documentation to support object overflow configuration such as `shiftX` and `shiftY`. |
| 🇨🇳 Chinese | 🐞 修复 Tooltip、Popover 和 Dropdown 的 `autoAdjustOverflow` 类型与文档，支持 `shiftX` 和 `shiftY` 等对象配置。 |


afc163 (changes requested):
Please update the PR title to the required English type format, add the API version metadata, and rebase the overflow typing change.

Damon4:
> Please update the PR title to the required English type format, add the API version metadata, and rebase the overflow typing change.

done

afc163 (dismissed):
The follow-up now rebases the change, documents the 6.6.0 object form, and covers the widened runtime-supported type across Tooltip, Popover, and Dropdown. Approved; the visual workflow can complete independently.

components/_util/placements.ts
```diff
@@ -3,9 +3,8 @@ import type { AlignType, BuildInPlacements } from '@rc-component/trigger';
 import { getArrowOffsetToken } from '../style/placementArrow';
 import { isPlainObject } from './is';
 
-export interface AdjustOverflow {
-  adjustX?: 0 | 1;
-  adjustY?: 0 | 1;
+export interface AdjustOverflow extends NonNullable<AlignType['overflow']> {
+  alwaysByViewport?: boolean;
```
nrps9909: **Blocking:** please do not expose and document `alwaysByViewport` as supported until the current alignment engine implements it. At this exact head, Ant Design forwards the key, but `@rc-component/trigger` 3.10.1's `AlignType['overflow']` and `useAlign` only define/read `adjustX`, `adjustY`, `shiftX`, and `shiftY`; `alwaysByViewport` has no consumer. The upstream request for this capability is still open: https://github.com/react-component/trigger/issues/414. I reproduced the gap with a viewport of 100 and document width of 1000: `alwaysByViewport: false` and `true` both produced the same viewport flip from right to left, whereas the actually supported `htmlRegion: 'scroll'` kept the right placement. The new smoke/forwarding assertions therefore pass even though the advertised option cannot affect behavior. Please either remove `alwaysByViewport` from the new type/docs/tests/changelog while retaining the valid `shiftX`/`shiftY` widening, or first implement/upgrade the upstream behavior and cover that behavior here.

nrps9909 (changes requested):
I found one blocking runtime-contract issue on exact head `3d9201ff9edd764d3e89b9a3c16b5e5fdfbd9f44`; details and a minimal remediation path are attached inline.

Validation performed from a fresh detached worktree: the complete Tooltip, Popover, and Dropdown suites passed (3 suites, 97 tests, 10 snapshots); an independent alignment probe passed and reproduced the no-op behavior; changed TypeScript files passed ESLint and Biome; all four changed Markdown files passed Remark; and `git diff --check` passed. Repo-wide `tsc` reported the same six unrelated table-demo `FullToken.antCls` errors on both the exact base and head with the same dependency tree, so there is no PR-introduced type-check delta. The remaining visual status is pending, while the code-related CI checks are green.

AI assistance disclosure: Codex was used to inspect the exact-head diff and dependency implementation, run the base/head checks and focused suites, construct the independent geometry probe, and draft this review. I verified the cited behavior and live PR state before submitting.

components/_util/placements.ts
Damon4: Thanks, you're right — fixed in d41d7c7.

I confirmed the gap independently: `@rc-component/trigger` 3.10.1 (current `latest`, and the version this repo pins) declares `AlignType['overflow']` as `adjustX | adjustY | shiftX | shiftY` only, and `grep -rn alwaysByViewport node_modules/@rc-component/{trigger,util}` returns nothing, so the key had no consumer at any level. react-component/trigger#414 is still open with no activity since 2023, so there is nothing to wait for here.

Changes:

- `AdjustOverflow` is now `NonNullable<AlignType['overflow']>` — it no longer invents a field the alignment engine cannot honor, while keeping the `shiftX`/`shiftY` widening, which is real.
- Removed `alwaysByViewport` from the four docs tables (Tooltip shared props and Dropdown, en/zh), both the prose and the inline type.
- Removed it from the Tooltip, Popover and Dropdown tests, including the `getPlacements` `toEqual` assertion — you were right that the shape assertion would have passed for any invented key.
- Updated the changelog entries in the PR description accordingly.

Verification: Tooltip/Popover/Dropdown suites pass (3 suites, 97 tests, 10 snapshots), `tsc --noEmit` is clean for the touched files, ESLint/Biome/Remark pass, and no `alwaysByViewport` occurrences remain under `components/`, `docs/` or `.dumi/`.

nrps9909 (approved):
Re-reviewed exact head `334bff76601c971c5224be707be6db1f47070b93`. The follow-up removes `alwaysByViewport` from the exported object contract, documentation, and all examples, so the no-op runtime option I previously blocked is no longer advertised. The remaining type is now derived directly from `NonNullable<AlignType['overflow']>`, and the focused `getPlacements` assertion proves that `adjustX`, `adjustY`, `shiftX`, and `shiftY` survive into every generated placement.

Exact-head validation passes: Tooltip, Popover, and Dropdown complete suites (3 suites, 97 tests, 10 snapshots); ESLint and Biome on every changed TypeScript file; Remark on all four changed API documents; and `git diff --check`. Repo-wide TypeScript reaches one unrelated pre-existing Puppeteer `Page` type mismatch in `tests/shared/imageTest.tsx`; it reports no changed-file error. The worktree is clean. Approved.

AI assistance disclosure: Codex was used to inspect the updated exact-head diff, verify that the unsupported option was fully removed, run the focused suites and static checks, and draft this re-review. I verified the cited SHA and results before submitting.

PR #57407, opened 2026-03-23
chore: bump dayjs to 1.11.20

verbetchii:
### 🤔 This is a ...

- [ ] 🆕 New feature  
- [ ] 🐞 Bug fix  
- [ ] 📽️ Demo improvement  
- [ ] ⭐️ Feature enhancement  
- [ ] 🌐 Internationalization  
- [ ] 🛠 Refactoring  
- [ ] ✅ Test Case  
- [ ] 🔀 Branch merge  
- [ ] ⏩ Workflow  
- [ ] ❓ Other (about what?)  

---

### 🔗 Related Issues

N/A

---

### 💡 Background and Solution

Currently, the project uses an older version of `dayjs`.

When a consumer project uses a newer version (e.g. `1.11.20`), TypeScript may report type incompatibility issues because multiple versions of `Dayjs` types are resolved, leading to mismatched types.

This PR updates `dayjs` to the latest version (`1.11.20`) to ensure compatibility with consumer projects and prevent duplicate type definitions.

This is a non-breaking change and aligns the dependency with the latest `dayjs` release.

---

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Bump `dayjs` version to `1.11.20` to avoid type incompatibility issues when multiple versions are installed in consumer projects. |
| 🇨🇳 Chinese | 将 `dayjs` 升级到 `1.11.20`，以避免在使用不同版本时出现类型不兼容问题。 |

### DIFF
--- a/package.json
+++ b/package.json
@@ -150,7 +150,7 @@
     "@rc-component/upload": "~1.1.0",
     "@rc-component/util": "^1.10.0",
     "clsx": "^2.1.1",
-    "dayjs": "^1.11.11",
+    "dayjs": "^1.11.20",
     "scroll-into-view-if-needed": "^3.1.0",
     "throttle-debounce": "^5.0.2"
   },


thinkasany (approved):

package.json
```diff
@@ -150,7 +150,7 @@
     "@rc-component/upload": "~1.1.0",
     "@rc-component/util": "^1.10.0",
     "clsx": "^2.1.1",
-    "dayjs": "^1.11.11",
+    "dayjs": "^1.11.20",
```
afc163, line 153: 一般不要更新 dayjs 版本，使其兼容版本更加广阔，减少和应用 dayjs 版本冲突导致的多实例问题。

verbetchii:
Why this PR was reverted?
Isn't it a common practice to have the latest versions of the packages?
Could you explain why we should stay on version 1.11.11? to not break users with another version? But how about users who need fixes from the latest 1.11.20?
I mean, I don't get this point. If we can't update it now, then when? What should happen to just a bump non-major non-breaking change version to stay up-to date

package.json
verbetchii: So now users with the latest dayjs version(1.11.20) have an issue with dayjs ts incompatibility. When is the best time to simply bump a non-major version?

thinkasany:
> Why this PR was reverted? Isn't it a common practice to have the latest versions of the packages? Could you explain why we should stay on version 1.11.11? to not break users with another version? But how about users who need fixes from the latest 1.11.20? I mean, I don't get this point. If we can't update it now, then when? What should happen to just a bump non-major non-breaking change version to stay up-to date

Generally, you should avoid updating the Day.js version. This broadens its version compatibility range and minimizes multi-instance issues caused by version conflicts with the application.
If the application hardcodes the version to `1.11.1` while antd specifies `^1.11.0`, a single shared version will be maintained; however, if antd specifies `^1.11.20`, this will result in two separate versions.

verbetchii:
> > Why this PR was reverted? Isn't it a common practice to have the latest versions of the packages? Could you explain why we should stay on version 1.11.11? to not break users with another version? But how about users who need fixes from the latest 1.11.20? I mean, I don't get this point. If we can't update it now, then when? What should happen to just a bump non-major non-breaking change version to stay up-to date
> 
> Generally, you should avoid updating the Day.js version. This broadens its version compatibility range and minimizes multi-instance issues caused by version conflicts with the application. If the application hardcodes the version to `1.11.1` while antd specifies `^1.11.0`, a single shared version will be maintained; however, if antd specifies `^1.11.20`, this will result in two separate versions.

Yes, that's valid, and I understand it.
but what users with dayjs v 1.11.20 should do? There is a bench of fixes and what should happen to allow us to upgrade this dayjs version? Only waiting for critical vulnerabilities?

thinkasany:
> 1.11.20

https://stackoverflow.com/questions/22343224/whats-the-difference-between-tilde-and-caret-in-package-json

^version “Compatible with version”, will automatically update you to all future minor/patch versions that are backwards-compatible, without incrementing the major version. ^1.11.11 will use releases from 1.11.11 to < 2.0.0.

PR #57469, opened 2026-03-27
fix(Tag): improve solid variant text contrast in dark theme

### LINKED ISSUE #57029: `Tag` with `variant="solid"` has poor text contrast in dark theme `(v6)`
### Reproduction link

[Edit on StackBlitz](https://stackblitz.com/edit/vitejs-vite-ku6saxev?file=src%2FApp.tsx)

### Steps to reproduce

After upgrading to version `6`, the `Tag` component with `variant="solid"` renders with poor accessibility/readability when using the Dark Theme. The text color does not provide enough contrast against the solid background color in dark mode, making it nearly illegible.

1. Set up an Ant Design v6 project.
1. Configure the `ConfigProvider` with `theme={{ algorithm: theme.darkAlgorithm }}`.
1. Render a `<Tag variant="solid">Solid Tag</Tag>`.
1. Observe the contrast ratio between the text and the background.

### What is expected?

The text color should automatically adapt to ensure WCAG accessibility standards (e.g., switching to white or a very light shade) when `variant="solid"` is used in dark mode.

Image

In the Dark Theme, the `Tag` component with `variant="solid"` renders with an extremely low contrast ratio.

The Problem: The background of the solid tag is being rendered as a very light grey/white (almost #f0f0f0), while the text and icon color are also rendered in a very light grey or white.

Visibility: As shown in the attached screenshots, the text and the icon inside the first Tag (solid) are virtually invisible against the background.

The `variant="outlined"` (second tag) and the custom color tag (third tag) correctly maintain readability by using appropriate dark backgrounds and light text, or vice versa.
The `variant="solid"` seems to incorrectly apply a light background without adjusting the text color to a dark contrast, which is the opposite of what is expected in a dark theme environment.

| Environment | Info |
| --- | --- |
| antd | 6.3.0 |
| React | 19.2.4 |
| System | MacOS |
| Browser | Chrome |

---
Similar issue: #56339

Gdhanush-13:
## What

Fixes #57029 - `Tag` with `variant="solid"` has poor text contrast in dark theme (v6).

## Root Cause

Two issues combine to make solid tags nearly illegible in dark mode:

1. **Missing CSS class**: The `ant-tag-default` class was never applied to solid tags without an explicit `color` prop. The CSS rule `&.ant-tag-default { color: solidTextColor }` existed but never activated because the class wasn't added to the DOM element.
2. **Hardcoded token values**: `prepareComponentToken` used hardcoded `'#fff'` / `'#000'` instead of theme-aware tokens (`colorBgContainer`, `colorTextSolid`, `colorTextLightSolid`).

## Fix

- **`components/tag/index.tsx`**: Add `ant-tag-default` class when `variant` is `'solid'` and no custom/preset color is set. This enables the existing CSS rule that applies `solidTextColor`.
- **`components/tag/style/index.ts`**: Use theme tokens instead of hardcoded values in `prepareComponentToken` for better theme adaptability.
- **`components/tag/__tests__/index.test.tsx`**: Add test verifying the `ant-tag-default` class is applied for solid tags without a color prop in dark mode.

## Changes

| File | Change |
|------|--------|
| `index.tsx` | Add `ant-tag-default` class for colorless solid tags |
| `style/index.ts` | Use `token.colorBgContainer` / `token.colorTextSolid` / `token.colorTextLightSolid` |
| `__tests__/index.test.tsx` | Add dark theme contrast test |

### DIFF
--- a/components/tag/__tests__/index.test.tsx
+++ b/components/tag/__tests__/index.test.tsx
@@ -407,7 +407,7 @@ describe('Tag', () => {
       </StyleProvider>,
     );
 
-    expect(container.querySelector('.ant-tag-default')).toBeTruthy();
+    expect(container.querySelector('.ant-tag')).toHaveClass('ant-tag-default');
   });
 
   it('legacy bordered={false}', () => {
--- a/components/tag/__tests__/index.test.tsx
+++ b/components/tag/__tests__/index.test.tsx
@@ -394,6 +394,22 @@ describe('Tag', () => {
     expect(document.head.innerHTML).toContain('--ant-tag-solid-text-color:#000;');
   });
 
+  it('dark theme solid without color prop should apply default class for contrast', () => {
+    const { container } = render(
+      <StyleProvider cache={createCache()}>
+        <ConfigProvider
+          theme={{
+            algorithm: darkAlgorithm,
+          }}
+        >
+          <Tag variant="solid">Tag</Tag>
+        </ConfigProvider>
+      </StyleProvider>,
+    );
+
+    expect(container.querySelector('.ant-tag-default')).toBeTruthy();
+  });
+
   it('legacy bordered={false}', () => {
     const { container } = render(<Tag bordered={false}>Tag</Tag>);
     expect(container.querySelector('.ant-tag-filled')).toBeTruthy();
--- a/components/tag/__tests__/index.test.tsx
+++ b/components/tag/__tests__/index.test.tsx
@@ -392,7 +392,7 @@ describe('Tag', () => {
 
   it('solid variant with status color should not have default class', () => {
     const { container } = render(
-      <Tag variant="solid" color="default">
+      <Tag variant="solid" color="success">
         Tag
       </Tag>,
     );
--- a/components/tag/__tests__/index.test.tsx
+++ b/components/tag/__tests__/index.test.tsx
@@ -1,6 +1,4 @@
 import React from 'react';
-import { darkAlgorithm } from '@ant-design/compatible';
-import { createCache, StyleProvider } from '@ant-design/cssinjs';
 import { CheckCircleOutlined, CloseCircleOutlined, LinkedinOutlined } from '@ant-design/icons';
 
 import Tag from '..';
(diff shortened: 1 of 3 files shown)


components/tag/__tests__/index.test.tsx
```diff
@@ -394,6 +394,22 @@ describe('Tag', () => {
...
+          theme={{
+            algorithm: darkAlgorithm,
+          }}
+        >
+          <Tag variant="solid">Tag</Tag>
+        </ConfigProvider>
+      </StyleProvider>,
+    );
+
+    expect(container.querySelector('.ant-tag')).toHaveClass('ant-tag-default');
```
afc163: 这个测试用例没什么意义。

Gdhanush-13:
@afc163 Removed the test case and cleaned up unused imports. Thanks for the feedback!

Gdhanush-13:
Addressed the codex review feedback - changed solidTextColor to use hardcoded '#000'/'#fff' values, matching the same pattern used in Button component's prepareComponentToken (components/button/style/token.ts).

The issue was that in dark theme, both colorBgSolid (near-white) and 	oken.colorText (also light) produced no contrast. The Button component solves this identically with hardcoded values.

afc163:
failing checks

    [codecov/project](https://github.com/ant-design/ant-design/pull/57469/checks?check_run_id=70116135714)
    codecov/projectFailing after 1s — 99.99% (target 100.00%)

Gdhanush-13:
@afc163 Added tests covering the new nt-tag-default class conditional for solid variant tags - both with and without a custom color. This should fix the codecov/project check.

Gdhanush-13:
Fixed - the previous test used color="default" which is a status color (isInternalColor=true), so it hit the existing ${prefixCls}-${mergedColor} path instead of the new mergedVariant === 'solid' && !isInternalColor && !mergedColor branch.

Updated to:
- **No color prop** ? exercises the new branch, expects .ant-tag-default`n- **Preset color (lue)** ? no default class
- **Status color (default)** ? no default class via the new branch

This should also help with the codecov/project check.

yoyo837:
Close by #57495

PR #57474, opened 2026-03-28
fix(Upload): show error message inline for touch screen accessibility (#57315)

### LINKED ISSUE #57315: Upload error is not touch screen friendly
### Reproduction link

[https://ant.design/components/upload](https://ant.design/components/upload)

### Steps to reproduce

1. Use touch screen device, like an iPad, to see the upload error of examples 

### What is expected?

Show the error (may be under the file name?)

The error is a tooltip. With touch screen, you can only click it and open file preview/download the file. Not able to hover.

| Environment | Info |
| --- | --- |
| antd | 6.3.2 |
| React | 18.3.0 |
| System | macOS |
| Browser | Chrome 145 |

1164148-7-eng:
### Problem
Upload error messages were displayed using a `Tooltip` component, which is inaccessible on touch devices (iPad, mobile phones) since Tooltips require hover to display.

When a user taps on an errored file item on a touch device, it triggers file preview/download instead of showing the error message.

### Solution
Show error messages as inline text directly in the file list item, replacing the Tooltip approach:

- **Before**: Error message hidden inside Tooltip (requires hover)
- **After**: Error message displayed inline below filename (always visible)

### Changes
- `components/upload/UploadList/ListItem.tsx`: Removed Tooltip wrapper, added inline error message span below filename
- `components/upload/style/list.ts`: Added `.upload-list-item-error-message` CSS styling with proper error color and layout

### Testing
1. Trigger an upload error on a touch device (iPad/iPhone)
2. Observe that error message is now visible without any interaction
3. Tap the filename still works for preview (no regression)

Fixes #57315

### DIFF
--- a/components/upload/UploadList/ListItem.tsx
+++ b/components/upload/UploadList/ListItem.tsx
@@ -7,7 +7,7 @@ import { clsx } from 'clsx';
 
 import { ConfigContext } from '../../config-provider';
 import Progress from '../../progress';
-import Tooltip from '../../tooltip';
+
 import type {
   ItemRender,
   UploadFile,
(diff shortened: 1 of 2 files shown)


afc163 (changes requested):
The default touch-device direction is valid. Please use the repository `isString` helper, add a regression test for the default error UI, rebase, and restore green CI.

PR #57494, opened 2026-03-29
fix:(57029) Tag with variant="solid" has poor text contrast in dark t…

dipeshpatel02:
(tag/style/index.ts:222-225)
ts
const solidTextColor = isBright(new AggregationColor(token.colorBgSolid), '#fff')
  ? '#000'
  : '#fff';
isBright() checks if the color is bright relative to '#fff' (a white background). In dark mode, colorBgSolid is rgba(255,255,255,0.95) — which is bright relative to #fff, so it correctly returns '#000'. But in dark mode the actual page background is dark (~#141414). The luminance of rgba(255,255,255,0.95) composited onto a dark background is still bright, so solidTextColor = '#000' — this part happens to be correct for the default tag.


afc163:
Thanks for the fix. The linked #57029 was resolved by the narrower merged #57495, which supplies the default solid colors for Tag and Button and includes the corresponding tests and documentation. Closing this superseded implementation.

PR #57662, opened 2026-04-12
feat(Dropdown): support left/right placement directions

SAY-5:
Dropdown only supported top and bottom placements, but Popover already supports left/right directions. This adds leftTop, leftBottom, rightTop, rightBottom, left, and right to the Dropdown placement type and the internal placements array.

Bare 'left' and 'right' map to leftTop and rightTop respectively, consistent with how 'top' and 'bottom' map to topLeft and bottomLeft.

Closes #56856

### DIFF
--- a/components/dropdown/dropdown.tsx
+++ b/components/dropdown/dropdown.tsx
@@ -33,11 +33,17 @@ const _Placements = [
   'bottomRight',
   'top',
   'bottom',
+  'leftTop',
+  'leftBottom',
+  'left',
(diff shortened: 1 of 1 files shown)


SAY-5:
Closing this -- the left/right placements need support in @rc-component/dropdown first before Dropdown can expose them. The RcDropdown type only accepts top/bottom variants, so adding these here causes a TS error at the handoff. Will reopen once the upstream component supports them.

PR #57695, opened 2026-04-16
fix(Alert): add focus-visible outline for Alert close button

KittyGiraudel:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

I don’t know if there is a related issue for this one — if there is, I couldn’t find it. 

### 💡 Background and Solution

The problem is that the close button from the `Alert` component hides its outline, but does *not* apply focus styles, so it’s essentially unusable with a keyboard (well, it’s usable, but it’s not easy since you don’t know if you’re focusing it).

The solution is just to apply the focus ring to it.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix the `Alert`’s close button not having focus styles          |
| 🇨🇳 Chinese | 修复 `Alert` 的关闭按钮没有焦点样式的问题          |

### DIFF
--- a/components/alert/style/index.ts
+++ b/components/alert/style/index.ts
@@ -2,7 +2,7 @@ import type { CSSProperties } from 'react';
 import type { CSSObject } from '@ant-design/cssinjs';
 import { unit } from '@ant-design/cssinjs';
 
-import { resetComponent } from '../../style';
+import { genFocusStyle, resetComponent } from '../../style';
 import type { FullToken, GenerateStyle, GetDefaultToken } from '../../theme/internal';
 import { genStyleHooks } from '../../theme/internal';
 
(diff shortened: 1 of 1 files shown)


components/alert/style/index.ts
```diff
@@ -216,8 +216,9 @@ export const genActionStyle: GenerateStyle<AlertToken, CSSObject> = (token) => {
         lineHeight: unit(fontSizeIcon),
         backgroundColor: 'transparent',
         border: 'none',
-        outline: 'none',
         cursor: 'pointer',
+        outline: 'none',
```
meet-student: Image

There should be no need to write an outline

KittyGiraudel: I can remove this, definitely. Makes sense not to author `outline` now that we use `genFocusOutline`. Give me a sec.

KittyGiraudel: Done, thank you, good point. :)

afc163 (approved):

PR #57858, opened 2026-05-05
fix(Tooltip): reposition popup when ancestor layout shifts due to asy…

### LINKED ISSUE #57809: Tooltip popup does not realign when trigger position changes while open
### Reproduction link

[Edit on StackBlitz](https://stackblitz.com/edit/9qdrqgd7?file=src%2Fdemo.tsx)

### Steps to reproduce

1. 打开复现链接：https://stackblitz.com/edit/9qdrqgd7?file=src%2Fdemo.tsx
2. 页面初始渲染一个受控的 Tooltip，设置 `open` 使其保持展示。
3. 等待 2 秒，页面上方会异步渲染一组数据。
4. 数据渲染后，Tooltip 的触发元素会被向下挤压移动。

### What is expected?

当 Tooltip 已经打开时，如果触发元素的位置因为页面内容变化而发生改变，Tooltip 浮层应该重新计算位置，并继续对齐到触发元素。

触发元素已经被新的页面内容向下挤压移动，但 Tooltip 浮层仍停留在原来的位置，没有重新对齐触发元素，导致浮层位置错误。

| Environment | Info |
| --- | --- |
| antd | 6.3.7 |
| React | 19.x |
| System | macOS |
| Browser | Chrome latest |

---
这个问题在 Tooltip 受控打开（`open`）时比较容易复现。实际业务中，页面异步加载数据、列表刷新、表单区域展开/收起等场景都可能导致触发元素位置变化。

nickmopen:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues
Fixes: https://github.com/ant-design/ant-design/issues/57809
> When a controlled Tooltip (`open`) is visible and content is asynchronously inserted above the trigger element, the trigger is pushed down by layout reflow but the popup stays at its original position.
>
> The trigger element's *position* changes without any `scroll` or `window.resize` event firing, so the popup is never re-aligned.

### 💡 Background and Solution

**Problem**

When new DOM content is inserted above the trigger (e.g. an async list renders), the trigger is shifted downward purely by layout reflow. No scroll occurs, no resize fires on the trigger element itself, so the popup is never re-aligned.

**Solution**

Added a `useEffect` in `InternalTooltip` that runs while `tempOpen` is `true`. It walks up the DOM from the trigger's `nativeElement` collecting every scrollable ancestor. A `ResizeObserver` is attached to each ancestor — when any of them grows because new content was inserted into it, `forceAlign()` is called to snap the popup back to the trigger.

The observer is disconnected as soon as `tempOpen` becomes `false` or the component unmounts, so there is no overhead when the tooltip is closed.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Tooltip popup position not updating when trigger element is shifted by async content insertion while `open` is controlled |
| 🇨🇳 Chinese | 🐞 修复 Tooltip 受控打开（`open`）时，触发元素因异步内容插入被向下挤压后，浮层位置未随之更新的问题 |

### DIFF
--- a/components/tooltip/__tests__/reposition.test.tsx
+++ b/components/tooltip/__tests__/reposition.test.tsx
@@ -1,0 +1,240 @@
+import React from 'react';
+import { spyElementPrototype } from '@rc-component/util/lib/test/domHook';
+
+import Tooltip from '..';
+import { act, pureRender, render, waitFakeTimer } from '../../../tests/utils';
+
+// ---------------------------------------------------------------------------
+// Helper: controllable ResizeObserver mock
+// ---------------------------------------------------------------------------
+
+type ROCallback = (entries: ResizeObserverEntry[]) => void;
+
+interface MockROInstance {
+  cb: ROCallback;
+  targets: Set<Element>;
+}
+
+function createResizeObserverMock() {
+  const instances: MockROInstance[] = [];
+
+  class MockResizeObserver {
+    private cb: ROCallback;
+
+    constructor(cb: ROCallback) {
+      this.cb = cb;
+      instances.push({ cb, targets: new Set() });
+    }
+
+    observe(target: Element) {
+      instances.find((i) => i.cb === this.cb)?.targets.add(target);
+    }
+
+    unobserve(target: Element) {
+      instances.find((i) => i.cb === this.cb)?.targets.delete(target);
+    }
+
+    disconnect() {
+      const idx = instances.findIndex((i) => i.cb === this.cb);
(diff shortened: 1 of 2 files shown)


nickmopen:
@thinkasany , Can you review this?

afc163:
Thanks for the contribution. Closing because observing the selected element does not reliably detect its internal layout shifts, so the original Tooltip positioning bug can remain. This needs an observer/trigger tied to the actual geometry change.

PR #40519, opened 2023-02-02
fix(Form): Fix Form.Item not rendering correctly in Modal when it has help

Yuiai01:
[[中文版模板 / Chinese template](https://github.com/ant-design/ant-design/blob/master/.github/PULL_REQUEST_TEMPLATE/pr_cn.md)]

### 🤔 This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / documentation update
- [ ] Demo update
- [ ] Component style update
- [ ] TypeScript definition update
- [ ] Bundle size optimization
- [ ] Performance optimization
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Other (about what?)

### 🔗 Related issue link
close https://github.com/ant-design/ant-design/issues/40500

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix `Form.Item` not rendering correctly in `Modal` when it has `help`           |
| 🇨🇳 Chinese | 修复 `Form.Item` 存在 help 值时在 `Modal` 中没有正确渲染元素         |

### ☑️ Self-Check before Merge

⚠️ Please check all items below before requesting a reviewing. ⚠️

- [x] Doc is updated/provided or not needed
- [x] Demo is updated/provided or not needed
- [x] TypeScript definition is updated/provided or not needed
- [x] Changelog is provided or not needed

PR #57863, opened 2026-05-05
fix(form): avoid forced reflow on Form.Item initial render (#57855)

### LINKED ISSUE #57855: Form.Item triggers forced reflow on initial render via isVisible(offsetParent)
### Reproduction link

[https://ant.design/~demos/form-demo-disabled](https://ant.design/~demos/form-demo-disabled)

### Steps to reproduce

Render a large vertical `Form` with many `Form.Item` fields and run Lighthouse or a Chrome Performance trace.

### What is expected?

`Form.Item` should avoid synchronous layout reads during initial render when there is no validation error or help state that needs margin adjustment.

A possible fix would be to only call `isVisible(itemRef.current)` when `hasError` is true, since `isOnScreen` is only used as a dependency for the margin-bottom effect.

`Form.Item` triggers forced reflows during initial render.

The trace points to:

- `antd/es/form/FormItem/ItemHolder.js`
- `@rc-component/util/es/Dom/isVisible.js`

`ItemHolder` computes `isOnScreen` during render:

    const isOnScreen = !!itemRef.current && isVisible(itemRef.current);

`isVisible()` reads `element.offsetParent`, and this happens for every `Form.Item` during render, even when the item has no errors or help text.

On one real page with about 70 visible fields this caused about 140 `offsetParent` reads and Lighthouse reported Forced reflow.

| Environment | Info |
| --- | --- |
| antd | 6.3.5 |
| React | 19 |
| System | macOs |
| Browser | Chrome |

---
Tested with Next.js 16

ChrisJr404:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [x] ⚡️ Performance optimization
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

close #57855

### 💡 Background and Solution

`Form.Item`'s `ItemHolder` always called `isVisible(itemRef.current)` during render. `isVisible` reads `element.offsetParent` (and falls back to `getBBox`/`getBoundingClientRect`), all of which trigger a synchronous layout. On a vertical `Form` with ~70 `Form.Item` fields the reporter observed ~140 `offsetParent` reads on initial render and Lighthouse flagged it as Forced reflow.

`isOnScreen` is only used as a dependency for the margin-bottom `useLayoutEffect`, and that effect's body is already gated by `hasError`:

```tsx
useLayoutEffect(() => {
  if (hasError && itemRef.current) {
    const itemStyle = getComputedStyle(itemRef.current);
    setMarginBottom(Number.parseInt(itemStyle.marginBottom, 10));
  }
}, [hasError, isOnScreen]);
```

So when there is no help/error/warning, calling `isVisible` during render is wasted work.

This PR gates the visibility read behind `hasError`:

```tsx
const isOnScreen = hasError && !!itemRef.current && isVisible(itemRef.current);
```

Behaviour preserved:

- When `hasError` flips from `false` to `true`, the effect still re-runs because `hasError` itself changed, so the margin is read on the first render that needs it.
- When `hasError` stays `true` and the item later becomes visible (e.g. the original Modal case from #40519), the next render still reads `offsetParent`, the dep flips, and the effect re-runs to recompute `marginBottom`.
- The dep array still contains `isOnScreen` so the visibility-flip path is unchanged.

Two regression tests are added in `components/form/__tests__/index.test.tsx`:

1. Renders a 5-item `Form` with no errors and asserts `HTMLElement.prototype.offsetParent` is not accessed during the initial render.
2. Renders a single `Form.Item` with `help` set and asserts `offsetParent` is still accessed, pinning the Modal/help-text path.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Avoid forced reflow on `Form.Item` initial render by skipping the `offsetParent` visibility check when there is no help, error, or warning. |
| 🇨🇳 Chinese | 在没有 help / error / warning 时跳过 `Form.Item` 的 `offsetParent` 可见性检查，避免初始渲染时的强制回流。 |

### DIFF
--- a/components/form/FormItem/ItemHolder.tsx
+++ b/components/form/FormItem/ItemHolder.tsx
@@ -65,7 +65,11 @@ export default function ItemHolder(props: ItemHolderProps) {
   const debounceWarnings = useDebounce(warnings);
   const hasHelp = isNonNullable(help);
   const hasError = !!(hasHelp || errors.length || warnings.length);
-  const isOnScreen = !!itemRef.current && isVisible(itemRef.current);
+  // `offsetParent` access in `isVisible` triggers a synchronous layout. Skip
+  // it when there is no help/error/warning to read computed margin from, so
+  // the initial render of large forms does not incur a forced reflow per item.
+  // See: https://github.com/ant-design/ant-design/issues/57855
+  const isOnScreen = hasError && !!itemRef.current && isVisible(itemRef.current);
   const [marginBottom, setMarginBottom] = React.useState<number | null>(null);
 
   useLayoutEffect(() => {
--- a/components/form/__tests__/index.test.tsx
+++ b/components/form/__tests__/index.test.tsx
@@ -2044,6 +2044,73 @@ describe('Form', () => {
     });
   });
 
+  // https://github.com/ant-design/ant-design/issues/57855
+  it('should not read offsetParent on initial render when there is no error or help', () => {
+    const originalDescriptor = Object.getOwnPropertyDescriptor(
+      HTMLElement.prototype,
+      'offsetParent',
+    );
+    const offsetParentGetter = jest.fn(() => document.body);
+    Object.defineProperty(HTMLElement.prototype, 'offsetParent', {
+      configurable: true,
+      get: offsetParentGetter,
+    });
+
+    try {
+      render(
+        <Form>
+          {Array.from({ length: 5 }).map((_, index) => (
+            <Form.Item key={index} name={`field-${index}`} label={`Field ${index}`}>
+              <Input />
+            </Form.Item>
+          ))}
+        </Form>,
+      );
+
+      expect(offsetParentGetter).not.toHaveBeenCalled();
+    } finally {
+      if (originalDescriptor) {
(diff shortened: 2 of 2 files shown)


afc163:
The focused performance change is semantically reasonable, but the linked issue was closed as not planned and the branch is far behind. Please reconfirm that maintainers still want the change, then rebase and refresh tests.

ChrisJr404:
Fair points. The change is behavior-preserving: isOnScreen only feeds the useLayoutEffect dep array, and that effect's body plus the marginBottom setState are already gated on hasError, so skipping the offsetParent read when there's no error/help/warning changes nothing except dropping the per-item reflow on initial render. I've rebased onto master and the visual diffs look like stale-baseline noise — one of them is grid-flex, which this doesn't touch. Before I push the rebase, do you still want this given #57855 was closed as not planned? Happy to reopen the perf discussion there if that's cleaner.

PR #57883, opened 2026-05-07
fix(Menu): navigate anchor links when Enter is pressed on menu item

### LINKED ISSUE #57766: Menu not setting focus to link in first menu item containig link when navigating with keyboard
### Reproduction link

[Edit on StackBlitz](https://stackblitz.com/edit/vuy6tohz?file=src%2Fdemo.tsx)

### Steps to reproduce

1. Create a DropdownButton component with menu where all items are links (check code block below).
2. Add autoFocus prop to DrodownButton (this ctep can be skipped, otherwise you just need to focus to menu manually with Tab key).
3. Focus with keyboard on DropdownButton, press enter.
4. Menu opens, focus on menu (if not using autoFocus)
5. Fist menu item is focused. Press Enter.

```
import React from 'react';

import type { MenuProps } from 'antd';
import { Button, Dropdown, Space } from 'antd';

const items: MenuProps['items'] = [
  {
    key: '1',
    label: (
      <a
        target="_blank"
        rel="noopener noreferrer"
        href="https://www.antgroup.com"
      >
        1st menu item
      </a>
    ),
  },
  {
    key: '2',
    label: (
      <a
        target="_blank"
        rel="noopener noreferrer"
        href="https://www.aliyun.com"
      >
        2nd menu item
      </a>
    ),
  },
  {
    key: '3',
    label: (
      <a
        target="_blank"
        rel="noopener noreferrer"
        href="https://www.luohanacademy.com"
      >
        3rd menu item
      </a>
    ),
  },
];

const App: React.FC = () => (
  <Space vertical>
    <Space wrap>
      <Dropdown trigger={['click']} menu={{ items }}>
        <Button>Menu with links</Button>
      </Dropdown>
    </Space>
  </Space>
);

export default App;
```

### What is expected?

User should be navigated to url from html link.

Menu closes, navigation not happening.

| Environment | Info |
| --- | --- |
| antd | 6.3.6 |
| React | React 19 |
| System | iOS 26.3.1 (a) |
| Browser | Chrome 146.0.7680.80 |

---
Currently focus is set on menu item container, and if menu item contains link, it is not possible to follow by link url with keyboard.

nickmopen:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

Fixes: https://github.com/ant-design/ant-design/issues/57766

### 💡 Background and Solution

**Problem:**

When a Dropdown menu opens, DOM focus lands on the `<li>` container element (`tabIndex="-1"`), not on the inner `<a>` element. When the user presses Enter, `rc-menu`'s `onKeyDown` handler fires `onItemClick` synchronously, which causes the Dropdown to call `setOpen(false)`. React 18/19 flushes this state update synchronously in the same event batch — before the browser can dispatch its native Enter→click on the focused anchor. By the time the browser tries to navigate, the popup is already unmounted and the anchor is detached from the DOM, so navigation never occurs.

**Solution:**

An `onKeyDownCapture` handler is added to `RcMenu` in menu.tsx. Because it runs in the **capture phase**, it fires before `rc-menu`'s own bubbling `onKeyDown`. When Enter is pressed and the active menu item contains an `<a href>`, it calls `link.click()` programmatically — while the popup is still mounted — to trigger navigation. `e.preventDefault()` is then called to suppress the browser's own deferred Enter→click and avoid double-navigation.

No new props or API changes are introduced.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Dropdown menu item with `<a>` link not navigating when activated via keyboard Enter |
| 🇨🇳 Chinese | 🐞 修复 Dropdown 菜单项包含 `<a>` 链接时，通过键盘 Enter 键无法跳转链接的问题 |


nickmopen:
@afc163 , can you review this PR when you have a moment? Thanks.

nickmopen:
@li-jia-nan , @meet-student , can you review this PR?

afc163:
The keyboard-navigation fix still matches the accessibility issue. Please rebase onto the current target branch and rerun the focused interaction tests.

nrps9909 (changes requested):
I reviewed exact head `709b2772d829dbccb05fc9655824a1d4225eb30c` and reproduced two merge-blocking regressions in the capture-phase handler.

1. Disabled items are bypassed. The handler checks only Enter plus a descendant `a[href]`, so an `items={[{ disabled: true, label: <a href="/target">…</a> }]}` entry still calls `anchor.click()`. A focused current-head regression expected zero calls and received one. This turns a semantically disabled menu item into an active navigation path.

2. The public consumer `onKeyDownCapture` callback is silently discarded. It remains in `passedProps`, but the explicit `onKeyDownCapture={onKeyDownCapture}` after the spread overwrites it. A focused current-head probe sent Escape to a plain item and the consumer spy received zero calls instead of one.

The five submitted anchor-navigation tests pass, but neither boundary is covered. Please guard disabled menu items and compose the internal behavior with the consumer capture handler (preserving the normal event contract), then add both regressions. The branch is also currently DIRTY against master and the complete focused Menu file shows five stale rc-menu ID snapshots with today's dependencies, so it needs a current-base refresh before final validation.

AI assistance disclosure: Codex was used to inspect the exact diff and run the two temporary focused regressions; I verified both failures and removed the probes afterward.

PR #57930, opened 2026-05-12
fix: add package.json exports field for Yarn PnP / strict module resolution

### LINKED ISSUE #56858: Missing exports field in package.json causes Yarn PnP resolution failures for subpath imports
### Reproduction link

[https://github.com/OrlovAlexei/antd-exports-issue](https://github.com/OrlovAlexei/antd-exports-issue)

### 1. Clone the repository

Clone the repository and follow the instructions in the README

```bash
git clone https://github.com/OrlovAlexei/antd-exports-issue.git
cd antd-exports-issue
```

### 2. Install dependencies

```bash
yarn install
```

### 3. Build the library and create a tarball

```bash
yarn build:lib
cd packages/lib && yarn pack && cd ../..
```

### 4. Install dependencies again (to pick up the tarball)

```bash
yarn install
```

### 5. Run the tests

```bash
cd packages/app && yarn vitest run
```

### What is expected?

Yarn PnP should be able to resolve subpath imports like antd/es/input/Search when antd has proper exports field configuration.

Yarn PnP fails with "Qualified path resolution failed" error because antd doesn't declare these paths in exports.

| Environment | Info |
| --- | --- |
| antd | 6.2.3 |
| React | 18.2.0 |
| System | Lunux |
| Browser | N/A (Node.js/Yarn PnP) |

---
When consuming libraries that re-export antd components from subpaths (e.g., antd/es/input/Search), Yarn PnP fails to resolve these paths because it strictly validates package exports through the exports field. The error occurs when a library imports from antd subpaths and is consumed as a tarball/npm package (not workspace).

 Suggested Solution

Add exports field to package.json:

```json
{
  "exports": {
    ".": {
      "import": "./es/index.js",
      "require": "./lib/index.js",
      "types": "./es/index.d.ts"
    },
    "./es/_": "./es/_",
    "./lib/_": "./lib/_",
    "./locale/_": "./locale/_",
    "./dist/reset.css": "./dist/reset.css",
    "./package.json": "./package.json"
  }
}
```

## Additional Context

• Issue only reproduces when library is installed from npm/tarball, not with workspace:\*
• This affects real-world usage where design systems are published to npm
• Related to modern ESM/Node.js module resolution standards
• Similar issue: https://github.com/ant-design/ant-design/issues/46413

TheSeydiCharyyev:

### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

close #56858

### 💡 Background and Solution

**Background**

When `antd` is consumed via Yarn PnP (Plug'n'Play), subpath imports like `antd/es/input/Search` fail to resolve with:

> Qualified path resolution failed

This is because Yarn PnP strictly validates package exports through the `exports` field in `package.json`, and `antd` currently does not declare one. As @MadCcc noted in the issue:

> Sounds good, but need to care not break current import path.

A minimal reproduction is provided by the reporter at https://github.com/OrlovAlexei/antd-exports-issue.

**Solution**

Added an `exports` field to `package.json` that:

1. Declares the main entry (`.`) with explicit `types`, `import` and `require` conditions matching the existing `typings`, `module` and `main` fields.
2. Maps `./es/*`, `./lib/*`, `./dist/*`, `./locale/*` to themselves using wildcard patterns. This preserves every existing subpath import without enumerating individual files.
3. Exposes `./package.json` so consumers and tooling can read it.

This is the minimal change needed to unblock Yarn PnP while keeping every current import pattern working:

| Import | Resolved via |
| ------ | ------------ |
| `import { Button } from 'antd'` | `.` |
| `import Button from 'antd/es/button'` | `./es/*` |
| `import Button from 'antd/lib/button'` | `./lib/*` |
| `import 'antd/dist/reset.css'` | `./dist/*` |
| `import zhCN from 'antd/locale/zh_CN'` | `./locale/*` |
| `import 'antd/es/button/style'` (deep subpath) | `./es/*` |
| `import pkg from 'antd/package.json'` | `./package.json` |

**Verification**

- `npm run lint` — passes (only pre-existing warnings unrelated to this change)
- `npm test` — 558 suites, 7041 tests, 3957 snapshots all passed (2 suites skipped as expected)
- No source files modified — only `package.json`

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 📦 Add `exports` field to `package.json` to fix Yarn PnP subpath resolution failures while preserving all existing import paths. |
| 🇨🇳 Chinese | 📦 为 `package.json` 添加 `exports` 字段，修复 Yarn PnP 下子路径解析失败的问题，同时保留所有现有导入路径。 |

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,18 @@
   "sideEffects": [
     "*.css"
   ],
+  "exports": {
+    ".": {
+      "types": "./es/index.d.ts",
+      "import": "./es/index.js",
+      "require": "./lib/index.js"
+    },
+    "./es/*": "./es/*",
+    "./lib/*": "./lib/*",
+    "./dist/*": "./dist/*",
+    "./locale/*": "./locale/*",
+    "./package.json": "./package.json"
+  },
   "main": "lib/index.js",
   "module": "es/index.js",
   "unpkg": "dist/antd.min.js",


afc163 (approved):

afc163:
## This PR introduces a TypeScript regression

The `exports` field added in this PR breaks TypeScript type resolution when `moduleResolution` is set to `"bundler"` (the default for Vite-based projects) or `"nodenext"`.

### Root cause

TypeScript's exports-based module resolution does **NOT** perform directory/index lookups. When resolving `antd/es/theme/interface`, it only tries direct file paths like `./es/theme/interface.d.ts` — it will never check `./es/theme/interface/index.d.ts`.

However, antd's `es/` output uses a **mixed structure**:

- **File-based**: `es/config-provider/context.d.ts` (standalone file)
- **Directory-based**: `es/theme/interface/index.d.ts` (directory with index)

No single wildcard pattern in `exports` can correctly resolve types for both cases:

| Pattern | File-based (`config-provider/context`) | Directory-based (`theme/interface`) |
|---|---|---|
| `"./es/*": { "types": "./es/*.d.ts" }` | ✅ | ❌ |
| `"./es/*": { "types": "./es/*/index.d.ts" }` | ❌ | ✅ |
| `"./es/*": "./es/*"` (simple string) | ❌ | ❌ |

### Impact

This causes ~200+ `TS2339` errors in downstream projects that use `antd-style`, which deep-imports `AliasToken` from `antd/es/theme/interface`. Since that import fails to resolve, `FullToken` becomes an empty interface, and all token properties (`colorText`, `fontSize`, `controlHeight`, `colorTextSecondary`, `screenXS`, etc.) are reported as "does not exist on type 'FullToken'".

### How to reproduce

1. Create a Vite project with `moduleResolution: "bundler"` in tsconfig.json
2. Install `antd@6.4.0` and `antd-style`
3. Use `createStyles` from antd-style with any token property:

```typescript
import { createStyles } from 'antd-style';

const useStyles = createStyles(({ token }) => ({
  text: { color: token.colorText }, // TS2339: Property 'colorText' does not exist on type 'FullToken'
}));
```

4. Run `npx tsc --noEmit` — you will see the errors

### Follow-up fix

I've opened #57968 to remove the `exports` field for now, restoring the traditional Node.js module resolution path which correctly handles both file and directory lookups.

If we want to re-introduce `exports` for Yarn PnP support in the future, we need one of these approaches:

1. **Generate re-export stub files** for every directory-based module (e.g. create `es/theme/interface.d.ts` that re-exports from `./theme/interface/index.js`), so that `./es/*.d.ts` pattern works for all paths. Currently there are 169 such directories.

2. **Enumerate all subpaths explicitly** in `exports` instead of using wildcards, so each path can point to the correct type declaration file. This is maintainable for `locale/` (limited set) but impractical for `es/` and `lib/` (hundreds of subpaths that change with each release).

3. **Restructure the build output** so all modules use file-based structure (no `index.d.ts` in subdirectories), making `"./es/*.d.ts"` universally correct.

afc163:
Real-world CI failure caused by this PR: https://github.com/ant-design/ant-design-pro/actions/runs/25850042548/job/75954108350?pr=11783

TheSeydiCharyyev:
Sorry for the regression. My pre-merge testing covered Yarn PnP and `moduleResolution: node16`, but didn't cover `bundler`/`nodenext` against the mixed file/dir output in `es/`/`lib/`. The wildcard `exports` pattern can't satisfy both layouts simultaneously — your analysis is correct.

Revert in #57968 is the right call. If proper `exports` support is wanted later, I can prototype approach 1 (generate stub re-export files for the 169 directory-based modules at build time) in a separate PR after this lands.

afc163:
Generating re-export stubs at build time is technically feasible, but the tradeoff doesn't seem worth it:

1. **169 stubs pollute the package for all consumers** — adding hundreds of wrapper files to `es/` and `lib/` for a Yarn PnP minority, increasing package size and directory noise for everyone.

2. **Fragile to maintain** — new components may introduce directory-based modules. If a stub is missing for a new module, types silently break without CI catching it (the existing module works fine, the new one just loses types). The build script needs constant sync.

3. **Yarn PnP users already have a simple alternative** — one line in `.yarnrc.yml` with `packageExtensions` resolves the issue without any antd changes.

If we want proper `exports` support in the future, the right approach is to **restructure the build output to eliminate directory-based modules entirely** (make all modules flat files, no `index.d.ts` in subdirectories), so that `"./es/*": { "types": "./es/*.d.ts" }` works universally. That's a one-time build refactor with no ongoing maintenance, and it benefits more than just `exports` — it also simplifies the `es/` structure for all consumers.

PR #57932, opened 2026-05-12
fix: keep Input onChange target attached for custom inputs

### LINKED ISSUE #46999: Antd Input: To use with react-number-format customInput feature.

landeqiming666:
## Summary

Fix `Input` compatibility with wrappers like `react-number-format` when they depend on `document.contains(e.target)` or element identity checks.

This keeps the fix intentionally narrow:
- normalize `event.target` / `event.currentTarget` only for non-`click` `onChange` paths
- preserve `allowClear` click semantics, which still need the synthesized empty value
- cover the regression with an observable controlled-input test instead of asserting only SyntheticEvent internals

## Why this approach

Issue #46999 is caused by `@rc-component/input` forwarding a detached cloned node in some change paths. Third-party `customInput` wrappers can treat that as an invalid target and skip their normal controlled update flow.

There are already open PRs around the same root cause. This patch tries to stay conservative by avoiding the `click` clear path entirely, since that path intentionally relies on a synthesized target value.

## Changes

- In `components/input/Input.tsx`, normalize `target/currentTarget` back to the mounted input element only when:
  - `onChange` exists
  - the event is not a `click`
  - the forwarded target is not the live mounted input
- In `components/input/__tests__/index.test.tsx`, add a regression test that updates a controlled input only when `document.contains(e.target as Node)` is true, and assert the visible value still updates.

## Verification

Ran:

```bash
npm test -- components/input/__tests__/index.test.tsx --runInBand
npx eslint components/input/Input.tsx components/input/__tests__/index.test.tsx
```

The test file passes in full. ESLint reports one pre-existing warning in `components/input/Input.tsx` for an unused `eslint-disable` directive unrelated to this patch.

Closes #46999

### DIFF
--- a/components/input/Input.tsx
+++ b/components/input/Input.tsx
@@ -239,8 +239,8 @@ const Input = forwardRef<InputRef, InputProps>((props, ref) => {
     ) {
       onChange(
         Object.create(e, {
-          target: { value: liveInput, enumerable: true, configurable: true },
-          currentTarget: { value: liveInput, enumerable: true, configurable: true },
+          target: { value: liveInput, enumerable: true, configurable: true, writable: true },
+          currentTarget: { value: liveInput, enumerable: true, configurable: true, writable: true },
         }) as React.ChangeEvent<HTMLInputElement>,
       );
       return;
--- a/components/input/Input.tsx
+++ b/components/input/Input.tsx
@@ -226,7 +226,22 @@ const Input = forwardRef<InputRef, InputProps>((props, ref) => {
 
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     removePasswordTimeout();
-    onChange?.(e);
+    if (!onChange) {
+      return;
+    }
+
+    const liveInput = inputRef.current?.input;
+    if (liveInput && e.type !== 'click' && e.target !== liveInput) {
+      onChange(
+        Object.create(e, {
+          target: { value: liveInput, enumerable: true, configurable: true },
+          currentTarget: { value: liveInput, enumerable: true, configurable: true },
+        }) as React.ChangeEvent<HTMLInputElement>,
+      );
+      return;
+    }
+
+    onChange(e);
   };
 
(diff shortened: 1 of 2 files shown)


landeqiming666:
Extra reviewer context: this patch intentionally avoids the `click` clear path and only normalizes `target/currentTarget` for non-`click` change flows. The goal is to fix the detached-target compatibility issue from #46999 while preserving existing `allowClear` empty-value semantics.

landeqiming666:
Updated this PR to tighten both the implementation and coverage:

- only normalize non-click targets when the forwarded `event.target.value` still matches the mounted input value
- keep `count.exceedFormatter` / trimmed-value semantics intact
- replace the previous behavioral-only regression with a direct mounted-target normalization assertion
- add a separate regression that verifies a controlled input still receives the trimmed `exceedFormatter` value

Re-ran `npm test -- components/input/__tests__/index.test.tsx --runInBand` after the update.

landeqiming666:
Follow-up update: I strengthened the mounted-target compatibility coverage and accepted the low-risk descriptor compatibility suggestion. The normalized event now keeps writable target/currentTarget descriptors, and the regression test verifies a controlled Input only updates when document.contains(e.target) succeeds. The separate exceedFormatter test still protects the trimmed-value path so formatter semantics stay intact.

landeqiming666:
Thanks for the review notes. I pushed a small follow-up that strengthens the mounted-target regression coverage through rc-input's cloned non-click change event path, so the normalization branch is exercised with a forwarded target while still preserving the existing count.exceedFormatter trimmed-value behavior.

afc163:

PR #57936, opened 2026-05-12
fix(Input): normalize detached event target for customInput compatibility

### LINKED ISSUE #46999: Antd Input: To use with react-number-format customInput feature.

islamicfinanceaaoifi-ops:
Closes #46999

### Problem

Since v5.13.3, `@rc-component/input`'s `resolveOnChange` clones the DOM node when synthesizing change events (e.g., `allowClear` clicks, IME composition). The cloned node is **detached** from the document tree.

Third-party libraries like [`react-number-format`](https://github.com/s-yadav/react-number-format) verify the event target via `document.contains(e.target)`. This check fails on detached clones, silently breaking controlled input behavior — values don't update, cursor jumps, and formatting stops working.

### Solution

In `Input.tsx`'s `handleChange`, detect when the event target is a detached clone using the DOM-standard `isConnected` property. When detected:

1. Sync the cloned target's value to the live input element
2. Create a normalized event via `Object.create()` that redirects `target` and `currentTarget` to the mounted input
3. Forward the normalized event to the consumer's `onChange`

Normal typing events (where `e.target` is already the live input) pass through unchanged — zero overhead in the common path.

### Why this approach

| Concern | This PR |
|---------|---------|
| Detection | `isConnected` (DOM standard, catches all detached nodes) |
| Descriptors | `enumerable + writable + configurable` (strict-mode safe) |
| Scope | Handles both allowClear and IME composition scenarios |
| Value sync | Syncs cloned value to live input before forwarding |
| Performance | Zero-cost for normal typing (early return) |

### Test plan

- [x] `allowClear` fires `onChange` with connected target
- [x] Cleared value preserved on normalized target  
- [x] Normal typing `onChange` unaffected
- [x] `document.contains(event.target)` returns `true`

All 48 existing input tests pass.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

- [#46999: Antd Input: To use with react-number-format customInput feature.](https://oss.issuehunt.io/repos/34526884/issues/46999)
---
</details>

### DIFF
--- a/components/input/Input.tsx
+++ b/components/input/Input.tsx
@@ -226,7 +226,23 @@ const Input = forwardRef<InputRef, InputProps>((props, ref) => {
 
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     removePasswordTimeout();
-    onChange?.(e);
+
+    if (!onChange) {
+      return;
+    }
+
+    const liveInput = inputRef.current?.input;
+    if (liveInput && e.target !== liveInput && !e.target.isConnected) {
+      liveInput.value = e.target.value;
+      const normalized = Object.create(e, {
+        target: { value: liveInput, enumerable: true, writable: true, configurable: true },
+        currentTarget: { value: liveInput, enumerable: true, writable: true, configurable: true },
+      });
+      onChange(normalized);
+      return;
+    }
+
+    onChange(e);
   };
 
   const suffixNode = (hasFeedback || suffix) && (
--- a/components/input/Input.tsx
+++ b/components/input/Input.tsx
@@ -233,6 +233,10 @@ const Input = forwardRef<InputRef, InputProps>((props, ref) => {
 
     const liveInput = inputRef.current?.input;
     if (liveInput && e.target !== liveInput && !e.target.isConnected) {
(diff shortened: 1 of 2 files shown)


islamicfinanceaaoifi-ops:
Thanks for the review feedback. Addressed both nits in ea2f2dd:

- Added an explanatory comment above the `liveInput.value = e.target.value` line clarifying why the direct DOM sync is needed before `onChange` (so third-party libraries like react-number-format see a connected target) and that controlled-component flow resumes via the parent re-render.
- Added the suggested `should normalize event when target is detached from DOM` test asserting `event.target === input`, `event.target.isConnected === true`, and `input.value === ''` after `allowClear`.

CI is re-running.

afc163:

PR #57937, opened 2026-05-12
feat: support semantic styles for Layout Sider

### LINKED ISSUE #13527: Make it possible to directly turn a Slider into a flexbox

Currently - in order to use flexbox to style the elements inside a `Sider`, you have to do this:

          <Sider style={{ background: "white" }}>
            <div style={{ display: "flex", flexDirection: "column", justifyContent: "space-evenly", height: "100%" }}>
              <SourcePicker/>
              <TimeSelector setTimeRange={this.setTimeRange} />
            </div>
          </Sider>

This is bad because it requires adding an unecessary `div` just to use flexbox. The reason this happens currently is because a `Sider` is composed of 2 nested divs, so adding `display: flex` to the style prop of the `Sider` only affects the outer div, thus making flexbox unusable. 

This is a problem because it forces developers to add unnecessary divs to their site. This

A] Might make the webpage more inaccessible since excessive use of divs, etc. harms screen readers 

B] bad for performance / webpage size. Also causes inelegant/bloated code 

Add a `bodyStyle` prop, or something similar, that allows developers to style the inner div of the sider, thus letting them use flexbox. Or, just remove the inner `div` inside the sider.

landeqiming666:
### 🤔 This is a ...

- [ ] New feature
- [x] Improvement
- [ ] Bug fix
- [ ] Demo update
- [ ] Component style update
- [ ] Enhancement feature
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [ ] Test Case
- [ ] Branch merge
- [ ] Other

### 🔗 Related issue link

Fixes #13527

### 💡 Background and solution

`Layout.Sider` renders its content inside an internal `${prefixCls}-children` wrapper. As described in #13527, applying `style={{ display: 'flex' }}` to `Sider` only styles the outer `aside`, so users still need an extra wrapper element when they want to lay out the Sider content with flexbox.

This adds the existing semantic DOM customization pattern to `Layout.Sider`:

- `classNames.children` for the internal content wrapper class
- `styles.children` for the internal content wrapper style

The outer `className` and `style` behavior is unchanged, including width/collapse styles and trigger behavior.

### 📝 Changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Support `classNames.children` and `styles.children` on `Layout.Sider` to customize its content wrapper. |
| 🇨🇳 Chinese | 为 `Layout.Sider` 支持 `classNames.children` 与 `styles.children` 以自定义内容包裹容器。 |

### ✅ Test Plan

- `npm test -- components/layout/__tests__/index.test.tsx --runInBand`
- `npx eslint components/layout/Sider.tsx components/layout/__tests__/index.test.tsx`
- `git diff --check`

### DIFF
--- a/components/layout/Sider.tsx
+++ b/components/layout/Sider.tsx
@@ -6,6 +6,8 @@ import RightOutlined from '@ant-design/icons/RightOutlined';
 import { omit } from '@rc-component/util';
 import { clsx } from 'clsx';
 
+import { useMergeSemantic } from '../_util/hooks';
+import type { SemanticClassNamesType, SemanticStylesType } from '../_util/hooks';
 import { isFunction } from '../_util/is';
 import type { Breakpoint } from '../_util/responsiveObserver';
 import { ConfigContext } from '../config-provider';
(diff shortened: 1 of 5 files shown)


landeqiming666:
I removed the temporary scripts snapshot change from the branch. The current diff only touches Layout component, tests, and docs. To avoid changing site snapshot counts, the Semantic DOM docs now use a list instead of adding another markdown table.\n\nLocal verification passed:\n\n- npm test -- components/layout/__tests__/index.test.tsx --runInBand\n- npm run lint:md -- components/layout/index.en-US.md components/layout/index.zh-CN.md\n- npm run site && npm run test:site

PR #57974, opened 2026-05-14
fix(Select): ensure consistent open-state visibility for rich labelRender

### LINKED ISSUE #57951: `labelRender` non-text content (e.g. `<img>`) remains visible when searching in Select/TreeSelect
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/sandbox/zen-brook-h3q2ys?file=%2Fsrc%2FApp.js)

### Steps to reproduce

Create a `Select` with `showSearch` and a `labelRender` that returns a fragment containing an `` followed by a text label (e.g. a flag image   country name)
Select any option so the label is rendered
Click the select to open it — the label should dim
Start typing to search

### What is expected?

Both the `` and the text in `labelRender` should be hidden (or at minimum dimmed) while the user is searching, matching the visual behavior of plain-text labels.

`color: transparent` is applied to `.ant-select-content` on open and `.ant-select-content-has-search-value` when typing (introduced by #56946). This hides text nodes, but `color` has no effect on replaced content such as `` elements. As a result, the flag image stays fully visible while the country name disappears, leaving a broken/inconsistent label during search.

| Environment | Info |
| --- | --- |
| antd | 6.3.7 |
| React | 18.x |
| System | Windows 11 Home (WSL2) |
| Browser | Chrome 148.0.7778.168 (64-bit) |

---
This regressed in 6.3.1 via [#56946](https://github.com/ant-design/ant-design/pull/56946) ("Fix Select & TreeSelect selected value visible when searching"). The fix used color: transparent which only affects text — , icons, or any non-text element placed inside labelRender is unaffected.

Suggested fix — replace color: transparent with opacity in two stages inside the component styles:
```css
[`&${componentCls}-open ${componentCls}-content`]: {
  color: token.colorTextPlaceholder,
  opacity: 0.5,  // dropdown opened: value fades

  '&-has-search-value': {
    color: 'transparent',
    opacity: 0,  // user is typing: value fully hidden
  },
}
```

`opacity: 0` affects the entire subtree regardless of element type, making it a universal solution for any `labelRender` content.

Workaround (CSS override until fixed):
```css
.ant-select-single:not(.ant-select-customize).ant-select-open {
  .ant-select-content > :first-child { opacity: 0.5; }
  .ant-select-content-has-search-value > :first-child { opacity: 0; }
}
```

kfylaktopoulos:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

close #57951

### 💡 Background and Solution

- In Select single mode, open-state content styling relied on element selectors, so plain text selected values could bypass the expected opacity behavior.\

Solution:
- Refined the open-state styling via pure CSS, exactly as requested by the maintainers. The opacity change is now safely scoped to the `open/search` states, ensuring that rich `labelRender` content (like images) transitions smoothly without adding any extra DOM elements (wrappers). The search input text and caret remain fully visible and unaffected.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     🐞 Fix Select single-mode value rendering by consistently wrapping displayed value content, ensuring open-state styling works for both custom labelRender and default label output      |
| 🇨🇳 Chinese |     -      |


QDyanbing:
Thanks for your contribution!

The related Select issue has already been fixed and released in `antd@6.4.1`. Please upgrade to `6.4.1` or later and verify again. If the problem still reproduces after upgrading, feel free to provide a new reproduction and we can take another look.

kfylaktopoulos:
> Thanks for your contribution!
> 
> The related Select issue has already been fixed and released in `antd@6.4.1`. Please upgrade to `6.4.1` or later and verify again. If the problem still reproduces after upgrading, feel free to provide a new reproduction and we can take another look.

Thanks so much for the quick response and for the fix in [antd@6.4.0](https://github.com/ant-design/ant-design/releases/tag/6.4.0) — I really appreciate both your time and the fast turnaround.

I checked your fix, and this PR is intended to address a different edge case:

- It focuses on single-select open/search visual behavior when label content structure differs (default label vs custom [labelRender](vscode-file://vscode-app/c:/Users/kfyla/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) output).

- It keeps default single-select labels as primitive values (to preserve built-in title fallback behavior), and only applies wrapper-specific handling when custom [labelRender](vscode-file://vscode-app/c:/Users/kfyla/AppData/Local/Programs/Microsoft%20VS%20Code/0958016b2a/resources/app/out/vs/code/electron-browser/workbench/workbench.html) is used.
So this is not meant to reopen the original issue, but to cover a related rendering/styling case that remains separate from the 6.4.0 fix.

Thanks again for reviewing — if you prefer, I can further narrow the scope or align the implementation style with your suggested approach.

QDyanbing:
Could you provide a `6.4.1` demo?

kfylaktopoulos:
> Could you provide a `6.4.1` demo?

The `img` should take the opacity as well.
Recording 2026-05-14 160440

kfylaktopoulos:
> Could you provide a `6.4.1` demo?

https://codesandbox.io/p/sandbox/pensive-stitch-dk3xkc?file=%2Findex.tsx

QDyanbing:
> > Could you provide a `6.4.1` demo?
> 
> The `img` should take the opacity as well. Recording 2026-05-14 160440 [ Recording 2026-05-14 160440 ](https://private-user-images.githubusercontent.com/55513145/592509138-25a5512f-0313-4453-a2a3-2d8f5ce01dd8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzg3NjQ4NTIsIm5iZiI6MTc3ODc2NDU1MiwicGF0aCI6Ii81NTUxMzE0NS81OTI1MDkxMzgtMjVhNTUxMmYtMDMxMy00NDUzLWEyYTMtMmQ4ZjVjZTAxZGQ4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTE0VDEzMTU1MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTVlNjYyNWY0NjVhMGQ2MzYyZTZmYzBhZTQyNzEwZDhhN2QwOGM1ZmU3MTJhZjIwNjJmNWM4ZmRhZDEyMWRlMzQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.0oyAoJWH3nTgo2K3mS1wbyNA8eCpOBG3WhxLIFpJcRc) [ ](https://private-user-images.githubusercontent.com/55513145/592509138-25a5512f-0313-4453-a2a3-2d8f5ce01dd8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzg3NjQ4NTIsIm5iZiI6MTc3ODc2NDU1MiwicGF0aCI6Ii81NTUxMzE0NS81OTI1MDkxMzgtMjVhNTUxMmYtMDMxMy00NDUzLWEyYTMtMmQ4ZjVjZTAxZGQ4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTE0VDEzMTU1MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTVlNjYyNWY0NjVhMGQ2MzYyZTZmYzBhZTQyNzEwZDhhN2QwOGM1ZmU3MTJhZjIwNjJmNWM4ZmRhZDEyMWRlMzQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.0oyAoJWH3nTgo2K3mS1wbyNA8eCpOBG3WhxLIFpJcRc)

I don't think we need to handle the opacity here, but let's wait for a team member to confirm.

If we do need to adjust it, I think it should be handled with opacity styles directly instead of adding an extra DOM wrapper.

kfylaktopoulos:
> > > Could you provide a `6.4.1` demo?
> > 
> > 
> > The `img` should take the opacity as well. Recording 2026-05-14 160440
> >       
> >         [
> >           
> >         
> >             Recording 2026-05-14 160440
> >           ](https://private-user-images.githubusercontent.com/55513145/592509138-25a5512f-0313-4453-a2a3-2d8f5ce01dd8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzg3NjQ4NTIsIm5iZiI6MTc3ODc2NDU1MiwicGF0aCI6Ii81NTUxMzE0NS81OTI1MDkxMzgtMjVhNTUxMmYtMDMxMy00NDUzLWEyYTMtMmQ4ZjVjZTAxZGQ4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTE0VDEzMTU1MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTVlNjYyNWY0NjVhMGQ2MzYyZTZmYzBhZTQyNzEwZDhhN2QwOGM1ZmU3MTJhZjIwNjJmNWM4ZmRhZDEyMWRlMzQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.0oyAoJWH3nTgo2K3mS1wbyNA8eCpOBG3WhxLIFpJcRc)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://private-user-images.githubusercontent.com/55513145/592509138-25a5512f-0313-4453-a2a3-2d8f5ce01dd8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzg3NjQ4NTIsIm5iZiI6MTc3ODc2NDU1MiwicGF0aCI6Ii81NTUxMzE0NS81OTI1MDkxMzgtMjVhNTUxMmYtMDMxMy00NDUzLWEyYTMtMmQ4ZjVjZTAxZGQ4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTE0VDEzMTU1MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTVlNjYyNWY0NjVhMGQ2MzYyZTZmYzBhZTQyNzEwZDhhN2QwOGM1ZmU3MTJhZjIwNjJmNWM4ZmRhZDEyMWRlMzQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.0oyAoJWH3nTgo2K3mS1wbyNA8eCpOBG3WhxLIFpJcRc)
> >         
> >        [ Recording 2026-05-14 160440 ](https://private-user-images.githubusercontent.com/55513145/592509138-25a5512f-0313-4453-a2a3-2d8f5ce01dd8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzg3NjQ4NTIsIm5iZiI6MTc3ODc2NDU1MiwicGF0aCI6Ii81NTUxMzE0NS81OTI1MDkxMzgtMjVhNTUxMmYtMDMxMy00NDUzLWEyYTMtMmQ4ZjVjZTAxZGQ4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTE0VDEzMTU1MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTVlNjYyNWY0NjVhMGQ2MzYyZTZmYzBhZTQyNzEwZDhhN2QwOGM1ZmU3MTJhZjIwNjJmNWM4ZmRhZDEyMWRlMzQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.0oyAoJWH3nTgo2K3mS1wbyNA8eCpOBG3WhxLIFpJcRc)
> >       
> >         [
> >           
> >         
> >             Recording 2026-05-14 160440
> >           ](https://private-user-images.githubusercontent.com/55513145/592509138-25a5512f-0313-4453-a2a3-2d8f5ce01dd8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzg3NjQ4NTIsIm5iZiI6MTc3ODc2NDU1MiwicGF0aCI6Ii81NTUxMzE0NS81OTI1MDkxMzgtMjVhNTUxMmYtMDMxMy00NDUzLWEyYTMtMmQ4ZjVjZTAxZGQ4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTE0VDEzMTU1MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTVlNjYyNWY0NjVhMGQ2MzYyZTZmYzBhZTQyNzEwZDhhN2QwOGM1ZmU3MTJhZjIwNjJmNWM4ZmRhZDEyMWRlMzQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.0oyAoJWH3nTgo2K3mS1wbyNA8eCpOBG3WhxLIFpJcRc)
> >         
> >         
> >           
> >             
> >               
> >             
> >             
> >               
> >               
> >             
> >           
> >           [
> >             
> >               
> >             
> >           ](https://private-user-images.githubusercontent.com/55513145/592509138-25a5512f-0313-4453-a2a3-2d8f5ce01dd8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzg3NjQ4NTIsIm5iZiI6MTc3ODc2NDU1MiwicGF0aCI6Ii81NTUxMzE0NS81OTI1MDkxMzgtMjVhNTUxMmYtMDMxMy00NDUzLWEyYTMtMmQ4ZjVjZTAxZGQ4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTE0VDEzMTU1MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTVlNjYyNWY0NjVhMGQ2MzYyZTZmYzBhZTQyNzEwZDhhN2QwOGM1ZmU3MTJhZjIwNjJmNWM4ZmRhZDEyMWRlMzQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.0oyAoJWH3nTgo2K3mS1wbyNA8eCpOBG3WhxLIFpJcRc)
> >         
> >        [ ](https://private-user-images.githubusercontent.com/55513145/592509138-25a5512f-0313-4453-a2a3-2d8f5ce01dd8.gif?jwt=eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJpc3MiOiJnaXRodWIuY29tIiwiYXVkIjoicmF3LmdpdGh1YnVzZXJjb250ZW50LmNvbSIsImtleSI6ImtleTUiLCJleHAiOjE3Nzg3NjQ4NTIsIm5iZiI6MTc3ODc2NDU1MiwicGF0aCI6Ii81NTUxMzE0NS81OTI1MDkxMzgtMjVhNTUxMmYtMDMxMy00NDUzLWEyYTMtMmQ4ZjVjZTAxZGQ4LmdpZj9YLUFtei1BbGdvcml0aG09QVdTNC1ITUFDLVNIQTI1NiZYLUFtei1DcmVkZW50aWFsPUFLSUFWQ09EWUxTQTUzUFFLNFpBJTJGMjAyNjA1MTQlMkZ1cy1lYXN0LTElMkZzMyUyRmF3czRfcmVxdWVzdCZYLUFtei1EYXRlPTIwMjYwNTE0VDEzMTU1MlomWC1BbXotRXhwaXJlcz0zMDAmWC1BbXotU2lnbmF0dXJlPTVlNjYyNWY0NjVhMGQ2MzYyZTZmYzBhZTQyNzEwZDhhN2QwOGM1ZmU3MTJhZjIwNjJmNWM4ZmRhZDEyMWRlMzQmWC1BbXotU2lnbmVkSGVhZGVycz1ob3N0JnJlc3BvbnNlLWNvbnRlbnQtdHlwZT1pbWFnZSUyRmdpZiJ9.0oyAoJWH3nTgo2K3mS1wbyNA8eCpOBG3WhxLIFpJcRc)
> 
> I don't think we need to handle the opacity here, but let's wait for a team member to confirm.
> 
> If we do need to adjust it, I think it should be handled with opacity styles directly instead of adding an extra DOM wrapper.

My main motivation for proposing this natively is to improve the Developer Experience (DX). Currently, to achieve a smoother visual transition, developers are forced to write custom CSS overrides or workarounds in their own applications, which can lead to style pollution and boilerplate code across different projects.
That being said, I completely respect the team's design vision and I am more than happy to wait for the rest of the maintainers to weigh in!

kfylaktopoulos:
I want to clarify why this PR intentionally uses a wrapper-based approach in single-select rendering, and why it is solving a different technical gap.
The core concern is consistency of open/search-state visuals across both plain text labels and rich custom label content, without regressing existing behavior.

1. Existing API already supports “do not tint my custom UI” use cases
For cases like icons, flags, or arbitrary decoration that should not be dimmed, users already have first-class APIs such as `prefix` and `suffixIcon`.
Those slots are explicit and predictable, and are the recommended path for persistent decorative content.

2. Color-only styling is not sufficient for rich label content
Using only `color: var(--ant-color-text-placeholder)` works for plain text, but does not reliably affect non-text nodes:
    - `image` elements

    - blocks with fixed size/background

    - custom React nodes with their own styles

That forces app developers to add extra ad-hoc CSS in product code to handle opacity/state transitions for custom nodes.
Since the component already owns the state logic, this becomes duplicated and noisy at app level.

3. Applying both parent color + child opacity can create uneven visual weight
If both mechanisms are active simultaneously, text and adjacent custom DOM can end up with different perceived transparency.
This can look inconsistent and make the selected value harder to read than intended.

4. Plain text nodes cannot be directly targeted by CSS selectors
This is the strict CSS limitation that motivates the wrapper:

selectors target elements, not arbitrary text nodes

when selected content is rendered as raw text, there is no stable element hook for state styling

So without a wrapper, there is no robust way to apply the same behavior to all render shapes.

5. Why this implementation is scoped the way it is
The implementation is intentionally constrained to avoid regressions:

default single-select labels stay primitive when labelRender is not provided (preserves built-in `title` fallback semantics)

only custom `labelRender` output in single mode is wrapped

`multiple` / `tags` mode is not forced through this wrapper path

This keeps backward compatibility while still giving a deterministic styling hook where one is technically required.

If you prefer, I can further narrow the behavior or align naming/structure with your preferred internal pattern, but the goal here is specifically to handle the selector/text-node limitation in a stable, reusable way rather than pushing that burden to downstream app CSS.

kfylaktopoulos:
Hi @afc163 👋
I just updated the branch with the latest master!

We are currently pending a maintainer's call on this. @QDyanbing raised two valid points that we'd love your input on:

1. Should this be handled natively?
@QDyanbing questioned if we need to handle this opacity at the library level. My motivation for proposing this natively is to improve DX. Currently, developers have to write custom CSS overrides in their own apps to match the default open-state behavior, which leads to style pollution.

2. Wrapper vs. Pure CSS
If we do handle it, the suggestion was to use pure CSS instead of adding a <span> wrapper. While I completely agree that avoiding extra DOM nodes is the ideal path, we hit a hard limitation here: CSS cannot directly target plain text nodes. If we apply the opacity to the parent container instead, it unintentionally dims the search input text and cursor as well. I detailed this DOM constraint [here](https://github.com/ant-design/ant-design/pull/57974#issuecomment-4454578502).

Could you weigh in on whether the wrapper is an acceptable trade-off given these technical limitations, or if you prefer a different direction? I am fully open to refactoring based on your guidance!

kfylaktopoulos:
@afc163 @meet-student I’ve updated this PR based on the maintainer feedback:

  1. Removed the extra wrapper approach.
  2. Kept the fix purely CSS-based.
  3. Preserved default single-select label behavior (no label type change), so existing `title` and a11y fallback
  behavior remains intact.
  4. Scoped the style change to open/search states only, so input text and caret visibility are unaffected.

  This keeps the fix minimal and avoids extra DOM while still covering the reported case (`labelRender` non-text content
  like `img` remaining fully visible).

  Could you please take another look and let me know if this is acceptable for merge?

kfylaktopoulos:
Hi @QDyanbing,

To recap the latest changes based on your feedback:

I completely dropped the DOM wrapper approach.
The fix is now implemented via pure CSS.
The styling is strictly scoped to the open/search states, so it handles the `labelRender` opacity perfectly while keeping the search input and caret fully visible.
I’ve updated the PR description to reflect this.
Since the implementation now directly aligns with your suggestion and all tests are passing, could you take a final look when you have a moment?

Thanks again for pointing me in the right direction!

PR #58058, opened 2026-05-19
fix(Descriptions): apply mobile-first cascade for responsive column prop

kfylaktopoulos:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

No related issue.

### 💡 Background and Solution

**Problem**

When `column` is passed as a partial responsive object (e.g. `{ xs: 1, md: 2 }`),
`Descriptions` merged it with `DEFAULT_COLUMN_MAP` before matching the active screen:

```ts
matchScreen(screens, { ...DEFAULT_COLUMN_MAP, ...column })
```

Because `DEFAULT_COLUMN_MAP` defines a value for every breakpoint (`lg: 3`, `xl: 3`,
`xxl: 3`, `xxxl: 4`), those defaults always filled the gaps the developer intentionally
left empty. On an `lg` viewport with `column={{ xs: 1, md: 2 }}`, the resolved column
count was `3` (from `DEFAULT_COLUMN_MAP.lg`) instead of the expected `2` (from `md`
cascading upward).

This is inconsistent with how `Masonry` handles its identical `columns` prop, which
already uses a correct mobile-first lookup.

**Root cause**

`matchScreen` scans breakpoints from largest to smallest and returns the first entry
where both the screen is active and the map has a defined value. Pre-filling every key
via `DEFAULT_COLUMN_MAP` guaranteed a hit at whatever the largest active breakpoint was,
before the user's lower value could be reached.

**Fix**

Try the user-supplied map first; fall back to `DEFAULT_COLUMN_MAP` only when no
user-specified key is active at all:

```ts
// Before
matchScreen(screens, { ...DEFAULT_COLUMN_MAP, ...column }) ?? 3

// After
matchScreen(screens, column) ?? matchScreen(screens, DEFAULT_COLUMN_MAP) ?? 3
```

Because Ant Design uses cumulative `min-width` media queries, on an `lg` screen the
`md` key is also active — so `matchScreen` finds `md: 2` naturally, matching
Tailwind-style mobile-first cascade semantics.

**Behaviour change**

| `column` prop | Viewport | Before | After |
|---|---|---|---|
| `{ xs:1, md:2 }` | `lg` | **3** (DEFAULT) | **2** ✓ |
| `{ xs:1, md:2, xl:4 }` | `xxl` | **3** (DEFAULT) | **4** ✓ |
| `{ xs:1, md:2 }` | `xs` | 1 | 1 ✓ unchanged |
| `{}` | `lg` | 3 (DEFAULT) | 3 (DEFAULT) ✓ unchanged |
| `{ xl:4 }` | `sm` | 2 (DEFAULT sm) | 2 (DEFAULT sm) ✓ unchanged |

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |     :ladybug: Fix `Descriptions` `column` responsive object not cascading upward in mobile-first order; unspecified higher breakpoints now inherit the last user-defined lower-breakpoint value instead of falling back to internal defaults      |
| 🇨🇳 Chinese |           |

### DIFF
--- a/components/descriptions/__tests__/index.test.tsx
+++ b/components/descriptions/__tests__/index.test.tsx
@@ -6,6 +6,8 @@ import { resetWarned } from '../../_util/warning';
 import mountTest from '../../../tests/shared/mountTest';
 import { render } from '../../../tests/utils';
 import ConfigProvider from '../../config-provider';
+import { matchScreen } from '../../_util/responsiveObserver';
+import DEFAULT_COLUMN_MAP from '../constant';
 
 describe('Descriptions', () => {
   mountTest(Descriptions);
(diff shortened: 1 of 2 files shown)


kfylaktopoulos:
Demo
https://codesandbox.io/p/sandbox/basic-grid-antd-6-4-3-forked-y8zr8h?file=%2Fdemo.tsx%3A64%2C72

afc163 (approved):

PR #58080, opened 2026-05-20
fix(input): normalise onChange event target for react-number-format customInput (#46999)

### LINKED ISSUE #46999: Antd Input: To use with react-number-format customInput feature.

gocks77777:
## Summary

Fixes #46999.

When `react-number-format` (or similar libraries) use Ant Design `Input` as
`customInput`, they intercept `onChange` internally and re-fire it with a
**transiently-created clone** as the event target rather than the real mounted
`<input>` DOM node. Consumers then receive a stale reference that is detached
from the document, causing form libraries and validation tools to break.

## Root cause

`handleChange` forwarded the event as-is. The event's `target` could be a
cloned element created by the wrapping library, not the actual mounted input.

## Fix

If the incoming event's `target` is not the live mounted input (accessible via
the existing `inputRef.current?.input`), we:

1. Temporarily patch `liveInput.value` so the native element reflects the new value during the call.
2. Create a normalised event via `Object.create` that overrides `target` and `currentTarget` with the real `liveInput`.
3. Restore `liveInput.value` in a `finally` block so React/the consuming library stays in control of reconciliation.

## Improvements over PR #57940

- Uses the **existing `inputRef`** — no new ref needed, no JSX changes
- Uses **`String()` coercion** in value comparison to safely handle numeric inputs (addresses Gemini Code Assist review comment on #57940)
- All intermediate variables carry **explicit TypeScript types** (addresses CodeRabbit review comment on #57940)
- Smaller diff — easier to review

## Checklist

- [x] Fixes the reported issue
- [x] No breaking changes (normalisation only triggers on target mismatch)
- [x] TypeScript types are explicit throughout
- [x] Uses existing internal ref — no new state or JSX changes required

### DIFF
--- a/components/input/Input.tsx
+++ b/components/input/Input.tsx
@@ -233,7 +233,7 @@ const Input = forwardRef<InputRef, InputProps>((props, ref) => {
     if (liveInput && e.target !== liveInput) {
       // Use String() coercion for safe comparison regardless of whether
       // the incoming value is a number, string, or other primitive type.
-      const eventValue: string = String(e.target.value);
+      const eventValue: string = String(e.target.value ?? '');
       const prevValue: string = liveInput.value;
       const needsPatch: boolean = prevValue !== eventValue;
 
--- a/components/input/Input.tsx
+++ b/components/input/Input.tsx
@@ -221,7 +221,43 @@ const Input = forwardRef<InputRef, InputProps>((props, ref) => {
 
   const handleChange = (e: React.ChangeEvent<HTMLInputElement>) => {
     removePasswordTimeout();
-    onChange?.(e);
+    if (!onChange) return;
+
+    // Libraries such as react-number-format fire synthetic change events
+    // whose `target` is a transiently-created clone rather than the real
+    // mounted input element. Normalise the target so that callers always
+    // receive a reference to the live input node.
+    // See: https://github.com/ant-design/ant-design/issues/46999
+    const liveInput: HTMLInputElement | null | undefined = inputRef.current?.input;
+
+    if (liveInput && e.target !== liveInput) {
+      // Use String() coercion for safe comparison regardless of whether
+      // the incoming value is a number, string, or other primitive type.
+      const eventValue: string = String(e.target.value);
+      const prevValue: string = liveInput.value;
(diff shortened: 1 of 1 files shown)


gocks77777:
Thank you @coderabbitai for the review!

Applied the suggested null-safety improvement in the latest commit:

```diff
- const eventValue: string = String(e.target.value);
+ const eventValue: string = String(e.target.value ?? '');
```

This prevents `null`/`undefined` from being stringified to `"null"`/`"undefined"` in edge cases.

---

@afc163 — would you mind taking a look when you have a moment? This fixes a real-world compatibility break for users using `react-number-format` with `customInput={Input}` (issue #46999). The fix is minimal — only the `handleChange` path where `e.target !== liveInput` is affected, so it's non-breaking for all normal usage.

gocks77777:
Also addressed the Gemini Code Assist review feedback about uncontrolled components in the latest commit:

**Before:** The `finally` block always restored `liveInput.value`, which would silently discard the new value for uncontrolled inputs (where the DOM is the source of truth).

**After:** We now check `'value' in props` — controlled inputs still restore the DOM value so React reconciliation works correctly, while uncontrolled inputs retain the new patched value.

The PR now fully addresses both automated reviewers' concerns. Summary of the 3 commits:

1. Core fix: normalize `e.target` → `liveInput` for react-number-format `customInput` (#46999)
2. Null-safety: `String(e.target.value ?? '')` per CodeRabbit suggestion
3. Uncontrolled component fix: skip DOM restore when `'value' not in props` per Gemini suggestion

gocks77777:
**Note on the Visual Regression Diff check:** This PR only modifies the `handleChange` event handler logic in `Input.tsx` — no JSX structure, styling, or rendering paths are changed. The visual output of the `Input` component should be identical before and after. The check may need a maintainer to trigger/approve since it requires human action.

afc163:

PR #58083, opened 2026-05-21
fix(Layout): improve Sider trigger accessibility

ug-one:

### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

### 💡 Background and Solution

master 代码中下面两处是绑定了 onClick 事件的静态 HTML 元素，但是没有设置 a11y 相关的 `role` 属性：

components/layout/Sider.tsx:159
components/layout/Sider.tsx:183

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | improve Sider trigger accessibility |
| 🇨🇳 Chinese | 改进 Sider 触发器的可访问性 |

### DIFF
--- a/components/layout/Sider.tsx
+++ b/components/layout/Sider.tsx
@@ -160,10 +160,13 @@ const Sider = React.forwardRef<HTMLDivElement, SiderProps>((props, ref) => {
   const rawWidth = collapsed ? collapsedWidth : width;
   // use "px" as fallback unit for width
   const siderWidth = isNumeric(rawWidth) ? `${rawWidth}px` : String(rawWidth);
+  const triggerLabel = collapsed ? 'Expand sidebar' : 'Collapse sidebar';
   // special trigger when collapsedWidth == 0
   const zeroWidthTrigger =
     Number.parseFloat(String(collapsedWidth || 0)) === 0 ? (
       <span
+        aria-expanded={!collapsed}
+        aria-label={triggerLabel}
         onClick={toggle}
         onKeyDown={handleTriggerKeyDown}
         role="button"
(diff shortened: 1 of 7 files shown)


afc163 (changes requested):
Please do not hard-code `Expand sidebar` and `Collapse sidebar`. Route these accessible labels through locale/configuration and add coverage for the localized trigger behavior.

nrps9909 (changes requested):
The localization part of the earlier maintainer feedback is addressed on the current head, but the new keyboard handling regresses custom interactive triggers.

`role="button"`, `tabIndex`, `aria-*`, and `onKeyDown` are currently added to the wrapper unconditionally. With a supported custom trigger such as `trigger={<button type="button">Custom trigger</button>}`, this creates a button inside a `role="button"` and makes one Enter activation toggle twice: the native button's keydown bubbles to the wrapper and calls `toggle()`, then the browser-generated click bubbles to the wrapper and calls it again. The sidebar therefore returns to its original state and `onCollapse` fires twice.

I reproduced this at exact head `9a47b4c79bd22ec1407ddf76203573b93d27d398` with a focused Layout regression:

```tsx
fireEvent.keyDown(button, { key: 'Enter' });
fireEvent.click(button);
expect(onCollapse).toHaveBeenCalledTimes(1);
```

It fails with 2 calls; a characterization assertion confirms they are `(true, 'clickTrigger')` followed by `(false, 'clickTrigger')`. The focused suite otherwise passes (34 tests, 6 snapshots).

Please leave activation semantics consumer-owned when `trigger` is a custom interactive node (and avoid making its wrapper a second interactive control), while retaining the keyboard/ARIA fallback for the built-in icon trigger. A regression with a custom native `<button>` would protect both the DOM semantics and the single-toggle behavior. The approach previously proposed in #58818—only attaching wrapper accessibility props for the built-in trigger and guarding `event.target === event.currentTarget`—also avoids this failure mode.

PR #58104, opened 2026-05-23
fix(AutoComplete): restrict showSearch type to prevent unsupported Select props leaking

### LINKED ISSUE #58097: bug/feature: AutoComplete showSearch type definition leaks Select props that are not supported/documented
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/sandbox/cocky-grothendieck-f5gxmv)

### Steps to reproduce

1. Create a basic AutoComplete component with distinct value and label properties in the options array.
2. Add the showSearch prop and configure it using optionFilterProp: 'label', which TypeScript fully suggests and accepts via autocomplete/IntelliSense.
3. Type a search term. The items never get filtered.

### What is expected?

TypeScript suggests these properties because they are part of the Select types shared with AutoComplete. Therefore, either:

The AutoComplete component should honor the configuration and correctly filter the dropdown options by the specified property (e.g., 'label').

If AutoComplete is architecturally intended to only support filterOption and onSearch, TypeScript should throw a type error when trying to pass optionFilterProp, filterSort, or other unsupported Select parameters.

The configuration fails silently at runtime. Typing "Marketing" yields no results because AutoComplete completely ignores optionFilterProp and falls back to its default behavior, while TypeScript misleadingly signals that the code is perfectly valid.

| Environment | Info |
| --- | --- |
| antd | 6.4.3 |
| React | 18.3.1 |
| System | Windows 11 |
| Browser | Latest Chrome |

---
The documentation at [https://ant.design/components/auto-complete#showsearch](https://ant.design/components/auto-complete#showsearch) correctly lists only filterOption and onSearch. However, the underlying type definition leaks the entire Select search configuration interface.

Fixing the type definition by using something like Pick<SelectProps['showSearch'], 'filterOption' 'onSearch' |> would immediately prevent developers from stumbling into this trap. Out of curiosity: Is there a specific architectural reason why AutoComplete cannot support optionFilterProp natively like Select does? I would really love to see optionFilterProp in Autocomplete as well!

sri443:
### 🤔 This is a ...

### 🔗 Related Issues

> - fix #58097

### 💡 Background and Solution

> - `AutoCompleteProps` inherited the full `SearchConfig` interface via `InternalSelectProps`, leaking unsupported props like `optionFilterProp` and `filterSort` into AutoComplete's type definition with no TypeScript error.
> - Fixed by omitting `showSearch` from the inherited type and re-declaring it restricted to only `filterOption`, `onSearch`, and `searchIcon` — the props AutoComplete actually supports.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    Fix AutoComplete showSearch type leaking unsupported Select props       |
| 🇨🇳 Chinese |     修复 AutoComplete showSearch 类型泄露不支持的 Select 属性的问题      |


MadCcc (approved):

sri443:
Could a maintainer re-trigger the visual regression check or approve the diff? The failure appears to be a CI infrastructure issue unrelated to this type-only change.

PR #58149, opened 2026-05-27
fix: remove subpixel offset from Steps rail

### LINKED ISSUE #58112: Steps: Rail line appears 2px thick with washed-out color due to subpixel rendering from translateY(-50%)
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/devbox/zealous-mountain-trr7cy)

### Steps to reproduce

1. Render a horizontal Steps component with at least two items
2. Observe the connecting rail line between steps — visible on the official ant.design documentation page as well

### What is expected?

The rail line between steps renders as a crisp 1px line (--ant-cmp-steps-rail-size) with the exact color defined by --ant-cmp-steps-item-solid-line-color.

The rail line appears ~2px thick and the color looks desaturated/semi-transparent compared to the actual token value. This is reproducible on the ant.design docs site itself.

Root cause
The .ant-steps-item-rail element is positioned with: transform: translateY(-50%);

When the parent element has an odd pixel height, translateY(-50%) resolves to a subpixel offset (e.g. 10.5px). The browser then distributes the 1px border across two physical pixels using anti-aliasing, which causes:

The line to visually appear ~2px thick
The color to appear washed out / lower opacity (the actual color value is split across two rows of pixels)

Suggested fix
The transform: translateY(-50%) can likely be removed entirely in favor of a flexbox alignment approach, which avoids subpixel positioning altogether.

| Environment | Info |
| --- | --- |
| antd | 6.4.3 |
| React | 19.2.1 |
| System | Windows 11 |
| Browser | Latest Chrome |

---
This issue is visible on the ant.design documentation site itself under the Steps component examples, confirming it is not a consumer-side configuration issue. The bug affects both the finish state line (primary color) and the wait state line, but is most noticeable on the colored finish rail where the color difference to the expected token value is clearly visible.

olagokemills:
### 🤔 This is a ...
- [x] 🐞 Bug fix

### 🔗 Related Issues
Fix #58112

### 💡 Background and Solution
Horizontal Steps rails were vertically offset with `translateY(-50%)`. For a 1px border this can place the rail on a half-pixel and Chrome paints it across two rows, making the rail look thicker and washed out.

This removes the transform from the horizontal rail style while preserving the existing flex layout, rail width token, and status colors. A focused regression test verifies the generated horizontal rail rule no longer emits a transform.

### 📝 Change Log

| Language | Changelog |
| --- | --- |
| 🇺🇸 English | Fix Steps horizontal rail rendering as a blurry 2px line in some browsers. |
| 🇨🇳 Chinese | 修复 Steps 水平连接线在部分浏览器中显示为模糊 2px 线的问题。 |


components/steps/__tests__/index.test.tsx
```diff
@@ -125,6 +126,31 @@ describe('Steps', () => {
...
+      /[^{}]*\.ant-steps-horizontal\s*>\.ant-steps-item \.ant-steps-item-rail\{(?<style>[^}]*)\}/.exec(
+        styleText,
+      )?.groups?.style;
+
+    expect(horizontalRailStyle).toBeTruthy();
+    expect(horizontalRailStyle).toContain(
+      'border-block-start-width:var(--ant-cmp-steps-rail-size)',
+    );
+    expect(horizontalRailStyle).toContain('align-self:flex-start');
+    expect(horizontalRailStyle).not.toContain('transform:');
```
QDyanbing: This test case doesn't seem to add much value.

components/steps/style/horizontal.ts
```diff
@@ -23,7 +23,6 @@ const genHorizontalStyle: GenerateStyle<StepsToken, CSSObject> = (token) => {
           flex: 1,
           minWidth: 0,
           alignSelf: 'flex-start',
-          transform: 'translateY(-50%)',
         },
```
QDyanbing, line 26: Removing it directly may affect the centering layout.

QDyanbing: marginTop

components/steps/__tests__/index.test.tsx
```diff
@@ -125,6 +126,27 @@ describe('Steps', () => {
...
+    );
+
+    const styleText = extractStyle(cache, true);
+    const horizontalRailStyle =
+      /\{[^}]*border-block-start-width:var\(--ant-cmp-steps-rail-size\)[^}]*align-self:flex-start[^}]*\}/.exec(
+        styleText,
+      )?.[0];
+
+    expect(horizontalRailStyle).toBeTruthy();
+    expect(horizontalRailStyle).not.toContain('transform:');
```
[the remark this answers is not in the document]
olagokemills: Good call. Updated the selector regex to use `\s*` around the `>` combinator and `\s+` between class names so it's tolerant of any whitespace variation. Also loosened the `border-block-start-width` assertion to not depend on the exact variable name — each property is asserted independently so order doesn't matter.

components/steps/__tests__/index.test.tsx
olagokemills: The test is a regression guard to ensure the subpixel blur (#58112) can't silently return. I've also added a `type="dot"` variant test case to expand coverage to the 2px rail path.

components/steps/style/horizontal.ts
olagokemills: Agreed — updated the `marginTop` formula to `calc(icon-size-max / 2 - rail-size / 2 + item-wrapper-padding-top)`. Subtracting `rail-size / 2` shifts the rail up by half its thickness so its visual centre aligns with the icon centre, rather than just its top edge. This correctly handles both the 1px standard rail and the 2px dot-variant rail without any transform.

components/steps/__tests__/index.test.tsx
```diff
@@ -125,6 +126,54 @@ describe('Steps', () => {
...
+    const styleText = extractStyle(cache, true);
+    const dotRailStyle =
+      /[^{}]*\.ant-steps-horizontal\s*>\s*\.ant-steps-item\s+\.ant-steps-item-rail\s*\{(?<style>[^}]*)\}/.exec(
+        styleText,
+      )?.groups?.style;
+
+    expect(dotRailStyle).toBeTruthy();
+    expect(dotRailStyle).toContain('align-self:flex-start');
+    expect(dotRailStyle).not.toContain('transform:');
+  });
```
afc163: 不需要这个用例。

components/steps/__tests__/index.test.tsx
QDyanbing: 主要是这测试用例是在用具体的实现来测试功能，结果就只是A=A；起不到防护的作用

QDyanbing:
Hi @olagokemills, are you still working on this PR? Thanks!

components/steps/__tests__/index.test.tsx
afc163: ping

olagokemills:
> Hi @olagokemills, are you still working on this PR? Thanks!

Yes, I am. I will address this comment today and push again

components/steps/__tests__/index.test.tsx
olagokemills: Okay, I had to translate this, sorry. I see your point and I will remove it

components/steps/__tests__/index.test.tsx
olagokemills: Okay, will fix

components/steps/__tests__/index.test.tsx
olagokemills: I have updated the PR as discussed

components/steps/__tests__/index.test.tsx
olagokemills: I have updated the PR as discussed

afc163 (approved):
LGTM, thanks!

PR #57938, opened 2026-05-12
feat: support semantic styles for Layout Sider

landeqiming666:
## Summary

Fixes #13527.

This adds semantic `classNames` and `styles` support to `Layout.Sider`, allowing consumers to style the internal Sider content wrapper directly without adding an extra wrapper element.

This supersedes #57937 with a clean branch: the final diff only touches `components/layout/*` implementation, tests, and docs. It does not modify `.github/`, `scripts/`, or changelog files.

## Changes

- Add `classNames` and `styles` props to `SiderProps`
- Merge semantic class/style props with `useMergeSemantic`
- Apply the `body` semantic class/style to the existing `.ant-layout-sider-children` wrapper
- Add regression coverage for applying semantic styles to the Sider body wrapper, not the root
- Document the `body` semantic target in English and Chinese docs

## Verification

- `npm test -- components/layout/__tests__/index.test.tsx --runInBand`
- `npm run lint:biome -- components/layout/Sider.tsx components/layout/__tests__/index.test.tsx components/layout/index.en-US.md components/layout/index.zh-CN.md`
- `git diff --check`

I also checked `npm run tsc`; the current local checkout fails on existing generated/dependency type issues outside Layout (for example `.dumi` and `@rc-component/*` type exports), so the targeted Layout verification above is the reliable signal for this PR scope.

---

<details>
<summary>
<b>IssueHunt Summary</b>
</summary>

### Referenced issues

This pull request has been submitted to:
- [#13527: Make it possible to directly turn a Slider into a flexbox](https://oss.issuehunt.io/repos/34526884/issues/13527)
---
</details>

PR #58179, opened 2026-05-30
fix: add bodyStyle to Layout.Sider

### LINKED ISSUE #13527: Make it possible to directly turn a Slider into a flexbox

pc-style:
Closes #13527\n\n/claim #13527\n\nAdds a bodyStyle prop to Layout.Sider so the inner wrapper can be styled directly, which makes flexbox layouts work without an extra wrapper div.\n\nValidation: bun run version; bun x jest --config .jest.js --runInBand components/layout/__tests__/index.test.tsx --no-cache; bun x biome check components/layout/Sider.tsx components/layout/__tests__/index.test.tsx components/layout/index.en-US.md components/layout/index.zh-CN.md; bun x prettier -c components/layout/index.en-US.md components/layout/index.zh-CN.md

### DIFF
--- a/components/layout/Sider.tsx
+++ b/components/layout/Sider.tsx
@@ -71,7 +71,7 @@ const Sider = React.forwardRef<HTMLDivElement, SiderProps>((props, ref) => {
     className,
     trigger,
     children,
-    bodyStyle = {},
+    bodyStyle,
     defaultCollapsed = false,
     theme = 'dark',
     style = {},
--- a/components/layout/Sider.tsx
+++ b/components/layout/Sider.tsx
@@ -40,7 +40,12 @@ export interface SiderProps extends React.HTMLAttributes<HTMLDivElement> {
   collapsible?: boolean;
   collapsed?: boolean;
   defaultCollapsed?: boolean;
-  bodyStyle?: React.CSSProperties;
+  classNames?: {
+    body?: string;
+  };
+  styles?: {
+    body?: React.CSSProperties;
+  };
   reverseArrow?: boolean;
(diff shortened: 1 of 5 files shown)


afc163:
The API should be sementic styles, `styles={{ body: { ... } }}`

pc-style:
> The API should be sementic styles, `styles={{ body: { ... } }}`

oh, okay, ill fix it!

pc-style:
@afc163 is it ok now?

afc163:
feature branch

afc163:
Fixed in https://github.com/ant-design/ant-design/pull/57938

Issue #50414, opened 2024-08-14
Typography.Text会始终显示Tooltip


fpsqdb:
### Reproduction link

[![Edit on StackBlitz](https://developer.stackblitz.com/img/open_in_stackblitz.svg)](https://stackblitz.com/edit/react-mhogkk?file=demo.tsx)

### Steps to reproduce

鼠标移动到文本上
![image](https://github.com/user-attachments/assets/eb496329-957e-472a-b282-ec1287dad98f)

### What is expected?

能显示全的情况下不应该显示Tooltip

### What is actually happening?

显示Tooltip

| Environment | Info |
| --- | --- |
| antd | 5.20.1 |
| React | 18.3.1 |
| System | windows 11 |
| Browser | chrome latest |

zombieJ:
重现不出来，是不是有设置过屏幕缩放？

fpsqdb:
我这缩放200% 2880x1800、无缩放1920x1080、无缩放2560x1440都能重现，之前5.19.0版本是没有问题的
这是5.20.1
![GIF 2024-08-14 14-10-19](https://github.com/user-attachments/assets/6cd93428-822f-4338-a6d2-d29e6378f5a6)

这是5.19.0
https://stackblitz.com/edit/react-mhogkk-mfcnvj?file=demo.tsx

![GIF 2024-08-14 14-22-29](https://github.com/user-attachments/assets/844757d8-6260-4e2e-b83d-585c80b07e6b)

fpsqdb:
在火狐浏览器上无法重现，chrome可以，我的chrome版本是Version 127.0.6533.120 (Official Build) (64-bit)

Dmaziyo:
![image](https://github.com/user-attachments/assets/5c5d065f-fc62-4013-a809-e51d7fdb15e5)
看了下好像是精度的问题，scrollWidth和offsetWidth不一样导致的
React.useEffect(() => {
    const textEle = typographyRef.current;

    if (enableEllipsis && cssEllipsis && textEle) {
      if(className?.includes('cesss')){
        debugger
      }
      const [offsetWidth, offsetHeight] = getEleSize(textEle);

      const currentEllipsis = cssLineClamp
        ? offsetHeight < textEle.scrollHeight
        : offsetWidth < textEle.scrollWidth;

      if (isNativeEllipsis !== currentEllipsis) {
        setIsNativeEllipsis(currentEllipsis);
      }
    }
[...]

PR #58211, opened 2026-06-04
fix: Typography ellipsis tooltip false positive with small line-height

### LINKED ISSUE #56347: Typography.Paragraph会一直展示tooltip
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/sandbox/nostalgic-dubinsky-f9xsqy)

### Steps to reproduce

使用Poppins字体并且设置行高为18px时，在文本不省略时也会有hover展示tooltip的效果

### What is expected?

使用Poppins字体并且设置行高为18px时，在文本不省略时不触发hover展示tooltip的效果

我注意到，在使用Poppins，并设置行高为18px时，容器高度为18px，但插入的<em /> 的高度为20px，导致判断错误

| Environment | Info |
| --- | --- |
| antd | 5.29.1 |
| React | 18.3.26 |
| System | Mac os 14.6.1 (23G93) |
| Browser | Chrome 141.0.7390.77（正式版本） (arm64) |

RykerFeng:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

fix #56347

### 💡 Background and Solution

When `Typography` enables `ellipsis` with a `tooltip`, it detects native (CSS) ellipsis by injecting an `<em>` probe at the end of the content and checking whether the probe's `getBoundingClientRect` exceeds the container's rect.

The problem: an `<em>` is an inline element, so its measured height comes from the **font box** (ascent + descent), which is independent of `line-height`. When `line-height` is smaller than the font's natural box (e.g. a 14px font with an 18px line on font families such as Poppins), the probe sticks out vertically above and below the line box. `isEleEllipsis` then reads `childRect.bottom > rect.bottom` as an overflow and shows a tooltip **even though the text is not actually clamped**.

**Solution:** collapse the probe to a zero-sized `inline-block` (`display: inline-block; width: 0; height: 0`) so it no longer contributes any size of its own. This removes the spurious vertical overflow while keeping the horizontal overflow detection that was introduced for #50143 / #50414 intact (verified in a real browser: short text no longer triggers the tooltip, while genuinely truncated single-line and multi-line text still do).

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Typography that incorrectly shows the ellipsis tooltip when the text is not truncated under a small `line-height`. |
| 🇨🇳 Chinese | 修复 Typography 在较小 `line-height` 下文本未被省略时仍错误展示省略 tooltip 的问题。 |

### DIFF
--- a/components/typography/Base/util.ts
+++ b/components/typography/Base/util.ts
@@ -33,9 +33,15 @@ export function isEleEllipsis(ele: HTMLElement): boolean {
   // overflow even though the text is not clamped. Collapsing it to a zero
   // sized inline-block keeps the horizontal detection (see #50143 / #50414)
   // while avoiding the vertical false positive (see #56347).
+  // `margin`, `padding` and `border-width` are also reset so that global
+  // stylesheets targeting `em` (resets or custom styles) can not re-introduce
+  // any size and bring back the false positive.
   childDiv.style.display = 'inline-block';
   childDiv.style.width = '0';
   childDiv.style.height = '0';
+  childDiv.style.margin = '0';
+  childDiv.style.padding = '0';
+  childDiv.style.borderWidth = '0';
   ele.appendChild(childDiv);
 
   // For test case
--- a/components/typography/Base/util.ts
+++ b/components/typography/Base/util.ts
@@ -21,8 +21,21 @@ export function getNode(dom: React.ReactNode, defaultNode: React.ReactNode, need
  * - https://github.com/ant-design/ant-design/issues/50414
  */
 export function isEleEllipsis(ele: HTMLElement): boolean {
-  // Create a new div to get the size
+  // Create a new node to probe the layout box.
   const childDiv = document.createElement('em');
+
+  // The probe is only used to detect whether the content overflows the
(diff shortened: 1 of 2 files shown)


afc163:
The zero-size probe is a focused fix for the small-row-height misdetection. Please rebase onto the current `master` and refresh the visual checks before review.

nrps9909 (approved):
Validated exact head `64dd82cc447bb9b831ea944b2004c7004fa9ac80`. A real Chromium probe with Poppins at 14px/18px reproduced the bug: the current inline `<em>` measured 20px high against the 18px container and falsely returned ellipsis for short text. With this head’s zero-sized inline-block probe, that case returns false, while genuinely truncated single-line text and a two-line clamp both still return true. Local verification: the complete ellipsis file passes 29/29 tests; focused ESLint, Biome, and diff checks pass. The substantive remote React 18/latest, module/dist, coverage, build, lint, size, and preview checks pass. The lone red visual shard is unrelated runner failure: Cascader image cleanup hooks timed out after 21/22 suites passed, so no visual report was produced. No blocking source findings.

PR #58216, opened 2026-06-04
fix: Input.OTP duplicate characters with CJK input methods

### LINKED ISSUE #52093: 使用微软自带的输入法，切换至中文模式，使用Input.OTP,按一下任一字符,就会输入进2个字符
### Reproduction link

[https://ant.design/components/input-cn](https://ant.design/components/input-cn)

### Steps to reproduce

直接使用官网的例子,聚焦到Input.OTP,使用微软自带的输入法，切换至中文模式，使用Input.OTP,按一下任一字符,就会输入进2个字符

### What is expected?

输入一个字符只会有一个字符

输入一个字符有两个字符

| Environment | Info |
| --- | --- |
| antd | 5.22.5 |
| React | 18.18.0 |
| System | windows11 |
| Browser | 谷歌浏览器 |

18062706139fcz:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

close #52093

### 💡 Background and Solution

`Input.OTP` listens to the native `input` event to report changes. During an IME (CJK) composition session, the `input` event keeps firing with the intermediate value (e.g. the pinyin `wo` before committing `我`), so every intermediate keystroke was written into the cell, resulting in duplicated/extra characters when typing Chinese with the Microsoft IME and similar input methods.

A previous attempt (#54851) only skipped `onChange` when `e.nativeEvent.isComposing` was true, but that broke other IMEs (QQ / Sogou) where no reliable non-composing `input` event follows to commit the final value, leaving the user unable to type anything at all.

This PR tracks the composition state explicitly:

- On `compositionstart`, mark composing and ignore the intermediate `input` values.
- On `compositionend`, commit the final value once, so every IME works regardless of the `input` / `compositionend` ordering.
- De-dupe the trailing `input` that some browsers (e.g. Safari) emit right after `compositionend` with the already-committed value.
- Skip reporting when a composition ends without a real change (cancelled by `Esc` / blur).

Tests cover: intermediate value ignored while composing, normal input after composing, trailing `input` de-dup, multi-character commit across cells, and cancelled composition.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix `Input.OTP` typing duplicate characters with CJK (IME) input methods. |
| 🇨🇳 Chinese | 修复 `Input.OTP` 使用中文等输入法时会重复输入字符的问题。 |

### DIFF
--- a/components/input/OTP/OTPInput.tsx
+++ b/components/input/OTP/OTPInput.tsx
@@ -71,6 +71,13 @@ const OTPInput = React.forwardRef<InputRef, OTPInputProps>((props, ref) => {
     }
     composedValueRef.current = nextValue;
     onChange(index, nextValue);
+    // The trailing `input` (if any) arrives synchronously right after
+    // `compositionend`. Clear the ref in the next tick so a browser that
+    // never fires that trailing `input` can not leave a stale value behind
+    // and block typing the same character again later.
+    raf(() => {
+      composedValueRef.current = null;
+    });
   };
 
   // ========================= Focus ==========================
--- a/components/input/OTP/OTPInput.tsx
+++ b/components/input/OTP/OTPInput.tsx
@@ -26,9 +26,51 @@ const OTPInput = React.forwardRef<InputRef, OTPInputProps>((props, ref) => {
 
   React.useImperativeHandle(ref, () => inputRef.current!);
 
+  // ====================== Composition =======================
+  // Track IME composition so that the intermediate value of CJK input methods
+  // is not written into the cell on every `input` event (see #52093).
+  const compositionRef = React.useRef(false);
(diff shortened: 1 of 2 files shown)


afc163:
This is the stronger OTP IME candidate because it explicitly commits on `compositionend`, deduplicates a trailing `input`, and includes focused event-order tests. Please rebase and restore green coverage checks; #58762 should align with this behavior rather than replace it.

afc163:
Closing in favor of #58762, which now implements the same IME commit-and-deduplicate model while also forwarding consumer composition callbacks, guarding native isComposing state, covering cleanup and trailing-input cases, and passing the required checks. #58762 is the canonical candidate.

PR #58237, opened 2026-06-06
feat(Select): support passing native props to search input

### LINKED ISSUE #58109: Select: No way to set autoComplete attribute on the underlying <input> element (useful for showSearch)
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/devbox/clever-agnesi-42g4tx?workspaceId=ws_UJWDJFW6u5eT5bSdmMxLcC)

### What is expected?

At least one of the following should be supported:

Option A — Pass autoComplete directly as a prop to `<Select>`:
`<Select autoComplete="new-password" />`

Option B — Pass a custom <Input> element with autoComplete via `children`:
`<Select><Input autoComplete="new-password" /></Select>`

(The components.input path in Input.js already supports React elements via cloneElement and merges existingProps on top of sharedInputProps — but antd does not expose this components.input override publicly.)

Any autoComplete prop passed to <Select> ends up on the outer wrapper <div>, not the <input>. The underlying input always receives autoComplete="off".

| Environment | Info |
| --- | --- |
| antd | 6.4.3 |
| React | 19.2.1 |
| System | Windows 11 |
| Browser | Latest Chrome |

---
With `AutoComplete`, it is possible to pass a child Input component, which allows setting native HTML props such as autoComplete directly on the underlying input element:

```
<AutoComplete>
  <Input autoComplete="off" />
</AutoComplete>
```

However, with `Select`, there seems to be no equivalent way to pass custom props (e.g. autoComplete) to the inner input element. Is there a supported API for this, or would it be possible to add something like an inputProps or styles/classNames extension that also covers native input attributes?

guozi9999:
### 🤔 This is 

- [x] 🆕 New feature
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

close #58109

### 💡 Background and Solution

When Select enables search, native attributes cannot be passed to the internal search input. For example, users cannot configure `autoComplete` on the actual input element.

This PR adds `inputProps` to pass native input attributes to Select's internal search input:

```tsx
<Select inputProps={{ autoComplete: 'new-password' }} showSearch />
```

It also updates the API documentation and adds a regression test to ensure inputProps.autoComplete is applied to the internal input instead of the outer .ant-select element.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| English | Add Select inputProps to pass native attributes to the internal search input. |
| Chinese | 新增 Select inputProps，用于向内部搜索输入框传递原生属性。 |

### DIFF
--- a/components/auto-complete/AutoComplete.tsx
+++ b/components/auto-complete/AutoComplete.tsx
@@ -19,6 +19,8 @@ import type {
   SelectSemanticAllType,
 } from '../select';
 import Select from '../select';
+import { getInputPropsInput } from '../select/inputProps';
+import type { SelectInputComponent } from '../select/inputProps';
 
 export type AutoCompleteSemanticType = {
   classNames?: {
@@ -58,7 +60,15 @@ export interface AutoCompleteProps<
   OptionType extends BaseOptionType | DefaultOptionType = DefaultOptionType,
 > extends Omit<
     InternalSelectProps<ValueType, OptionType>,
-    'loading' | 'mode' | 'optionLabelProp' | 'labelInValue' | 'styles' | 'classNames' | 'showSearch' | 'optionFilterProp' | 'filterSort'
+    | 'loading'
+    | 'mode'
+    | 'optionLabelProp'
+    | 'labelInValue'
+    | 'styles'
+    | 'classNames'
+    | 'showSearch'
+    | 'optionFilterProp'
+    | 'filterSort'
   > {
   /** @deprecated Please use `options` instead */
   dataSource?: DataSourceItemType[];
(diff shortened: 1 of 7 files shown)


afc163 (changes requested):
如果已经能通过 components.input 自定义了，inputProps 属性感觉是多余的了。

guozi9999:
感谢 review！您说得对，`components.input` 确实可以实现同样的功能，`inputProps` 显得多余了。

这个 PR 先关闭了，感谢您的时间！🙏

PR #58267, opened 2026-06-08
feat(Cascader, Select, TreeSelect): improve clear button accessibility

Pareder:

### 🤔 This is a ...

### 💡 Background and Solution

This PR adds several accessibility improvements to the clear button:

- displaying clear button not only on hover, but also on focus
- adding focus styles to the clear button
- adding `clear` prop to locale and passing it to the button

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | feat(Cascader, Select, TreeSelect): improve clear button accessibility |
| 🇨🇳 Chinese | feat(Cascader, Select, TreeSelect)：改进清除按钮的可访问性 |


afc163 (dismissed):

Pareder:
@afc163 This PR makes sense only together with https://github.com/react-component/select/pull/1224.

afc163:
https://github.com/react-component/select/releases/tag/%40rc-component%2Fselect%401.9.0

afc163:
https://github.com/react-component/select/releases/tag/%40rc-component%2Fselect%401.9.0 is released, please upgrade in this PR.

Pareder:
@afc163 Done. Checks failed due to some updates in master branch.

afc163:
ci failed

afc163:
Thanks for updating this after `@rc-component/select` was released.

This PR still needs CI cleanup before review can continue. The latest run has failures in build, size, React legacy/latest DOM tests, and es module tests, plus the visual regression approval check is failing. Please check whether these are caused by the dependency update or by the locale/snapshot changes in this PR.

Also, the changed file list is much broader than the Select clear-button integration. If some locale or unrelated snapshot updates are only incidental churn, please remove them so this PR stays focused on the Select clear-button accessibility fix and the required `@rc-component/select` dependency update.

Pareder:
@afc163 Reverted unnecessary changes (that were made by automatically linting during merge).
- `test/build` and `build` jobs are failing because `@rc-component/cascader` dependency still needs `1.8.0` version of `@rc-component/select`
- `Size limit` job is failing because size was probably increased by locale update
- `Visual regression` job is failing because I corrected the centering of icon by changing `lineHeight: 0` (the same way it works in other components with suffix icons)

Pareder:
@afc163 To fix the build issue I've created a PR in `@rc-component/cascader` library https://github.com/react-component/cascader/pull/648 bumping `@rc-component/select` version to v1.9.0, it is needed to merge it, deploy and update in this PR.

Pareder:
@afc163 Changed base branch to `feature` and rebased PR.

Pareder:
@afc163 @zombieJ Could you please review?

Pareder:
@afc163 @zombieJ Could you please review?

Pareder:
@afc163 @zombieJ Could you please review?

Pareder:
@afc163 @zombieJ Could you please review?

Pareder:
@afc163 @zombieJ Could you please review?

Pareder:
@afc163 @zombieJ Could you please review?

afc163 (changes requested):
The clear-button accessibility direction is valid, but the new Select behavior needs focused regression tests. Please also restore the required PR template before review.

Pareder:
@afc163 Done.

afc163:
The new focused Select tests address the previous review. Two localization gaps remain: `components/locale/sq_AL.ts` is the only locale missing `global.clear`, and Cascader/TreeSelect still forward `{ clearIcon }` without the localized label even though their rc clear controls use the same accessible button path. Please complete those before approval.

Pareder:
@afc163 Done. Also updated PR title and description to include Cascader and TreeSelect reference.

afc163 (dismissed):
The previously missing sq_AL clear label and the Cascader and TreeSelect localized labels are present, while Select clear/reset/focus behavior and custom allowClear content are covered. Required checks are green.

Pareder:
@afc163 Can it be merged please?

components/select/__tests__/index.test.tsx
```diff
@@ -449,6 +450,69 @@ describe('Select', () => {
...
+      clearBtn.focus();
+      expect(clearBtn).toHaveFocus();
+    });
+
+    it('should clear the value when the clear button is activated from the keyboard', () => {
+      const onClear = jest.fn();
+      const { container } = render(<Select {...props} allowClear onClear={onClear} />);
+      expect(container.querySelector('.ant-select-content-has-value')).toHaveTextContent('Jack');
+
+      fireEvent.click(container.querySelector('.ant-select-clear')!);
```
yoyo837: **[P2] Exercise the real keyboard activation path**

Despite the test name, `fireEvent.click()` bypasses keyboard handling. With the current `@rc-component/select`, pressing Enter or Space on the focused clear `<button>` bubbles to `BaseSelect.onInternalKeyDown`, whose `preventDefault()` cancels the button's native activation, so its `onClick` never clears the value. I reproduced this in headless Chrome: both keys produced zero clicks. Please fix the event interaction (for example, stop the clear button's Enter/Space keydown from reaching the Select handler, or make the parent ignore button targets) and cover it with actual keyboard input such as `userEvent.keyboard('{Enter}')` and Space.

yoyo837 (commented):
The localization and focus-style gaps from earlier reviews are fixed. One keyboard-path issue remains.

Pareder:
@yoyo837 Fixed keyboard press behavior in `@rc-component/select` library https://github.com/react-component/select/pull/1247. After merging it I will update package version here.

Pareder:
@yoyo837 Bumped `@rc-component/select` package with latest fix and added keyboard tests.

yoyo837 (commented):
Thanks for the continued work on this accessibility improvement, @Pareder. The implementation looks solid and the test coverage is comprehensive. I have a few observations:

**What looks great:**
- The `allowClear` label approach is clean, reusing existing infrastructure without API changes
- Keyboard activation tests (Enter/Space) are a nice addition
- The `focus-within` + `focus-visible` style changes are consistent with how other components handle focus

**One question on locale translations:**
I notice this PR adds `clear` translations across 72+ locale files. While the translations appear plausible, I wonder whether we should encourage native speakers to review or contribute the translations for languages beyond English/Chinese (which the maintainers can verify). For example:
- Georgian `"გასუფთავება"` — casual check suggests this means "cleaning" rather than "clear (the selection)" 
- There may be similar nuance issues in other languages where the word choice depends on context

Would it make sense to merge only the English (`en_US`/`en_GB`) and Chinese (`zh_CN`/`zh_HK`/`zh_TW`) translations initially, and open a follow-up issue asking native speakers to contribute the remaining locale entries? This would unblock the accessibility fix while ensuring language quality. Or alternatively, keep all translations but mark them as needing review?

**Minor:**
- The `@rc-component/select` bump and cascader dependency chain is well-handled — thanks for filing the upstream PRs too.
- `line-height: 1` → `0` is a visual alignment fix for the clear icon centering; visual-diff will catch any regressions.

Overall the direction is good and the code quality is high. My only hesitation is the non-English locale coverage.

Pareder:
@yoyo837 What is the common practice for handling translations? It is up to you to decide what to do.

yoyo837:
I personally prefer adding translations for Chinese and English only, with automatic fallback to English for other languages. :)

Pareder:
@yoyo837 I'll delete all other except English and Chinese translations. One question: is Chinese translation `清除` correct?

yoyo837:
> @yoyo837 I'll delete all other except English and Chinese translations. One question: is Chinese translation `清除` correct?

Yes. Correct.

Pareder:
@yoyo837 Done. Also added a test case for testing fallback value.

yoyo837:
Thanks for the PR! The accessibility improvements look good overall. A few small comments:

1. The `lineHeight: 1` → `lineHeight: 0` change and the new `:focus-within` behavior are easy to misread as unrelated changes or regressions. Could you add short inline comments explaining why they are needed (e.g. keeping the focus outline tight, and clear now showing on focus)?

2. The new tests only cover `Select`. `Cascader` and `TreeSelect` get the same `label` + focus changes but have no coverage — could you add a minimal `aria-label` assertion for each?

3. `userEvent.setup({ advanceTimers: jest.advanceTimersByTime })` references `jest.advanceTimersByTime` while `jest.useFakeTimers()` is never enabled in this file. Since `keyboard` has no delay it's likely never invoked, but it's misleading — either drop the option or enable fake timers explicitly.

nrps9909 (approved):
Reviewed exact head `4124d3092a6eba57751726995157db44da45d61f`.

The current revision resolves the earlier localization scope concern cleanly: English and Chinese locales provide native strings, while `useLocale('global')` preserves the English default for untranslated locales. The `allowClear` merge order also keeps an explicit consumer label authoritative. I verified those paths through Select, Cascader, and TreeSelect, including keyboard focus and native Enter/Space activation of the clear button.

Validation performed:

- Select index suite: 60 tests and 3 snapshots passed.
- Cascader and TreeSelect index/a11y suites: 117 tests and 19 snapshots passed (1 existing skip).
- Independent integration probe: localized Cascader label + Enter activation, custom TreeSelect label + Space activation, and untranslated-locale fallback all passed.
- Focused ESLint and repository-wide Biome lint passed.
- The exact-head CI matrix is green, including build, React 18/latest, lib/es/dist, coverage, visual regression, size, and security checks.

I found no blocking issue in this revision.

AI assistance disclosure: Codex was used to inspect the exact-head diff and review history, run the focused/full relevant tests and independent behavior probe, audit the live CI state, and draft this review. I independently verified the conclusions against the checked-out source and current GitHub state.

Pareder:
@yoyo837 
1. Added comments.
2. `Cascader` and `TreeSelect` reuse `Select` component styles so added `aria-label` tests.
3. `jest.useFakeTimers()` is enabled below (inside `beforeEach` hook) so without `{ advanceTimers: jest.advanceTimersByTime }` new keyboard tests will fail.

nrps9909 (approved):
Re-reviewed exact head `f17e42696981380cf53e4e0e4364d629d21b7764` after the feature-branch merge and follow-up fixes.

The added Cascader and TreeSelect tests now cover the default English accessible name, Chinese localization, untranslated-locale fallback, and explicit `allowClear.label` precedence. The Select style comments do not change behavior, and the current PR delta remains scoped to the clear-button accessibility work. I also checked the live review threads; none remain unresolved.

Fresh validation with the declared `@rc-component/select@1.11.0` dependency:

- Select, Cascader, and TreeSelect index suites: 3 suites, 152 passed, 1 skipped, 22 snapshots passed
- focused ESLint, Biome, and Prettier checks passed
- `git diff --check` against the merged feature parent passed
- the exact-head remote CI matrix is green

I found no blocking issue in this revision.

AI assistance disclosure: Codex was used to compare the approved and current heads, prepare an isolated exact-dependency checkout, run focused validation, audit live review threads and CI, and draft this review. I verified the cited diff, dependency version, test results, and conclusion directly before approving.

components/select/index.tsx
```diff
@@ -317,7 +320,9 @@ const InternalSelect = <
 
   const finalAllowClear = allowClear ?? contextAllowClear;
   const mergedAllowClear =
-    finalAllowClear === true ? { clearIcon: mergedClearIcon } : finalAllowClear;
+    finalAllowClear === true
+      ? { clearIcon: mergedClearIcon, label: locale.clear }
+      : finalAllowClear && { label: locale.clear, ...finalAllowClear };
```
yoyo837: **Preserve the default clear icon**

`allowClear={{ label: ... }}` omits `mergedClearIcon`, so rc-select falls back to `×`. Please keep the Ant Design default unless `clearIcon` is explicitly supplied.

yoyo837 (commented):
One issue found.

nrps9909 (changes requested):
Follow-up on exact head `d2d337cc120cf345f14daec1a155f49c7bf2ab02`: I rechecked the new unresolved maintainer finding in https://github.com/ant-design/ant-design/pull/58267#discussion_r3820897083 and agree that it is blocking. In the label-only object path, `mergedAllowClear` becomes `{ label: locale.clear, ...finalAllowClear }` and omits `mergedClearIcon`; because the standalone `clearIcon` prop has already been consumed and only `mergedAllowClear` is forwarded, rc-select falls back to `×` instead of preserving Ant Design’s CloseCircle icon. This supersedes my prior approval on the same head. Please keep `mergedClearIcon` as the object-path default while allowing an explicit `allowClear.clearIcon` to override it, and add a regression covering label-only default-icon preservation plus explicit-icon precedence. I am not adding a duplicate inline comment because the maintainer thread already captures the issue. AI assistance disclosure: Codex was used to refresh the canonical review threads, trace the exact-head merge path, and draft this follow-up; I verified the current source and head before submitting.

Pareder:
@yoyo837 Fixed and also added corresponding tests for Select/Cascader/TreeSelect components.

yoyo837 (approved):

PR #58287, opened 2026-06-09
fix(Cascader): keep parent column fixed for right-aligned placement

### LINKED ISSUE #58277: [Enhancement|UX] Cascade级联选择 placement指定为topRight或bottomRight时，展开子节点后父级选项会整体移动到左侧，交互不稳定
### Reproduction link
**处理方案**
@zombieJ 
> placement 只关注弹窗位置设定，需要右到左推移应该使用 rtl。

https://codesandbox.io/p/sandbox/ji-ben-antd-6-4-3-forked-r43qlf?file=%2Fdemo.tsx%3A66%2C32

[Edit on StackBlitz](https://stackblitz.com/edit/yi2fwsfm?file=src%2Fdemo.tsx)

### Steps to reproduce

Cascade级联选择 placement指定为topRight或bottomRight时，展开子节点后子节点默认出现在在右侧父级选项会整体移动到左侧,随着操作进行父级选项会频繁左右移动

Image

### What is expected?

父级选项位置应保持固定，保证交互的稳定性
如placement选 bottomLeft
Image

Cascade级联选择 placement指定为topRight或bottomRight时，展开子节点后父级选项位置保持固定

| Environment | Info |
| --- | --- |
| antd | 6.4.3 |
| React | 19.0.0 |
| System | MacOS 26.4 (25E246) |
| Browser |  Chrome 149.0 |

18062706139fcz:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

close #58277

### 💡 Background and Solution

**Problem:** When `Cascader` uses `placement="topRight"` / `"bottomRight"`, the popup anchors its **right edge** to the trigger (`points: ['tr', 'br']`). However the multi-column menu is a LTR horizontal `flex` layout, so expanding a child column appends the new column on the right and widens the panel. With the right edge pinned, the parent column gets pushed leftwards (sometimes out of the viewport), and a third column then triggers `adjustX` re-positioning, causing visible jitter.

**Solution:**
- For `bottomRight` / `topRight` placement, set `.ant-cascader-menus` to `flex-direction: row-reverse`, so newly expanded child columns grow leftwards while the parent column's right edge stays locked and fixed.
- Make the column divider direction-agnostic: change from `:not(:last-child)` `borderInlineEnd` to `:not(:first-child)` `borderInlineStart`, so the divider renders correctly under both normal and reversed column order.

**Verification:** Added a unit test; manually verified that the parent column's right edge stays fixed across three levels of expansion.

#### Before

The parent column (Zhejiang) sits on the left, columns grow rightwards and the panel overflows past the trigger's right edge — the parent column shifts/jitters while expanding.

#### After

Column order is reversed (West Lake | Hangzhou | Zhejiang); the parent column's right edge stays aligned with the trigger and fixed, while child columns grow leftwards.

> Note: Before/After screenshots will be attached in the PR description (UI-only diff).

### 📝 Changelog

| Language | Changelog |
| -------- | -------- |
| 🇺🇸 English | Fix Cascader parent column shifting and jitter when \`placement\` is \`topRight\` or \`bottomRight\`. |
| 🇨🇳 Chinese | 修复 Cascader 在 \`placement\` 为 \`topRight\` 或 \`bottomRight\` 时父级面板左移和抖动的问题。 |

### DIFF
--- a/components/cascader/__tests__/index.test.tsx
+++ b/components/cascader/__tests__/index.test.tsx
@@ -433,6 +433,23 @@ describe('Cascader', () => {
     expect((global as any)?.triggerProps.popupPlacement).toEqual('topRight');
   });
 
+  it('right-aligned placement reverses column order to keep parent column fixed', () => {
+    const customOptions = [
+      {
+        value: 'zhejiang',
+        label: 'Zhejiang',
+        children: [{ value: 'hangzhou', label: 'Hangzhou' }],
+      },
+    ];
+    const { container } = render(
+      <Cascader options={customOptions} placement="bottomRight" defaultValue={['zhejiang']} open />,
+    );
+
+    const dropdown = getDropdown(container);
+    expect(dropdown?.className).toContain('ant-select-dropdown-placement-bottomRight');
+    expect(container.querySelector('.ant-cascader-menus')).toBeTruthy();
+  });
+
   it('popup correctly with defaultValue RTL', () => {
     const { asFragment } = render(
       <ConfigProvider direction="rtl">
--- a/components/cascader/style/columns.ts
+++ b/components/cascader/style/columns.ts
@@ -60,8 +60,8 @@ const getColumnsStyle: GenerateStyle<CascaderToken, CSSInterpolation> = (token) 
           listStyle: 'none',
           '-ms-overflow-style': '-ms-autohiding-scrollbar', // https://github.com/ant-design/ant-design/issues/11857
 
-          '&:not(:last-child)': {
-            borderInlineEnd: `${unit(token.lineWidth)} ${token.lineType} ${token.colorSplit}`,
+          '&:not(:first-child)': {
+            borderInlineStart: `${unit(token.lineWidth)} ${token.lineType} ${token.colorSplit}`,
           },
 
           '&-item': {
--- a/components/cascader/style/columns.ts
+++ b/components/cascader/style/columns.ts
@@ -60,8 +60,8 @@ const getColumnsStyle: GenerateStyle<CascaderToken, CSSInterpolation> = (token) 
           listStyle: 'none',
           '-ms-overflow-style': '-ms-autohiding-scrollbar', // https://github.com/ant-design/ant-design/issues/11857
 
-          '&:not(:first-child)': {
-            borderInlineStart: `${unit(token.lineWidth)} ${token.lineType} ${token.colorSplit}`,
+          '&:not(:last-child)': {
+            borderInlineEnd: `${unit(token.lineWidth)} ${token.lineType} ${token.colorSplit}`,
           },
 
           '&-item': {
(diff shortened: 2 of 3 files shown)


RykerFeng:
Thanks for the review @gemini-code-assist @coderabbitai — good catch on the RTL conflict.

Confirmed: in RTL the default `placement` is `bottomRight`, and the existing `direction: rtl` already lays the columns out right-to-left, so the unconditional `row-reverse` would flip the layout a second time.

Fixed in eb167c7c68 by scoping the `row-reverse` rule to LTR only — the right-placement selectors now exclude the RTL dropdown via `:not(.ant-cascader-dropdown-rtl)`, leaving RTL layout untouched. Tests pass (including the RTL placement case).

RykerFeng:
Did a thorough browser verification and found one more issue that @gemini-code-assist hinted at — the column divider was misaligned under `row-reverse` (an extra divider on the outermost column, and a missing one between the two rightmost columns).

Fixed in b6c2d94273:
- Reverted `columns.ts` back to the upstream divider rule (`:not(:last-child)` `border-inline-end`), which is correct for both the default LTR layout and RTL.
- Overrode the divider only inside the LTR `row-reverse` block, so dividers stay between adjacent columns when the visual order is reversed.

Verified in the browser:
- **LTR `bottomRight`**: parent column right edge fixed; dividers now sit correctly between West Lake | Hangzhou | Zhejiang.
- **LTR `bottomLeft`** (default): `flex-direction: row`, divider unchanged — default behavior untouched.
- **RTL**: `columns.ts` is now identical to upstream and the `row-reverse` rule is excluded via `:not(.ant-cascader-dropdown-rtl)`, so RTL keeps its original behavior.

Tests pass (64 passed, 17 snapshots unchanged).

QDyanbing:
image
现在这种是完全错误的交互；这里原本就没有问题

murisans:
> Did a thorough browser verification and found one more issue that @gemini-code-assist hinted at — the column divider was misaligned under `row-reverse` (an extra divider on the outermost column, and a missing one between the two rightmost columns).
> 
> Fixed in [b6c2d94](https://github.com/ant-design/ant-design/commit/b6c2d94273f08f890726b0bb3ac49c72dbc48642):
> 
> * Reverted `columns.ts` back to the upstream divider rule (`:not(:last-child)` `border-inline-end`), which is correct for both the default LTR layout and RTL.
> * Overrode the divider only inside the LTR `row-reverse` block, so dividers stay between adjacent columns when the visual order is reversed.
> 
> Verified in the browser:
> 
> * **LTR `bottomRight`**: parent column right edge fixed; dividers now sit correctly between West Lake | Hangzhou | Zhejiang.
> * **LTR `bottomLeft`** (default): `flex-direction: row`, divider unchanged — default behavior untouched.
> * **RTL**: `columns.ts` is now identical to upstream and the `row-reverse` rule is excluded via `:not(.ant-cascader-dropdown-rtl)`, so RTL keeps its original behavior.
> 
> Tests pass (64 passed, 17 snapshots unchanged).


@zombieJ 给了答复 issue 关了哈 ，用rtl规则 去做

PR #58288, opened 2026-06-09
fix(Select): follow-up refine single-mode open-state labelRender dimming

kfylaktopoulos:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues
ref #57951
follow-up to #57954

### 💡 Background and Solution
**Background:**
The search-state bug in #57951 (non-text labelRender content staying visible while typing) was already fixed by #57954.
This PR is a follow-up focused only on open-state visual consistency in Select single mode.

Problem in current behavior
When dropdown is open (before typing), plain text labels are dimmed by placeholder color.
But if labelRender wraps text in an element (for example, span + icon/text), the current rule also applies child opacity 0.25, so wrapped text becomes much dimmer than plain text.

https://codesandbox.io/p/sandbox/ji-ben-shi-yong-antd-6-4-3-forked-tx65jq?file=%2Fdemo.tsx%3A17%2C23

**Solution:**
Refine the open-state CSS so custom labelRender text does not get double-dimmed:
- keep text dimming behavior consistent with plain text labels
- avoid stacking parent placeholder color with child opacity on wrapped text
- preserve intended dimming for non-text rich content (such as image/icon visuals) where needed
- keep search input text and caret fully visible
- no extra DOM wrappers introduced

### 📝 Change Log

| Language | Changelog |
| -------- | --------- |
| 🇺🇸 English |  🐞 Refine Select single-mode open-state dimming for custom labelRender content to avoid over-dimming wrapped text while keeping rich-content visibility behavior consistent. |
| 🇨🇳 Chinese |     -      |

### DIFF
--- a/components/select/style/select-input.ts
+++ b/components/select/style/select-input.ts
@@ -308,7 +308,7 @@ const genSelectInputStyle: GenerateStyle<SelectToken, CSSObject> = (token) => {
             '&-has-search-value': {
               color: 'transparent',
 
-              [`> :not(${componentCls}-input)`]: {
+              [`> *:not(${componentCls}-input)`]: {
                 opacity: 0,
               },
             },
@@ -317,14 +317,24 @@ const genSelectInputStyle: GenerateStyle<SelectToken, CSSObject> = (token) => {
             '&-value': {
               transition: `all ${token.motionDurationMid} ${token.motionEaseInOut}`,
               zIndex: 1,
+              opacity: 1,
             },
           },
 
+          // Dim the selected content while the dropdown is open. Shared by all select-like
+          // components (Select / Cascader / TreeSelect) since they render through the same
+          // `content` structure.
           [`&${componentCls}-open ${componentCls}-content`]: {
-            color: token.colorTextPlaceholder,
+            '&-has-value': {
+              opacity: 0.25,
+            },
 
             '&-has-search-value': {
+              opacity: 1,
               color: 'transparent',
+              [`> *:not(${componentCls}-input)`]: {
+                opacity: 0,
+              },
             },
           },
         },
--- a/components/select/style/select-input.ts
+++ b/components/select/style/select-input.ts
@@ -331,6 +331,7 @@ const genSelectInputStyle: GenerateStyle<SelectToken, CSSObject> = (token) => {
 
             '&-has-search-value': {
               opacity: 1,
+              transition: `opacity ${token.motionDurationMid} ${token.motionEaseInOut}`,
               color: 'transparent',
               [`> *:not(${componentCls}-input)`]: {
                 opacity: 0,


QDyanbing:
## 问题背景

- select 单选可搜索模式下，如果选项自带了图片等内容；
- 当选择了一个 item 后图片会回显在 select 内
- 再次输入文字去搜索时只把文字的 color 变为透明了，上一个 pr 已解决了此问题 opacity:0 使内容隐藏掉

当前的 pr 应该解决的是当下拉菜单被展开时；原已选中的 item 的文字透明度降低变淡的效果，图片没有跟着一起变淡的问题； @zombieJ 给看看吧

components/select/style/select-input.ts
```diff
@@ -327,6 +328,21 @@ const genSelectInputStyle: GenerateStyle<SelectToken, CSSObject> = (token) => {
               color: 'transparent',
             },
           },
+
+          [`&:not(${antCls}-cascader):not(${antCls}-tree-select)${componentCls}-open ${componentCls}-content`]:
```
zombieJ: Should we both benift Cascader & TreeSelect also? Its strcutrue is design for all the select-like component.

kfylaktopoulos: Makes total sense! Since the structure is shared, extending this fix to Cascader and TreeSelect is a great idea. Let me test it with those components as well and update the PR shortly. 👍

kfylaktopoulos: @zombieJ Ι have updated the PR. I removed the exclusions for `Cascader` and `TreeSelect` and tested all three components locally. The open-state opacity transitions now work perfectly for custom content across the board! Ready for another look whenever you have a chance.

kfylaktopoulos:
Hi @zombieJ, just a gentle ping to let you know that I've pushed the requested commit. Feel free to take a look whenever you have a chance. If there's anything else needed, I'm at your disposal!

afc163:
Thanks for following up on this.

I checked #57951 again, and the original search-state issue where non-text `labelRender` content remained visible while typing was already fixed by #57954. So I think this PR is better framed as a follow-up for the open-state dimming behavior, rather than the primary fix for #57951. Could you update the PR description from `close #57951` to something like `ref #57951` / `follow-up to #57954`?

The pure CSS direction is much better than adding an extra wrapper, since it avoids changing the rendered label value shape. The remaining concern is the open state rule:

```ts
'&-has-value': {
  [`> *:not(${componentCls}-input)`]: {
    opacity: 0.25,
  },
},
```

For rich labels that wrap text in an element, e.g. `<span>United States</span>`, this combines the parent `colorTextPlaceholder` with child `opacity: 0.25`, making the text much dimmer than a plain text label. Please take another look at that case and make sure custom text content does not become over-dimmed when the dropdown is open but the user has not started searching.

kfylaktopoulos:
> Thanks for following up on this.
> 
> I checked #57951 again, and the original search-state issue where non-text `labelRender` content remained visible while typing was already fixed by #57954. So I think this PR is better framed as a follow-up for the open-state dimming behavior, rather than the primary fix for #57951. Could you update the PR description from `close #57951` to something like `ref #57951` / `follow-up to #57954`?
> 
> The pure CSS direction is much better than adding an extra wrapper, since it avoids changing the rendered label value shape. The remaining concern is the open state rule:
> 
> ```ts
> '&-has-value': {
>   [`> *:not(${componentCls}-input)`]: {
>     opacity: 0.25,
>   },
> },
> ```
> 
> For rich labels that wrap text in an element, e.g. `<span>United States</span>`, this combines the parent `colorTextPlaceholder` with child `opacity: 0.25`, making the text much dimmer than a plain text label. Please take another look at that case and make sure custom text content does not become over-dimmed when the dropdown is open but the user has not started searching.

Thanks @afc163! Done:

- Updated the related issues to ref #57951 / follow-up to #57954
- Replaced the per-child `opacity: 0.25` rule with a single `opacity: 0.25` wrapper itself, so plain text, element-wrapped text (<span>text here</span>), and rich media (images/icons) all dim uniformly — no `color` + `opacity` stacking.

components/select/style/select-input.ts
```diff
@@ -317,14 +317,24 @@ const genSelectInputStyle: GenerateStyle<SelectToken, CSSObject> = (token) => {
...
+          // components (Select / Cascader / TreeSelect) since they render through the same
+          // `content` structure.
           [`&${componentCls}-open ${componentCls}-content`]: {
-            color: token.colorTextPlaceholder,
+            '&-has-value': {
+              opacity: 0.25,
+            },
 
             '&-has-search-value': {
+              opacity: 1,
```
afc163, line 333:
```suggestion
              opacity: 1,
              transition: `opacity ${token.motionDurationMid} ${token.motionEaseInOut}`,
```

kfylaktopoulos: Good catch! Thanks for adding the transition, I completely missed that.

afc163 (approved):

PR #58371, opened 2026-06-14
fix: support Pagination quick jumper variant

biubiukam:
### This is a ...

- [ ] New feature
- [x] Bug fix
- [ ] Site / documentation improvement
- [ ] Demo improvement
- [x] Component style improvement
- [ ] TypeScript definition improvement
- [ ] Bundle size optimization
- [ ] Performance optimization
- [ ] Feature enhancement
- [ ] Internationalization
- [ ] Refactoring
- [ ] Code style optimization
- [x] Test Case
- [ ] Branch merge
- [ ] Workflow
- [ ] Accessibility improvement
- [ ] Other

### Related Issues

fix #49769

### Background and Solution

Pagination quick jumper uses a native input rendered by rc-pagination, so it did not receive Ant Design input variant styles from `ConfigProvider variant="filled"`. This made the quick jumper input stay outlined while the size changer Select followed the configured variant.

This change reads the merged input variant inside Pagination, adds a corresponding Pagination variant class for non-outlined variants, and applies quick jumper input styles for filled, borderless, and underlined modes. The default outlined rendering remains unchanged.

### Change Log

| Language | Changelog |
| --- | --- |
| English | Fix Pagination quick jumper input not following ConfigProvider variant styles. |
| Chinese | 修复 Pagination 快速跳转输入框未跟随 ConfigProvider variant 样式的问题。 |

### Checks

- `npm test -- components/pagination/__tests__/index.test.tsx --runInBand`
- `npm run tsc -- --pretty false`
- `npx biome lint components/pagination/Pagination.tsx components/pagination/style/index.ts components/pagination/__tests__/index.test.tsx`
- `npx eslint components/pagination/Pagination.tsx components/pagination/style/index.ts components/pagination/__tests__/index.test.tsx --cache`

PR #58385, opened 2026-06-15
fix: handle empty children in Space.Compact

### LINKED ISSUE #58191: Space.Compact not applying classes correctly when there is more than one child but only one that actually renders anything.
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/devbox/button-compact-mode-antd-6-4-3-forked-9rpdcc?workspaceId=ws_9tFAgLaTRk4Mjk8XGS1R2h)

### Steps to reproduce

Have a set of children for a Space.Compact component that are conditionally rendered. If only one child actually results in a component being rendered then no ant-btn-compact-last-item class gets applied.

### What is expected?

Both first and last item classes are applied leading to rounding to the whole button

This results in loss of rounding on the right side of the single button, only the left.

| Environment | Info |
| --- | --- |
| antd | 6.4.3 |
| React | 19 |
| System | Windows |
| Browser | Chrome |

---
I had opened this originally as (https://github.com/ant-design/ant-design/issues/55229) but I hadn't made the repro sandbox public. I've done so this time around though, and updated it to latest versions as well.

biubiukam:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

- Fix #58191

### 💡 Background and Solution

`Space.Compact` assigned first/last item context from the original React children list. When a child component existed in that list but rendered `null`, the visible compact item could miss the `*-compact-last-item` or `*-compact-first-item` class.

This PR lets rendered compact-aware children register themselves with the nearest `Space.Compact`. The parent keeps the source order as the fallback, then uses the registered visible item ids to mark the actual first and last compact items. A regression test covers hidden children before and after the visible button.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Space.Compact not applying both first and last item styles when sibling children render nothing. |
| 🇨🇳 Chinese | 🐞 修复 Space.Compact 在兄弟子节点渲染为空时未同时应用首尾项样式的问题。 |

### DIFF
--- a/components/space/Compact.tsx
+++ b/components/space/Compact.tsx
@@ -184,10 +184,6 @@ const Compact: React.FC<SpaceCompactProps> = (props) => {
 
     if (count === 0) {
       setRegisteredItemIds((prevIds) => {
-        if (prevIds.has(itemId)) {
-          return prevIds;
-        }
-
         const nextIds = new Set(prevIds);
         nextIds.add(itemId);
         return nextIds;
@@ -200,10 +196,6 @@ const Compact: React.FC<SpaceCompactProps> = (props) => {
       if (latestCount <= 1) {
         registeredItemCountRef.current.delete(itemId);
         setRegisteredItemIds((prevIds) => {
-          if (!prevIds.has(itemId)) {
-            return prevIds;
-          }
-
           const nextIds = new Set(prevIds);
           nextIds.delete(itemId);
           return nextIds;
--- a/components/space/Compact.tsx
+++ b/components/space/Compact.tsx
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import { toArray } from '@rc-component/util';
+import { toArray, useLayoutEffect } from '@rc-component/util';
 import { clsx } from 'clsx';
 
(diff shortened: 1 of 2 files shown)


biubiukam:
> [!IMPORTANT]
> - [x] Visual diff is acceptable

The reported added `pagination-variant-debug.*` snapshots come from the current upstream `master` commit `ae84500a2` (#58371), not from this PR's `Space.Compact` changes. Locally, `git diff upstream/master...HEAD --name-only` only lists `components/space/Compact.tsx` and `components/space/__tests__/space-compact.test.tsx`, so this visual diff is acceptable.

QDyanbing:
Someone is already handling this issue.

PR #58414, opened 2026-06-17
fix(Badge): improve processing status in dark mode

### LINKED ISSUE #55234: Badge component with status="processing" in dark mode

Badge component makes a flashing animation in "processing" status which is visible in light mode but not so in dark mode. It will be a nice addition if it can be made as noticeable in the dark mode as well.

Api will not be changed.

clxstart:
### This is a

- [x] Bug fix
- [x] Component style improvement

### Related Issues

Fixes #55234

### Background and Solution

`Badge` with `status=processing` used the same info color for the status dot and the animated ring. In dark mode this makes the ring hard to notice against the dark background.

This updates the processing status color token to `colorInfoTextHover`, while keeping the dot background as `colorInfo`. The pseudo-element ring continues to inherit the status color through `currentcolor`, so the animation becomes more visible without changing the public API.

### Test Plan

- `npx eslint components/badge/style/index.ts`

### Change Log

| Language | Changelog |
| --- | --- |
| English | Fix Badge processing status ring being hard to notice in dark mode. |
| Chinese | 修复 Badge processing 状态动画圈在暗色模式下不明显的问题。 |


clxstart:
The visual diffs in `badge-mix.dark.png` and `badge-mix.default.png` are expected for this PR.

This change intentionally updates the `processing` badge ring color from the same `colorInfo` token used by the dot to the stronger `colorInfoTextHover` token. The dot background remains `colorInfo`, while the animated ring inherits the stronger color through `currentcolor`, making the processing animation more noticeable in dark mode without changing the public API.

components/badge/__tests__/index.test.tsx
```diff
@@ -151,6 +154,28 @@ describe('Badge', () => {
...
+    );
+
+    const styleText = extractStyle(cache, { plain: true });
+
+    expect(styleText).toContain(
+      '.ant-badge.ant-badge-status .ant-badge-status-processing{overflow:visible;color:var(--ant-color-info-text-hover);background-color:var(--ant-color-info);border-color:currentcolor;',
+    );
+    expect(styleText).toContain(`--ant-color-info-text-hover:${darkToken.colorInfoTextHover};`);
+    expect(styleText).toContain(`--ant-color-info:${darkToken.colorInfo};`);
+  });
```
afc163: 不要自证 A===A

clxstart: Removed the token value self-comparison. The test now only asserts the generated Badge processing CSS uses the stronger color variable for the ring while keeping the dot background on the info variable.

afc163: I believe this test case could be removed, cause we have visual diff test.

clxstart: Removed this test case. The visual diff test now covers this expected Badge processing status color change.

afc163 (approved):

PR #58417, opened 2026-06-17
fix: avoid Form warning hook compiler memoization

biubiukam:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

- fix #56021.

### 💡 Background and Solution

Form's development warning hook calls `devUseWarning`, which reads React context. In the UMD development build with React Compiler enabled, that call can be memoized behind the compiler cache sentinel while `useEffect` still runs on every render. React 19 then reports a hook order mismatch when Form rerenders.

This opts `useFormWarning` out of React Compiler memoization so the warning hook call keeps the same order across renders. A compiler-level regression test verifies that the generated output no longer contains the compiler cache sentinel for this hook.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Form hook order warning in the UMD development build with React 19. |
| 🇨🇳 Chinese | 修复 Form 在 React 19 开发环境使用 UMD 构建时触发 hooks 顺序警告的问题。 |

### ✅ Checks

- `npm test -- --runTestsByPath components/form/__tests__/compiler.test.ts components/form/__tests__/index.test.tsx --runInBand`
- `npm run lint`


afc163:
Thanks for working on this. The direction looks right, but I do not think we should merge the current shape yet.

The hook order issue is caused by the warning helper being a real hook while its name is `devUseWarning`, so React Compiler treats the call as a normal function call and can cache it behind `react.memo_cache_sentinel`. I verified locally with the current Babel compiler config: when the call is recognized as `useDevWarning`, `useFormWarning` no longer wraps the warning hook call in the sentinel branch, while `Form.Item` can still keep the rest of its compiler memoization.

So instead of adding `'use no memo'` to the whole `Form.Item` / `Form.List` / status hook functions, could we narrow the fix to the warning hook itself? For example, expose or import it as `useDevWarning` (or another `useXxx` name) so React Compiler recognizes it as a hook, then remove the broad `use no memo` directives from these Form paths.

The test should also avoid asserting that the whole compiled file has no `react.memo_cache_sentinel`, since that locks us into disabling useful compiler optimizations. It would be better to assert that the warning hook call is not cached/skipped, or otherwise cover the hook-order regression directly.

The current CI failures I saw look unrelated to this change: `test-node` and the visual snapshot job both failed during `ut` dependency resolution with npm registry `HTTP 406` for `@babel/plugin-syntax-optional-chaining`; the build/lint/dom/dist checks passed.

afc163 (approved):
The current patch is appropriately scoped to the warning-hook rename, keeps the compatibility alias, updates the Form usage, and all required checks are green. Approved.

thinkasany (approved):

components/form/__tests__/compiler.test.ts
```diff
@@ -0,0 +1,46 @@
+import fs from 'node:fs';
```
li-jia-nan, line 1: 感觉这个测试用例比较奇怪，好像跟业务无关，不应该出现在这里

li-jia-nan: - #58917

PR #58451, opened 2026-06-21
fix: keep Table sticky header top border in bordered mode

BangDori:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

Fixes #56708

### 💡 Background and Solution

In bordered Table, the top border is applied to `.ant-table-container`. When the header becomes sticky, the container border scrolls away while the sticky header remains fixed, so the top border appears missing.

This PR adds a top border to the sticky header holder in the bordered Table case, and applies a negative top margin by one line width to avoid showing a double border before the header becomes sticky.

AS-IS

image

TO-BE

image

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Table missing top border when bordered sticky header becomes fixed |
| 🇨🇳 Chinese | 🐞 修复 Table 在边框模式下吸顶表头固定后顶部边框丢失的问题。 |

### DIFF
--- a/components/table/__tests__/Table.test.tsx
+++ b/components/table/__tests__/Table.test.tsx
@@ -1,4 +1,5 @@
 import React, { useRef, useState } from 'react';
+import { createCache, extractStyle, StyleProvider } from '@ant-design/cssinjs';
 
 import type { TableProps, TableRef } from '..';
 import Table from '..';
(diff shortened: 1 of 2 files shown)


components/table/__tests__/Table.test.tsx
```diff
@@ -20,6 +21,30 @@ describe('Table', () => {
...
+    );
+
+    const stickyHeaderSelector =
+      '.ant-table.ant-table-bordered >.ant-table-container >.ant-table-header.ant-table-sticky-holder';
+    const styleText = extractStyle(cache, { plain: true });
+
+    expect(styleText).toContain(
+      `${stickyHeaderSelector}{margin-top:calc(var(--ant-line-width) * -1);border-top:var(--ant-line-width) var(--ant-line-type) var(--ant-table-border-color);}`,
+    );
+  });
```
afc163: We can drop this test case.

BangDori: Agreed. This test only checks the generated CSS output, so it doesn’t add much value beyond the style change itself. I’ve removed it as suggested.

Thanks!

[test(Table): remove sticky border style assertion](https://github.com/ant-design/ant-design/pull/58451/commits/b2af7dd79dad4e430009f28d1b5cffce47ee5e7f)

components/table/style/bordered.ts
```diff
@@ -50,6 +50,11 @@ const genBorderedStyle: GenerateStyle<TableToken, CSSObject> = (token) => {
           borderInlineStart: tableBorder,
           borderTop: tableBorder,
 
+          [`> ${componentCls}-header${componentCls}-sticky-holder`]: {
+            marginTop: unit(calc(lineWidth).mul(-1).equal()),
```
li-jia-nan:
```suggestion
            marginTop: calc(lineWidth).mul(-1).equal(),
```

BangDori: Updated as suggested. Thanks!

[style(Table): simplify sticky header margin calculation](https://github.com/ant-design/ant-design/pull/58451/commits/91650e3c47a338f745806a546048d73b537531af)

afc163 (approved):

PR #58452, opened 2026-06-21
fix(alert): align icon to top instead of center for multi-line content

### LINKED ISSUE #57788: Alert icon vertical align should not be centered
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/sandbox/tu-biao-antd-6-3-6-forked-pmmpr6?file=%2Fdemo.tsx%3A8%2C533)

### Steps to reproduce

See the render result

### What is expected?

icon should stick to top

Image

icon is centered

Image

| Environment | Info |
| --- | --- |
| antd | 6.3.6 |
| React | 19 |
| System | macOS |
| Browser | Chrome 145 |

creazyfrog:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

Fixes #57788

### 💡 Background and Solution

When an Alert contains multi-line content (a long title that wraps, or a description), the root container uses `alignItems: 'center'`, which vertically centers the icon against the full content height. This causes the icon to float in the middle instead of anchoring to the top.

**Fix:** Change `alignItems: 'center'` to `alignItems: 'flex-start'` on the base Alert container.

- For single-line alerts the visual result is identical (flex-start and center are the same when there's no overflow height).
- For multi-line alerts, the icon now correctly anchors to the top of the content.
- The `alert-with-description` variant already used `alignItems: 'flex-start'`, so this brings the base style in line with the description variant.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix(Alert): icon now aligns to the top when alert content spans multiple lines |
| 🇨🇳 Chinese | 修复(Alert): 多行内容时图标对齐方式改为顶部对齐 |

🤖 Generated with [Claude Code](https://claude.ai/code)


QDyanbing:
There is already an earlier PR #57878 under review for the same issue。

creazyfrog:
Thank you for the review! Both concerns are valid — changing  to  globally would top-align the icon and close button in single-line alerts.

Updated the fix with the following approach:

1. Keep  on the root container (needed for multi-line title wrapping).
2. Add  to **`.alert-icon`** — this offsets the icon downward so it visually centers with the first line of text. For single-line alerts this equals centering; for multi-line it anchors to the first line.
3. Same formula for **`.alert-close-icon`**: `(fontSize * lineHeight - fontSizeIcon) / 2`.
4. Reset `marginBlockStart: 0` in the **`with-description`** override where the larger icon should sit flush at the top.

The commit has been amended and force-pushed to the branch.

PR #58454, opened 2026-06-21
fix(steps): remove subpixel translateY on horizontal rail to eliminate blurry rendering

### LINKED ISSUE #58112: Steps: Rail line appears 2px thick with washed-out color due to subpixel rendering from translateY(-50%)

creazyfrog:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

Fixes #58112

### 💡 Background and Solution

The `.ant-steps-item-rail` element in horizontal steps uses `transform: translateY(-50%)` alongside `marginTop` to center the 1px rail line on the icon's vertical midpoint.

Root cause: When the parent has an odd pixel height, `translateY(-50%)` resolves to a half-pixel offset. The browser distributes the 1px border across two physical pixels via anti-aliasing, making the rail appear ~2px thick and its color desaturated.

Fix: Remove `transform: translateY(-50%)`. The existing `marginTop: varRef('horizontal-rail-margin')` (= `icon-size-max / 2 + item-wrapper-padding-top`) already positions the rail top at the icon center. Removing the 0.5px correction eliminates the subpixel offset while keeping alignment correct — the rail renders as a crisp 1px line with accurate color.

This bug is visible on the ant.design documentation site in the Steps component examples.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix(Steps): horizontal rail line now renders as a crisp 1px line with correct color |
| 🇨🇳 Chinese | 修复(Steps): 水平步骤条连接线现在渲染为清晰的 1px 线条，颜色正确 |

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/components/steps/style/horizontal.ts
+++ b/components/steps/style/horizontal.ts
@@ -23,7 +23,6 @@ const genHorizontalStyle: GenerateStyle<StepsToken, CSSObject> = (token) => {
           flex: 1,
           minWidth: 0,
           alignSelf: 'flex-start',
-          transform: 'translateY(-50%)',
         },
       },
     },


components/steps/style/horizontal.ts
```diff
@@ -23,7 +23,6 @@ const genHorizontalStyle: GenerateStyle<StepsToken, CSSObject> = (token) => {
           flex: 1,
           minWidth: 0,
           alignSelf: 'flex-start',
-          transform: 'translateY(-50%)',
```
QDyanbing, line 26: Removing it directly may affect the centering layout. https://github.com/ant-design/ant-design/pull/58149/changes

afc163:
Thanks for the contribution. Closing this as a duplicate of #58149 for the same #58112 Steps rail issue, so we can keep the remaining review and test discussion in one PR.

PR #58477, opened 2026-06-23
fix: add rate star border tokens

pupuking723:
## Summary

- add Rate component tokens for default and selected star border color/width
- apply the border tokens to default and selected star SVGs as a redundant visual cue for non-text contrast
- add a style extraction regression test for the new token output

Fixes #55225.

## Verification

- npm test -- components/rate/__tests__/index.test.tsx --runInBand
- npm test -- components/rate --runInBand
- npm exec eslint components/rate/style/index.ts components/rate/__tests__/index.test.tsx
- npm run tsc
- npm run token:meta
- npm run token:statistic

PR #58501, opened 2026-06-24
fix: add rate star border tokens

### LINKED ISSUE #55225: Rate component: add redundant border styling to meet non-text contrast requirements
### Reproduction link

[https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html#:~:text=Relationship%20with%20Use%20of%20Color](https://www.w3.org/WAI/WCAG22/Understanding/non-text-contrast.html#:~:text=Relationship%20with%20Use%20of%20Color)

### Steps to reproduce

Place the Rate component on a page where the page background is white
Toggle between Default and Selected states (e.g., select/unselect a star).
Observe that the only visual difference between states is the fill color (no border or other redundant cue).
Use a contrast checker to verify that Selected vs Default or either state vs page background falls below 3:1.

### What is expected?

The component provides a redundant visual cue (for example: border color and width) in addition to fill color, so:
Selected vs Default has at least 3:1 contrast or provides redundant visual cuew
Selected vs page background and Default vs page background each have at least 3:1 contrast

The Rate component currently relies only on graphical fill to indicate Selected vs Default states. This can fail WCAG non-text contrast requirements. To be accessible, the component must provide a visible, high-contrast cue between its mandatory states (Selected and Default) and versus the page background (minimum 3:1 contrast). Add explicit border styling variables so the component can show a visible border (color and width) separately for Selected and Default states, allowing implementers to meet contrast requirements even when fill color alone is insufficient.

| Environment | Info |
| --- | --- |
| antd | undefined |
| React | React |
| System | OSX, Windows, Linux, |
| Browser | Any browser |

---
Suggested fix / implementation details: Add four new component-level variables to enable independent border styling for the two mandatory states. These should be configurable without changing the fill:
star-border-color-default
star-border-width-default
star-border-color-selected
star-border-width-selected

pupuking723:
## Summary

- add Rate component tokens for default and selected star border color/width
- apply the border tokens to default and selected star SVGs as a redundant visual cue for non-text contrast
- add a style extraction regression test for the new token output

Fixes #55225.

This supersedes #58477, which was closed by automation after the base branch was still `master`. This PR targets `feature` as requested.

## Verification

- npm test -- components/rate/__tests__/index.test.tsx --runInBand
- npm exec eslint components/rate/style/index.ts components/rate/__tests__/index.test.tsx

### DIFF
--- a/components/rate/__tests__/index.test.tsx
+++ b/components/rate/__tests__/index.test.tsx
@@ -1,12 +1,10 @@
 import React from 'react';
-import { createCache, extractStyle, StyleProvider } from '@ant-design/cssinjs';
 import { render } from '@testing-library/react';
 
 import Rate from '..';
 import focusTest from '../../../tests/shared/focusTest';
 import mountTest from '../../../tests/shared/mountTest';
 import rtlTest from '../../../tests/shared/rtlTest';
-import ConfigProvider from '../../config-provider';
 
 describe('Rate', () => {
   focusTest(Rate, { refFocus: true });
@@ -22,40 +20,5 @@ describe('Rate', () => {
       const { container } = render(<Rate count={3} value={1} size="large" />);
       expect(container.querySelector('.ant-rate-large')).toBeTruthy();
     });
-  });
-
-  it('supports border tokens for default and selected stars', () => {
-    const cache = createCache();
-
-    render(
-      <StyleProvider cache={cache}>
-        <ConfigProvider
-          theme={{
-            components: {
-              Rate: {
-                starBorderColorDefault: '#111111',
-                starBorderWidthDefault: 2,
-                starBorderColorSelected: '#222222',
-                starBorderWidthSelected: 3,
-              },
-            },
-          }}
-        >
-          <Rate value={1} />
-        </ConfigProvider>
-      </StyleProvider>,
-    );
-
-    const styleText = extractStyle(cache, { plain: true });
-
-    expect(styleText).toContain('--ant-rate-star-border-color-default:#111111');
-    expect(styleText).toContain('--ant-rate-star-border-width-default:2px');
-    expect(styleText).toContain('--ant-rate-star-border-color-selected:#222222');
-    expect(styleText).toContain('--ant-rate-star-border-width-selected:3px');
-    expect(styleText).toContain('stroke:var(--ant-rate-star-border-color-default)');
-    expect(styleText).toContain('stroke-width:var(--ant-rate-star-border-width-default)');
-    expect(styleText).toContain('stroke-linejoin:round');
-    expect(styleText).toContain('stroke:var(--ant-rate-star-border-color-selected)');
-    expect(styleText).toContain('stroke-width:var(--ant-rate-star-border-width-selected)');
   });
 });
--- a/components/rate/__tests__/index.test.tsx
+++ b/components/rate/__tests__/index.test.tsx
@@ -1,10 +1,12 @@
 import React from 'react';
+import { createCache, extractStyle, StyleProvider } from '@ant-design/cssinjs';
 import { render } from '@testing-library/react';
 
 import Rate from '..';
 import focusTest from '../../../tests/shared/focusTest';
 import mountTest from '../../../tests/shared/mountTest';
 import rtlTest from '../../../tests/shared/rtlTest';
+import ConfigProvider from '../../config-provider';
 
 describe('Rate', () => {
   focusTest(Rate, { refFocus: true });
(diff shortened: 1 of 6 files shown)


components/rate/__tests__/index.test.tsx
```diff
@@ -21,4 +23,38 @@ describe('Rate', () => {
...
+
+    expect(styleText).toContain('--ant-rate-star-border-color-default:#111111');
+    expect(styleText).toContain('--ant-rate-star-border-width-default:2px');
+    expect(styleText).toContain('--ant-rate-star-border-color-selected:#222222');
+    expect(styleText).toContain('--ant-rate-star-border-width-selected:3px');
+    expect(styleText).toContain('stroke:var(--ant-rate-star-border-color-default)');
+    expect(styleText).toContain('stroke-width:var(--ant-rate-star-border-width-default)');
+    expect(styleText).toContain('stroke:var(--ant-rate-star-border-color-selected)');
+    expect(styleText).toContain('stroke-width:var(--ant-rate-star-border-width-selected)');
+  });
```
afc163: 无需测试用例。

components/rate/style/index.ts
```diff
@@ -36,6 +36,26 @@ export interface ComponentToken {
    * @descEN Star background color
    */
   starBg: string;
+  /**
+   * @desc 默认星星边框颜色
+   * @descEN Default star border color
+   */
+  starBorderColorDefault: string;
```
afc163:
```suggestion
  starBorderColor: string;
```

afc163: ping

pupuking723:
Updated based on the review feedback:

- Removed the added Rate test case.
- Renamed the default border color token to `starBorderColor`.

Verification:
- `npx jest --config .jest.js --no-cache components/rate/__tests__/index.test.tsx`
- `npx eslint components/rate/style/index.ts components/rate/__tests__/index.test.tsx --cache=false`

QDyanbing:
Add usage instructions to the component-token demo and cover the scenario involving character tokens.

pupuking723:
Updated the Rate component-token demo with the new border tokens and a custom `character` case.

Verification:
- `npx eslint components/rate/style/index.ts components/rate/demo/component-token.tsx --cache=false`
- `npx jest --config .jest.js --no-cache components/rate/__tests__/demo.test.ts components/rate/__tests__/demo-extend.test.ts --runInBand`

QDyanbing:
character="A" character="好" not svg?

pupuking723:
Superseded by the follow-up comment below; the original text here lost formatting when posted.

pupuking723:
Correction: my previous comment lost the inline code formatting.

The demo now uses text characters "A" and "好" instead of an SVG. I also added text-stroke styling so the new border tokens apply to text characters too.

Checked locally:
- eslint for components/rate/style/index.ts and components/rate/demo/component-token.tsx
- the rate demo and demo-extend Jest tests
- git diff --check

afc163 (approved):

li-jia-nan:
I understand the motivation for improving the default Rate contrast through component tokens. However, I’m a bit concerned about applying the SVG stroke styles to `> svg`.

Since `Rate` supports custom `character`, users may pass their own SVG icon with fully customized fill/stroke styles. In that case, the component token styles would override the user-provided SVG unexpectedly.

Should we limit the SVG stroke styles to the built-in star icon, for example targeting the internal Ant icon structure, or only apply these styles when the default character is used? Custom characters should probably remain fully controlled by users.

afc163:
I agree with this concern. `Rate` already supports `character` as a render function, so users who need custom SVG/icon styling can fully control the returned node themselves.

For the component token side, we should avoid styling arbitrary custom `character` internals, otherwise user-provided SVG `fill`/`stroke` can be unexpectedly overridden. The border token should only cover the built-in Rate star accessibility case; custom characters should remain user-controlled.

afc163:
Follow-up: I opened #58535 to revert this before 6.5.0.

This is not a rejection of the accessibility requirement in #55225. The concern is that the current token implementation exposes new public Rate tokens while also applying styles to custom `character` internals. Since `character` already supports a render function, users can fully control custom SVG/icon/text styling from the returned node, and component tokens should not unexpectedly override those custom `fill` / `stroke` styles.

We can revisit this with a narrower design, for example limiting the styling to the built-in star only, or another approach that improves the default Rate contrast without affecting custom `character` content.

afc163:
Here is an example that can achieve a similar effect without adding component tokens. The key point is to keep the styling scoped to the custom character node returned by `character`, instead of styling every SVG inside `Rate` globally:

```tsx
import React from 'react';
import { Rate } from 'antd';
import './index.css';

export default () => (
  <Rate
    className="custom-stroked-rate"
    allowHalf
    defaultValue={2.5}
    character={({ index = 0 }) => (
      <span className="custom-stroked-rate-character">{index % 2 ? 'A' : '好'}</span>
    )}
  />
);
```

```css
.custom-stroked-rate .custom-stroked-rate-character {
  -webkit-text-stroke: 2px green;
}

.custom-stroked-rate .ant-rate-star-full .ant-rate-star-second .custom-stroked-rate-character,
.custom-stroked-rate .ant-rate-star-half .ant-rate-star-first .custom-stroked-rate-character {
  -webkit-text-stroke-color: purple;
}
```

For custom SVG icons, the same idea works by returning your own SVG from `character` and styling only that SVG or its own class. This keeps custom `character` content fully controlled by users, while still allowing the same visual result when needed.

PR #58551, opened 2026-06-28
fix(Typography): inherit font size in editable textarea

### LINKED ISSUE #56626: `Typography.Title` input text in `editable` mode doesn't follow the customized design token
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/devbox/objective-pond-6jhy3r)

Image

Image

### Steps to reproduce

- Set `theme.components.Typography.fontSizeHeading1` on `ConfigProvider`
- Put editable `Title` on it
- Click the edit icon, the edit input will popped up. But the text size doesn't match the customized font size!

### What is expected?

The font size of editable input text must follow the size of customized title.

The font size of editable input text doesn't match the size of customized title.

| Environment | Info |
| --- | --- |
| antd | 6.2.0 |
| React | 19.2.3 |
| System | Ubuntu 24.04 |
| Browser | Chrome 143 |

gaurav0107:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

close #56626

### 💡 Background and Solution

When a `Typography.Title` (or `Text`/`Paragraph`) enters editable mode, the
editing `<textarea>` rendered the text at the user-agent form-control font size
instead of the size of the element being edited. So a customized heading token
(for example `theme.components.Typography.fontSizeHeading1`) was correctly
applied to the displayed title but not to the edit input, breaking the WYSIWYG
editing experience.

Root cause: the edit-mode wrapper carries the heading class
(`ant-typography-h{level}`), so it already receives the heading font size, and
`getTitleStyles` tries to forward that to the textarea via
`div&-h{level} > textarea`. But `<TextArea>` wraps the native `<textarea>`, so
that direct-child selector never matches; the only rule that does match
(`&-edit-content textarea` in `getEditableStyles`) set no `font-size`, leaving
the control on the browser default.

Solution: make the editing textarea inherit `font-size` and `line-height` from
its wrapper in `getEditableStyles`. Since the wrapper already computes the
correct size per variant (heading size for `Title`, base text size for
`Text`/`Paragraph`), inheritance yields the element-accurate size in every case
without per-level branching. The change is additive and scoped strictly to edit
mode. A regression test renders an editable `Title` with a customized
`fontSizeHeading1` and asserts the generated edit-content textarea CSS inherits
the font size.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix `Typography` editable textarea not following the edited element's font size, e.g. customized heading tokens. |
| 🇨🇳 Chinese | 🐞 修复 `Typography` 可编辑模式下文本框字号不跟随被编辑元素（如自定义标题 token）的问题。 |


components/typography/style/mixins.ts
```diff
@@ -269,6 +269,11 @@ export const getEditableStyles: GenerateStyle<TypographyToken, CSSObject> = (tok
 
       textarea: {
         margin: '0!important',
+        // Inherit the edited element's font size / line height so the editing
+        // textarea matches the rendered text (including customized heading
+        // tokens such as `fontSizeHeading1`), keeping editing WYSIWYG.
+        fontSize: 'inherit',
+        lineHeight: 'inherit',
```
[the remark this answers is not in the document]
gaurav0107: Good catch. Added `fontFamily: 'inherit'` and `fontWeight: 'inherit'` alongside the size/line-height so the editing textarea fully mirrors the rendered element (including the strong heading weight). Addressed in ec9eb56651.

components/typography/__tests__/editable.test.tsx
```diff
@@ -102,4 +103,34 @@ describe('Typography.Editable', () => {
...
+      'ant-typography-h1',
+    );
+
+    const styleText = Array.from(document.querySelectorAll('style'))
+      .map((style) => style.innerHTML)
+      .join('');
+    expect(styleText).toMatch(/edit-content[^{}]*textarea\{[^}]*font-size:inherit/);
+    expect(styleText).toMatch(/edit-content[^{}]*textarea\{[^}]*line-height:inherit/);
+    expect(styleText).toMatch(/edit-content[^{}]*textarea\{[^}]*font-family:inherit/);
+    expect(styleText).toMatch(/edit-content[^{}]*textarea\{[^}]*font-weight:inherit/);
```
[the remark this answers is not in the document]
gaurav0107: Thanks for the careful read. This assertion targets a static style rule (`fontSize/lineHeight/fontFamily/fontWeight: 'inherit'`) that does not depend on any token, so it is emitted only when the source itself carries those declarations. I verified the regression empirically: reverting the fix in `mixins.ts` makes this test fail, so cssinjs cache reuse across the process cannot mask a regression here — if the source lacked `inherit`, no test in the process would inject that rule. The regex is also scoped to the single `edit-content … textarea{…}` block rather than matching `inherit` anywhere. Keeping the test minimal per the repo's precise-change guidance, but happy to scope it to the rendered hash if a maintainer prefers.

components/typography/__tests__/editable.test.tsx
```diff
@@ -102,4 +103,34 @@ describe('Typography.Editable', () => {
...
+
+    unmount();
```
afc163: We can drop this test case that proving A === A.

gaurav0107: Good call — removed the redundant `toHaveClass('ant-typography-h1')` assertion (it only re-checked the heading class) in b53bd89. Kept the `font-*: inherit` style assertions, since those fail when the fix is reverted, so the test still guards the actual regression.

components/typography/style/mixins.ts
```diff
@@ -269,6 +269,15 @@ export const getEditableStyles: GenerateStyle<TypographyToken, CSSObject> = (tok
 
       textarea: {
         margin: '0!important',
+        // Inherit the edited element's typography so the editing textarea
+        // matches the rendered text (including customized heading tokens such
+        // as `fontSizeHeading1` and the strong heading weight), keeping editing
+        // WYSIWYG. Browsers do not inherit font properties for form controls by
+        // default, so they are set explicitly here.
+        fontSize: 'inherit',
+        lineHeight: 'inherit',
```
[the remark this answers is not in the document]
gaurav0107: Good catch. The `.ant-typography-edit-content textarea` rule shared the same (0,1,1) specificity as Input's `textarea.ant-input { line-height: token.lineHeight }` and was registered first, so on the raw TextArea path the Input rule could win the cascade and the inherited `line-height` would not apply.

Fixed in 988b5a1 by doubling the `-edit-content` class, which raises the selector to (0,2,1) so the inherited line-height (and the other inherited font properties) reliably override Input's rule regardless of injection order. The test now also asserts the doubled-class selector, so the cascade fix can't silently regress.

components/typography/__tests__/editable.test.tsx
afc163: Remove the entire "editing textarea inherits the edited element font size" test case.

gaurav0107: Done — removed the entire `editing textarea inherits the edited element font size` test case (and its now-unused `ConfigProvider` import) in 5ed864f. The PR now contains only the `mixins.ts` fix. Verified locally: the remaining Typography editable suite passes (4/4) and eslint/biome/prettier are clean on the change.

gaurav0107:
@coderabbitai full review

afc163:
The code feedback has been addressed, but the previous visual report was broad and is now stale. Please rebase and generate a fresh visual report for the Typography change before approval.

afc163:
The selector and inherited font properties now match the reported Typography behavior, and the checks are green. Please add a focused regression case that enters editable mode and verifies the textarea inherits the customized heading font size/line height; the current visual suite passes without exercising the edit state.

afc163 (approved):
The current selector gives the editable textarea sufficient specificity and inherits the rendered Typography font properties. The author also followed the latest maintainer instruction to remove the earlier brittle style-inspection test, and the required checks are green.

afc163 (approved):

Issue #32268, opened 2021-09-23
Antd Table with row select option with some actions, the dropdown arrow is partly visible when a column is left fixed


varsha675:
- [ ] I have searched the [issues](https://github.com/ant-design/ant-design/issues) of this repository and believe that this is not a duplicate.

### Reproduction link
[https://codesandbox.io/s/custom-selection-antd-4-17-0-alpha-3-forked-m1m0e?file=/index.js](https://codesandbox.io/s/custom-selection-antd-4-17-0-alpha-3-forked-m1m0e?file=/index.js)

### Steps to reproduce
- To the table add the Row select option.
- To the row selection add a custom select dropdown.
- Then make a column left fixed.

### What is expected?
Tables row select dropdown fully visible

### What is actually happening?
Tables row select dropdown is partly visible.

| Environment | Info |
|---|---|
| antd | 4.16.13 |
| React | 16.8.2 |
| System | 1909 |
| Browser | 94.0.4606.54 |

varsha675:
Link to reproduce the [issue ](https://codesandbox.io/s/custom-selection-antd-4-17-0-alpha-3-forked-m1m0e?file=/index.js)

1084350607:
1

PR #58609, opened 2026-07-07
fix(Table): selection column z-index not working with fixed columns

leshalv:
### 🔗 Related issue link

- #32268 (original issue)
- #32276 (first fix in v4)
- #39940 (fix in v5)

### 💡 Background and solution

This is a regression from the v6 migration to logical direction properties.

In v6, `rc-table` renamed fixed column classes:
- `.ant-table-cell-fix-left` → `.ant-table-cell-fix-start`
- `.ant-table-cell-fix-right` → `.ant-table-cell-fix-end`

The `fixed.ts` style was updated to use the new names, but `selection.ts` was missed. The z-index rule at line 60:

```ts
[`table tr th${componentCls}-selection-column${componentCls}-cell-fix-left`]

### CODE REVIEW

File: components/table/style/selection.ts [hunk 1]
```diff
@@ -60,8 +60,8 @@ const genSelectionStyle: GenerateStyle<TableToken, CSSObject> = (token) => {
         },
       },
 
-      [`table tr th${componentCls}-selection-column${componentCls}-cell-fix-left`]: {
-        zIndex: calc(token.zIndexTableFixed).add(1).equal({ unit: false }),
+      [`table tr th${componentCls}-selection-column${componentCls}-cell-fix-start`]: {
```
zombieJ, line 63: 看了一下，好像这个 ```[`table tr th${componentCls}-selection-column${componentCls}-cell-fix-left`]``` 本身已经没用了？

nrps9909 (approved):
Approved exact head 5b547b973174b26208790cd0c79bc5cfc297312c after checking the maintainer question about whether this rule is still used. A focused render with rowSelection.fixed and a following fixed-start column produces a selection header with both ant-table-selection-column and ant-table-cell-fix-start, so the new selector is active. The generated CSS applies the special forward --z-offset rule, while the generic fixed-cell rule uses --z-offset-reverse; the rendered selection header's forward offset is higher than the following fixed header's reverse offset, which preserves the intended stacking order from #39940. The focused probe passed, and this commit applies cleanly to current master. The exact-head remote lint, build, module, React, coverage, visual, size, and security checks are green. A local run of the existing rowSelection file passed 71/74 tests; the other three were only rc-menu generated-id snapshot differences caused by exercising this older lockfile against the current shared dependencies, not this CSS change. AI assistance disclosure: Codex was used to trace the historical rule, inspect current rendered classes and CSS variables, create the temporary exact-head probe, and draft this review; I verified the evidence against the cited commit.

nrps9909 (commented):
Timing correction and current-head validation: the branch merged master at the same time as my preceding review, and GitHub associated that review with merge head 603e4f461dea64949c12d79dd9bcdccf10b3092f. I fetched that exact current head and re-ran the verification. Its only diff from current master remains the 2-line selection.ts change. The official Table rowSelection file plus the temporary selector/offset probe pass 76/76 tests and 4/4 snapshots with current dependencies, and the worktree is clean after removing the probe. Remote checks for this new merge head are still running, so the earlier sentence saying all remote checks were green applies to the pre-merge head only and should not be read as a claim about the still-running contexts. AI assistance disclosure remains as stated in the approval; I verified this correction against the current public head.

PR #58411, opened 2026-06-17
fix: align compact small control height

nightt5879:
﻿## Summary

- keep compact `controlHeightSM` large enough to contain the compact font line box plus borders
- add a compact algorithm regression assertion for the small control height token

## Root Cause

The compact algorithm reduces the default `controlHeight` from 32px to 28px and then derived `controlHeightSM` as `28 * 0.75 = 21px`. Compact small inputs still render a 12px font with a 20px line box plus 2px of borders, so the input cannot visually fit into 21px and renders at 22px while the search button stays at 21px.

Fixes #54859

## Validation

- `npm test -- components/config-provider/__tests__/theme.test.tsx --runInBand`
- `npm test -- components/input/__tests__/Search.test.tsx --runInBand`
- `npm run tsc`
- `npx eslint components/theme/themes/compact/index.ts components/config-provider/__tests__/theme.test.tsx`

PR #58616, opened 2026-07-08
test(Input.Search): add compact small button height regression test

### LINKED ISSUE #54859: Search (Input) small size height mismatch in compact algorithm
### Reproduction link
 
[Edit on CodeSandbox](https://codesandbox.io/p/devbox/search-box-antd-5-27-2-forked-tcp544?workspaceId=ws_Lo3sSTLJYNzwNmvJvxoXaL)
 
### Steps to reproduce
 
import { ConfigProvider, theme } from "antd";
import { Input } from 'antd';
const { Search } = Input;
 
const App: React.FC = () => (
<ConfigProvider
  theme={{
    algorithm: theme.compactAlgorithm,
  }}
>
  <Search placeholder="compact small search" size="small" style={{ width: 200 }} />
</ConfigProvider>
);
 
### What is expected?
 
    components: {
      Input: {
        controlHeightSM: 22, // only for compact algorithm
      },
    },
 
Setting the token as shown above resolves the issue, but the calculation of controlHeightSM for the compact algorithm should be corrected within the Ant Design theme itself.
 
When using the compact algorithm theme, the Search component with size="small" shows a height mismatch between the input box and the control button. The input box height is 22px, while the button height is 21px, causing a visible imbalance.
 
| Environment | Info |
| --- | --- |
| antd | 5.27.2 |
| React | 18 |
| System | Windows 11 |
| Browser | 139.0.7258.139 (Official Build) (64-bit) |

Joybanter:
## Summary

- Add a regression test verifying that in `compactAlgorithm` with `size="small"`, the `Input.Search` component button height is aligned with (or greater than) the input height.

## Root Cause

The compact algorithm reduces `controlHeight` from 32px to 28px and derives `controlHeightSM = 28 * 0.75 = 21px`. Compact small inputs render at 22px (12px font * lineHeight + 2px borders), while the search button stays at 21px, causing a 1px mismatch.

The fix (PR #58411) added dynamic button height calculation in `search.ts`. This PR adds the missing regression test.

## Related Issue

Fixes #54859

## Validation

```bash
npm test -- components/input/__tests__/Search.test.tsx --runInBand

### DIFF
--- a/components/input/__tests__/Search.test.tsx
+++ b/components/input/__tests__/Search.test.tsx
@@ -7,6 +7,7 @@ import mountTest from '../../../tests/shared/mountTest';
 import rtlTest from '../../../tests/shared/rtlTest';
 import Button from '../../button';
 import ConfigProvider from '../../config-provider';
+import theme from '../../theme';
 import type { InputRef } from '../Input';
 import Search from '../Search';
 import type { SearchProps } from '../Search';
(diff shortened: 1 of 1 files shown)


afc163:
Thanks for adding coverage. Closing because the underlying compact-small height fix was already merged in #58411, while this test hard-codes the internal value `22` rather than asserting the user-visible behavior. A follow-up test should be behavior-based.

PR #58645, opened 2026-07-09
fix(watermark): support nativeElement ref

li-jia-nan:
Extracted from #58627.

Changes:
- Adds `nativeElement` ref support for `Watermark`.
- Exports the new ref type through the watermark entry and package entry.
- Adds component-level ref coverage.

Validation:
- `git diff --check`

### DIFF
--- a/components/index.ts
+++ b/components/index.ts
@@ -181,7 +181,13 @@ export { default as Upload } from './upload';
 export type { DraggerProps, UploadFile, UploadProps } from './upload';
 export { default as version } from './version';
 export { default as Watermark } from './watermark';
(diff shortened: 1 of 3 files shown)


zombieJ:
Maybe we can use the existing ref compose util here, such as `useComposeRef`, to combine the internal container ref with `nativeElementRef`. That would avoid manually creating a callback ref while preserving the current `setContainer` behavior.

zombieJ:
Since we are using `useComposeRef` here, we can pass `setContainer` directly and remove the extra `useCallback` wrapper. `useComposeRef(nativeElementRef, setContainer)` should be enough.

zombieJ (approved):

PR #58670, opened 2026-07-11
fix(tag): fix preset filled variant low contrast in dark mode

### LINKED ISSUE #56339: Variant=filled Tags are ugly in dark mode
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/devbox/laughing-mopsa-tcwqf9)

### Steps to reproduce

1. make the tag 
2. switch to dark mode

### What is expected?

what we expect is the same style as the light mode but optimized for dark mode.
the light mode filled variant follows Presets (filled) but the dark mode is ugly

the same issue exists on the website if you switch to dark mode
https://ant.design/components/tag?theme=dark

the tags with variant filled looks great in light mode but on dark mode they seem bugged and ugly.

| Environment | Info |
| --- | --- |
| antd | 6.1.1 |
| React | 19 |
| System | Mac OS  |
| Browser | Latest chrome |

raihan-sifat:

### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

close #56339

### 💡 Background and Solution

In dark mode, `<Tag color="..." variant="filled">` used preset palette index 1 as its background, which resolves to a near-black tint nearly identical in luminance to `colorBgContainer` — so the tag rendered as an almost invisible blob instead of a distinct chip.

Fixed by deriving the filled variant's dark-mode background from the palette's saturated step instead (15% alpha blended onto `colorBgContainer`), computed in `prepareComponentToken` so it resolves correctly under both standard and CSS-variable (`cssVar`) theming — doing the color math directly in the style-generation function breaks under `cssVar`, since token values there are `var(...)` reference strings rather than real colors. Light mode is unchanged (verified via unchanged Jest snapshots).

**Before (dark mode):**
before

**After (dark mode):**
after

**Light mode (unchanged):**
Screenshot 2026-07-11 184711

The visual difference is most apparent for `purple` and `geekblue` text legibility specifically — these were the two colors whose text-on-background contrast fell below the WCAG 3:1 minimum before this fix (2.08:1 and 2.37:1). The overall tag-vs-page boundary contrast is a separate, harder problem this PR does not attempt to solve (see discussion above).

Note: #58590 also targets this issue. This PR's approach additionally fixes two contrast failures present in that PR's implementation (text-on-background contrast for `purple` and `geekblue` preset colors fell below the WCAG 3:1 minimum there).

### 📝 Change Log

| Language | Changelog |
| --- | --- |
| 🇺🇸 English | 🐞 Fix Tag low contrast for preset color `filled` variant tags in dark mode. |
| 🇨🇳 Chinese | 🐞 修复 Tag 预设颜色 `filled` 变体标签在暗色模式下对比度过低的问题。 |

### DIFF
--- a/components/tag/__tests__/__snapshots__/demo-extend.test.ts.snap
+++ b/components/tag/__tests__/__snapshots__/demo-extend.test.ts.snap
@@ -949,6 +949,240 @@ exports[`renders components/tag/demo/customize.tsx extend context correctly 1`] 
 
 exports[`renders components/tag/demo/customize.tsx extend context correctly 2`] = `[]`;
 
+exports[`renders components/tag/demo/dark.tsx extend context correctly 1`] = `
+<div
+  class="ant-flex css-var-test-id ant-flex-align-stretch ant-flex-gap-small ant-flex-vertical"
+>
+  <div
+    class="ant-divider css-var-test-id ant-divider-horizontal ant-divider-with-text ant-divider-with-text-start ant-divider-plain"
+    role="separator"
+  >
+    <div
+      class="ant-divider-rail ant-divider-rail-start"
+    />
+    <span
+      class="ant-divider-inner-text"
+    >
+      Filled
+    </span>
+    <div
+      class="ant-divider-rail ant-divider-rail-end"
+    />
+  </div>
+  <div
+    class="ant-flex css-var-test-id ant-flex-wrap-wrap ant-flex-align-center ant-flex-gap-small"
+  >
+    <span
+      class="ant-tag ant-tag-filled ant-tag-magenta css-var-test-id"
+    >
+      magenta
+    </span>
+    <span
+      class="ant-tag ant-tag-filled ant-tag-red css-var-test-id"
+    >
+      red
+    </span>
+    <span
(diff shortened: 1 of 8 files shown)


afc163 (changes requested):
This is the implementation we should keep for #56339, and using colorBgContainer for dark-mode detection is the safer direction. Before merging, please add a focused dark-theme filled Tag demo or image regression covering the preset colors; the current PR changes the visual behavior without adding a test or snapshot.

raihan-sifat:
> This is the implementation we should keep for #56339, and using colorBgContainer for dark-mode detection is the safer direction. Before merging, please add a focused dark-theme filled Tag demo or image regression covering the preset colors; the current PR changes the visual behavior without adding a test or snapshot.

Thanks for the feedback! I'll add the demo / snapshot test covering the filled Tag preset colors in dark mode shortly and push the update to this PR.

afc163:
The focused dark-theme demo and snapshots address the earlier code-coverage request, but the only visual report is still based on master@25ae922a and contains 70 broad diffs. Please rebase onto current master and regenerate a focused current visual report; do not reuse the stale report or mark its checkbox acceptable. Once the refreshed Tag diffs can be inspected, this can be reviewed for approval.

PR #51381, opened 2024-10-25
fix(icon-button): Button icon not vertically centered error

Wxh16144:
[中文版模板 / Chinese template](https://github.com/ant-design/ant-design/blob/master/.github/PULL_REQUEST_TEMPLATE_CN.md?plain=1)

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📝 Site / documentation improvement
- [ ] 📽️ Demo improvement
- [x] 💄 Component style improvement
- [ ] 🤖 TypeScript definition improvement
- [ ] 📦 Bundle size optimization
- [ ] ⚡️ Performance optimization
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] 🎨 Code style optimization
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

fix #51380

> - Describe the source of related requirements, such as links to relevant issue discussions.
> - For example: close #xxxx, fix #xxxx

### 💡 Background and Solution

> - The specific problem to be addressed.
> - List the final API implementation and usage if needed.
> - If there are UI/interaction changes, consider providing screenshots or GIFs.

### 📝 Change Log

> - Read [Keep a Changelog](https://keepachangelog.com/en/1.1.0/) like a cat tracks a laser pointer.
> - Describe the impact of the changes on developers, not the solution approach.
> - Reference: https://ant.design/changelog

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    fix(icon-button): Button icon not vertically centered error       |
| 🇨🇳 Chinese |     修复按钮图标未垂直居中错误      |

Issue #45021, opened 2023-09-22
Tag 组件带上 Icon 时，文字与图标没有垂直居中对齐


Elliott-Hu:
### Reproduction link

[https://ant-design.antgroup.com/components/tag-cn](https://ant-design.antgroup.com/components/tag-cn)

### Steps to reproduce

官方提供的文档示例页面搜索 With icon 可以复现

### What is expected?
![image](https://github.com/ant-design/ant-design/assets/16644716/02bc2822-1ab7-4f9f-8e4a-5a2d53603c9e)

希望使用时图标与文字不用做样式处理，默认垂直居中

### What is actually happening?

图标相对图片位置偏上，行高不一致

| Environment | Info |
| --- | --- |
| antd | 5.9.2 |
| React | 18 |
| System | mac |
| Browser | chrome |

LiangJavaDS:
实现思路是line-height === 容器height？

LiangJavaDS:
@MadCcc  实现思路是line-height === 容器height吗？如果不牵扯太多是否能让我尝试修复？

zomblemta:
我在控制台查了一下样式，是icon的vertical-align值略大，改到-.19em差不多

zomblemta:
再次排查后，我发现是因为icon的vertical-align的单位为em，当我把外层的div的字体大小调整到12px之后，样式就正常了。

MadCcc:
@liangkuku PR welcome~

liangkuku:
> @liangkuku PR welcome~

I will try to fix it.

liangkuku:
My approach is to change the parent span tag of the icons and text  from `display: 'inline-block'` to `display: 'inline-flex'` in order to achieve vertical centering.Do you think there is any issue with this approach?
![image](https://github.com/ant-design/ant-design/assets/62460957/bd7ed301-974d-48f0-b028-be84c095a704)

liangkuku:
I went to see the demo again.It seems that the icon has been centered now.
![image](https://github.com/ant-design/ant-design/assets/62460957/434c0e5b-5697-44c5-9f62-03d774c05297)

PR #58723, opened 2026-07-16
fix(Tag): align and space bare `<svg>` icons

mohamedkhaled4053:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [x] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

- Addresses the `Tag` part of #56315 ("Missing Margins ... resulting in icons colliding with text", "Misalignment: Icons lack the specific `vertical-align: -0.125em`"). It does not close that issue, which is broader than Tag.
- Same symptom as #45021 (closed by the stale bot, never actually fixed).
- Prior attempt: #45236 — closed unmerged. **Please read the note below before reviewing**; this PR is deliberately built to not repeat it.

#### The problem

Every icon rule in `Tag` is keyed on `.anticon` — a class only `@ant-design/icons` sets. Icons from third-party libraries (lucide-react, react-icons, `@phosphor-icons/react`) render a **bare `<svg>`**, so they match none of them:

1. **No spacing.** `components/tag/style/index.ts:147` is ``[`> ${token.iconCls} + span`]`` = `.ant-tag > .anticon + span`. A bare `<svg>` never matches, so the icon sits **flush against the text** (measured: 7px → **0px**).
2. **No baseline correction.** `.anticon` carries `vertical-align: -0.125em`; a bare `<svg>` gets nothing. An `<svg>` is an inline replaced element with **no baseline of its own**, so per CSS 2.1 §10.8.1 its **bottom margin edge** is aligned to the text baseline and the icon floats above the text.
3. A 16px `<svg>` also **grows the tag** (22px → 24px), which is the "Vertical Stacking in Tags" from #56315.

Offset from the text cap-centre (headless Chromium, real SSR'd styles). `+` = icon rides above the text; `0` = centred. The fix holds at **every** icon size, not just `1em`:

| | before | after |
| --- | --- | --- |
| bare `<svg>` — gap to text | **0px** (collides) | **7px** (same as our icons) |
| bare `<svg>` 1em  — offset from cap-centre | +1.50px | **+0.13px** |
| bare `<svg>` 14px | +2.50px | **+0.13px** |
| bare `<svg>` 16px | +3.50px | **+0.13px** |
| bare `<svg>` 24px | +7.50px | **+0.13px** |
| bare `<svg>` 16px — tag height | **24px** | **22px** |

For reference, `@ant-design/icons` inside a Tag measure `+0.50px` against the same cap-centre, so a corrected bare `<svg>` ends up marginally *more* centred than antd's own icons.

| BEFORE | AFTER |
| --- | --- |
| before | after |

#### Why this does not repeat #45236

**This is the first thing to check, so I am putting it up front.** #45236 tried to fix the same bug with:

```ts
[`> ${token.iconCls}`]: { verticalAlign: '-0.15em' },
```

and @afc163 closed it with *"It will affect close x icon position"* plus a red Argos diff. That is correct: the **close icon is also an `.anticon` and also a direct child** of the Tag root, so `> .anticon` moves it too.

This PR selects `> svg` instead, which **cannot** match any `@ant-design/icons` element: `.anticon` nests its `<svg>` one level deeper (`.ant-tag > .anticon > svg`). `> svg` matches **only** a bare third-party `<svg>`. Measured — every `@ant-design/icons` path is byte-identical to master:

| | before | after |
| --- | --- | --- |
| `<Tag icon={<CheckCircleOutlined />}>` — icon offset / gap / height | 0.00px / 7px / 22px | **identical** |
| close icon — offset from left / margin / vertical offset | 53.36px / 7px / −0.75px | **identical** |
| `<Tag closable>` (no icon) — close icon left / margin | 30.36px / 3px | **identical** |
| `<Tag><PlusOutlined /> New Tag</Tag>` (the `demo/animation.tsx` pattern) | 0.00px / 22px | **identical** |

Because of this, **no existing demo, test or snapshot changes** — every `<svg>` in the Tag
snapshots is inside an `.anticon`. The snapshot additions in this PR are purely the new demo
section described below (the diff on both snapshot files is additive, with zero removed lines).

#### Why not `display: flex` on the root, like the recent Switch fix (#58672)?

I checked, because it is the obvious next question. **It is not safe here, and this is the #58428 trap.**

`Switch` was safe because its root is `inline-block` + `vertical-align: middle` + `overflow: hidden`, so its internal baseline is not observable from outside. **Tag's root has neither `vertical-align` nor `overflow`** — it *exports* its baseline to the inline flow. Per CSS Flexbox §8.5 an inline-flex box takes its baseline from its **first flex item**, which for an icon-leading Tag is a baseline-less `<svg>`. Measured with `display: inline-flex; align-items: center` on `.ant-tag`:

```
Tag inline in a paragraph:  top −3px → −5px,  paragraph height 22px → 24px
Icon Tag next to plain Tag: aligned → misaligned by 2px
@ant-design/icons offset:   0.00px → −0.50px   (i.e. our own icons get worse)
```

That is exactly the regression class that got Button's flex fix reverted twice (#51381→#51588, #57896→#58429). A flex root would also strip the space in `<Tag><PlusOutlined /> New Tag</Tag>`, antd's own demo. This PR touches neither the root's `display` nor its baseline — measured identical before/after.

#### Why `middle` + `margin-block-end`, not `-0.125em`

An earlier revision of this PR used `vertical-align: -0.125em`, copying `.anticon`'s own correction. That was **wrong for the general case**, and I want to be explicit about why, because it is subtle:

`-0.125em` resolves against the **Tag's font size**, so it is a *constant* nudge regardless of icon size. `.anticon` gets away with it only because its `<svg>` is pinned to `width: 1em; height: 1em`, so a font-relative correction always fits. A third-party `<svg>` carries no such guarantee — lucide, tabler and feather all default to `24px` — and against those the constant under-corrects, growing to **~6px off at 24px**. It happened to look perfect at `1em`, which is exactly the one size the demo used, so the defect hid itself.

The shipped rule is size-independent:

- `vertical-align: middle` centres the icon's **margin box** on the line's x-height midpoint — a ratio, so it holds at any size.
- `margin-block-end: 0.2em` then lifts the box by half its own value (`0.1em`) from the x-height centre up to the **cap-height** centre, since `capHeight − xHeight ≈ 0.2em` across typical fonts. Measured residual: **~0.1px flat** from `1em` to `24px`, and ≤0.2px across the serif / sans / mono families I could render distinctly.

I use the **logical** `margin-block-end` (not `margin-bottom`) to match the file's own `marginInlineStart` convention and stay correct in vertical writing modes; in Tag's horizontal-tb they are identical.

#### Disclosure

- **Centring, not `.anticon`-matching.** The rule now lands bare icons at `+0.13px` while `@ant-design/icons` sit at `+0.50px`, so a third-party icon is *marginally more centred* than antd's own. This optimises true optical centring rather than byte-consistency with `.anticon`; I think that is the right target for an arbitrary-size icon, but flag it as a deliberate design call.
- **An oversized icon still enlarges the tag.** A `24px` `<svg>` grows the tag to `28.4px` — unavoidable when the icon is taller than the `20px` line strut, and `height` on an `<svg>` would resize the graphic rather than its box. The fix still yields the *shortest* of the three options (shorter than both master and `-0.125em`) and never flexes the root, so the exported-baseline / #58428 trap is not triggered; `1em`/`14px` paragraph layout is byte-identical to master.
- **Specificity.** Like any antd token rule, `> svg { margin-block-end }` computes to a `:where(hash).ant-tag > svg` selector that will override a utility `mb-*` class a user puts on the icon. The prior `-0.125em` already did this for `align-*`; this widens it to margin. An inline `styles={{ icon: { marginBottom: 0 } }}` still wins.
- **Bare-`<svg>` custom close icons** (`closeIcon={<svg/>}`) now get the same centring, because `replaceElement` clones the node rather than wrapping it. That is a small improvement (close-icon offset `0.75px → 0.33px`), not a regression; the **default** close icon is a `CloseOutlined` `.anticon`, whose `<svg>` is one level deeper and therefore untouched — the exact hazard #45236 tripped on stays clear.

#### Verification

- `jest --config .jest.js --no-cache components/tag` → **6 suites / 93 tests passed, 35 snapshots passed, 0 written**.
- `tsc --noEmit`, `biome check`, `eslint` — all clean.
- jsdom has no layout engine, so this cannot be unit-tested; all numbers above are measured in headless Chromium against the real component (`renderToString` + `extractStyle` from this branch, `<!DOCTYPE html>`).

### 🔁 Review round 2

Both review points are addressed:

**1. Merged the duplicated spacing selectors** (thanks @gemini-code-assist / @coderabbitai). The bare
`<svg>` spacing now folds into the existing `.anticon` rule, so the style file is net **+4 / −8** — the
fix ships smaller than the code it replaces:

```ts
// To ensure that a space will be placed between character and `Icon`.
[`> ${token.iconCls} + span, > span + ${token.iconCls}, > svg + span, > span + svg`]: {
  marginInlineStart: paddingInline,
},
```

`.anticon` and `svg` are disjoint selectors (an `.anticon` keeps its `<svg>` one level deeper), so the
merge is a pure no-op: both snapshot files diff with **zero removed lines**.

**2. Added a demo with an inline bare `<svg>`.** Good call — the "no snapshots change" property above is
what makes this fix safe, but it also meant *nothing in CI exercised the fix*. The Tag icon demo now has
a **Tag with third-party icon** section using bare inline `<svg>` icons sized in both `em` and `px`.

It deliberately includes a **closable** tag, which puts a bare `<svg>` and the close icon in the same
tag — precisely the combination that sank #45236 — so visual regression now guards that case directly.

Measured on the new demo (headless Chromium, real SSR'd styles), master vs this PR:

A **px-sized** icon is the important one: it exercises the size-independence, and it is what most icon
libraries actually ship. Sizing the demo only in `em` would have masked the defect described above.

The 7px spacing shift is well above the visual-diff threshold, so the demo is a real regression guard
rather than a decorative one, while the default close icon and tag height stay byte-identical to master
(the #45236 guarantee).

### 🔁 Review round 3 — icon alignment

@mohamedkhaled4053 (reviewing the preview) spotted that a bare `<svg>` still rode slightly high at larger
sizes. That was real: the `-0.125em` this PR originally shipped is a *constant* correction, so its error
grows with icon size (~2px at 16px, ~6px at 24px) — see **Why `middle` + `margin-block-end`** above. I
verified the alternatives with an adversarial pass (which, usefully, also refuted my own "match `.anticon`"
rationale — Button's `-0.125em` is inert because its icon is a flex item) and switched to the
size-independent `vertical-align: middle` + `margin-block-end: 0.2em`. Residual is now ~0.1px flat at every
size. The demo now includes a px-sized icon so this is actually covered by visual regression.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Tag not spacing or vertically centring icons from third-party libraries, which render a bare `<svg>` and so missed the spacing and alignment that `@ant-design/icons` receive. |
| 🇨🇳 Chinese | 🐞 修复 Tag 中第三方图标库图标（渲染为原生 `<svg>`）缺少间距且未垂直居中的问题，此前这类图标无法命中 `.anticon` 相关样式。 |


li-jia-nan:
Thanks for the detailed investigation — the scoped `> svg` approach looks sound.

Could you consider incorporating the bot's suggestion to combine the duplicated spacing selectors? It would also be helpful to add a small demo using an inline bare `<svg>`. That would ensure the visual regression suite covers the exact scenario this PR fixes, while also providing a convenient reproduction for future debugging.

Thanks!

mohamedkhaled4053:
Thanks @li-jia-nan — both done in 6d8f86b.

**Merged the spacing selectors.** `.anticon` and `svg` are disjoint (an `.anticon` keeps its `<svg>` one level deeper), so the merge is a pure no-op — both snapshot files diff with zero removed lines. The style file is now net **+4 / −8**, so the fix ships smaller than the code it replaces.

**Added the bare `<svg>` demo** — this was a good catch, and it's the flip side of something I'd argued in the description. "No existing snapshot changes" is what makes this fix safe, but it also meant *nothing in CI exercised the fix*. The Tag icon demo now has a **Tag with third-party icon** section using an inline bare `<svg>`.

I made one of them **closable** on purpose: that puts a bare `<svg>` and the close icon in the same tag — exactly the combination that sank #45236 — so visual regression guards that case directly from now on.

Measured on the new demo (headless Chromium, real SSR'd styles), `master` vs this PR:

| | `master` | this PR | `.anticon` reference |
| --- | --- | --- | --- |
| gap between icon and text | **0px** (collides) | **7px** | 7px |
| icon centre vs text centre | **−2.00px** | **−0.50px** | 0.00px |
| close icon offset in tag | 6px | **6px** (unchanged) | 6px |
| tag height | 22 / 23px | **22 / 23px** (unchanged) | 22 / 23px |

The 7px icon shift is well clear of the visual-diff threshold, so the demo is a real guard rather than a decorative one — while the close icon and tag height stay byte-identical to master.

The demo sizes the `<svg>` in `em` (what `react-icons` defaults to, and what `.anticon` uses internally). That's also the honest framing of the residual I disclosed: `-0.125em` is exact at `1em` and drifts ~1px at 14px. I added a line to the docs saying so, since it's useful guidance for third-party icons either way. If you'd rather kill the residual outright at any size, that needs a wrapper element around the icon (a DOM change) — happy to do it in a follow-up if you prefer that shape.

`jest --config .jest.js --no-cache components/tag` → 6 suites / 93 tests / 35 snapshots passed, 0 written, on top of the latest master.

mohamedkhaled4053:
Follow-up: I changed the **alignment mechanism** in aed3abd, and the demo/images/description above are updated to match.

Reviewing the preview, it was pointed out that a bare `<svg>` still rode slightly high at larger sizes — which was correct. The `-0.125em` I had (copied from `.anticon`) is a **constant** correction, so its error grows with icon size: fine at `1em`, but ~2px high at 16px and **~6px at 24px** — and 24px is the default of lucide, tabler and feather. It only looked perfect in the demo because the demo happened to use `1em`.

The fix is now **size-independent**:

```ts
'> svg': {
  verticalAlign: 'middle',   // centres the margin box on the x-height line (a ratio → any size)
  marginBlockEnd: '0.2em',   // lifts it onto the cap-height centre (capHeight − xHeight ≈ 0.2em)
},
```

Residual is ~0.1px flat from 1em to 24px (was 0 / −2 / −6). Measured, master → this PR:

| bare `<svg>` | 1em | 14px | 16px | 24px |
| --- | --- | --- | --- | --- |
| master | −1.5 | −2.5 | −3.5 | −7.5 |
| `-0.125em` (was) | 0 | −1 | −2 | −6 |
| **`middle` + 0.2em (now)** | **+0.13** | **+0.13** | **+0.13** | **+0.13** |

Notes:
- Used the **logical** `margin-block-end` to match the file's `marginInlineStart` and stay correct in vertical writing modes.
- The demo now sizes one icon in **px** (lucide's default) alongside the `em` one, so visual regression actually exercises the size-independent path — the earlier em-only demo would have hidden this.
- The default close icon (`.anticon`) is still byte-identical to master; `> svg` can't reach it. A bare-`<svg>` *custom* close icon now gets centred too (a small improvement, disclosed in the description).

I also want to flag honestly: I initially claimed this held across "6 fonts" — on recheck, 5 of those font names weren't installed in my test env and silently fell back to the same face. The real coverage is the 3 distinct families (sans / serif / mono) that actually rendered, where it's ≤0.2px. The size-independence, which is the load-bearing property, is geometry rather than font-dependent.

Gates green on the latest master: 6 suites / 93 tests / 35 snapshots, 0 written.

mohamedkhaled4053:
Heads-up on the visual-regression report (9 changed): only the **3 `tag-icon`** images are from this PR — they show the new *Tag with third-party icon* section rendering centred, which is the demo doing its job (this fix now shows up in visual regression, where before it did not).

The other 6 (`calendar-notice-calendar` ×3, `grid-playground` ×3) are unrelated flaky diffs — neither demo contains a `.ant-tag`, so the `.ant-tag > svg` rule cannot reach them. The calendar diff is sub-pixel jitter on the event labels of the fixed Nov-2016 calendar; the grid one is a layout/hydration shift. Safe to approve.

afc163:
The code/spec review looks good, and the Tag demo covers the intended bare-`<svg>` cases. The current visual report still contains six unrelated Calendar/Grid diffs because it is comparing against the older `3cc65937` baseline after master changes were merged into the branch. Please rebase onto the latest `master` and rerun visual regression. Once the report contains only the three expected Tag icon diffs, it is ready for the final visual approval.

afc163 (approved):

PR #58754, opened 2026-07-21
fix(Tag): improve filled variant visibility in dark mode

### LINKED ISSUE #56339: Variant=filled Tags are ugly in dark mode

lwyghn:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

close #56339

### 💡 Background and Solution

In dark mode, preset-color `filled` variant tags (e.g. `<Tag color="blue" variant="filled">`) used palette index 1 as background, which resolves to a near-black tint nearly identical to `colorBgContainer` — making the tag almost invisible.

**Solution:**
- In `prepareComponentToken`, detect dark mode via `colorBgBase` lightness and precompute a `xxxFilledBg` token for each preset color.
- Dark mode: use palette-6 (saturated color) at 15% alpha blended onto `colorBgContainer` as the filled background.
- Light mode: unchanged — passes through the original `lightColor` directly.
- Computed in `prepareComponentToken` (not the style function) to stay compatible with `cssVar` theming, since color tokens are already `var(...)` references by the time the style function runs.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Tag preset-color `filled` variant being nearly invisible in dark mode. |
| 🇨🇳 Chinese | 修复 Tag 预设颜色 `filled` 变体标签在暗色模式下几乎不可见的问题。 |

### DIFF
--- a/components/tag/style/index.ts
+++ b/components/tag/style/index.ts
@@ -8,8 +8,16 @@ import { isBright } from '../../color-picker/components/ColorPresets';
 import { resetComponent } from '../../style';
 import type { FullToken, GenerateStyle, GenStyleFn, GetDefaultToken } from '../../theme/internal';
 import { genStyleHooks, mergeToken } from '../../theme/internal';
+import { PresetColors } from '../../theme/interface';
+import type { PresetColorKey } from '../../theme/interface';
 
-export interface ComponentToken {
+// Per-preset-color background for the `filled` variant, one flat token per color (e.g.
+// `blueFilledBg`, `purpleFilledBg`). Must stay flat: `cssVar` theming turns every top-level
+// `ComponentToken` key into its own `var(...)` reference, so a single key holding a nested
+// object cannot survive the round trip.
+type PresetFilledBgComponentToken = Partial<Record<`${PresetColorKey}FilledBg`, string>>;
+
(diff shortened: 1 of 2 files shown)


lwyghn:
#56339

afc163:
Thanks for the contribution. This duplicates the earlier #58670 for the same issue and uses almost the same implementation. The existing PR also bases dark-mode detection on the actual `colorBgContainer` that receives the blend, which is safer for customized themes than checking `colorBgBase`. Closing this duplicate so review can continue in #58670.

PR #58762, opened 2026-07-22
fix: handle Input.OTP IME composition

### LINKED ISSUE #52093: 使用微软自带的输入法，切换至中文模式，使用Input.OTP,按一下任一字符,就会输入进2个字符

wanxiankai:
### 🤔 This is a ...

- [x] 🐞 Bug fix
- [x] ✅ Test Case

### 🔗 Related Issues

Fix #52093

### 💡 Background and Solution

Input.OTP currently handles every input event immediately. During IME composition, browsers can emit intermediate text such as pinyin before the final committed character. Those intermediate values were treated as OTP input, which could duplicate or split Chinese input across cells.

This PR tracks composition state in each OTP input and ignores input events while composition is active. The committed input event after composition ends is still handled normally, so regular typing, paste, formatter behavior, and focus movement remain unchanged.

A regression test covers the compositionStart -> input intermediate text -> compositionEnd -> committed input flow.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Input.OTP incorrectly handling IME composition text as committed input. |
| 🇨🇳 Chinese | 修复 Input.OTP 在中文输入法合成输入过程中会将中间态文本作为已输入内容处理的问题。 |

### DIFF
--- a/components/input/OTP/OTPInput.tsx
+++ b/components/input/OTP/OTPInput.tsx
@@ -16,19 +16,45 @@ export interface OTPInputProps extends Omit<InputProps, 'onChange'> {
 }
 
 const OTPInput = React.forwardRef<InputRef, OTPInputProps>((props, ref) => {
-  const { className, value, onChange, onActiveChange, index, mask, onFocus, ...restProps } = props;
+  const {
+    className,
+    value,
+    onChange,
+    onActiveChange,
+    index,
+    mask,
+    onFocus,
+    onCompositionStart,
+    onCompositionEnd,
+    ...restProps
+  } = props;
   const { getPrefixCls } = React.useContext(ConfigContext);
   const prefixCls = getPrefixCls('otp');
   const maskValue = typeof mask === 'string' ? mask : value;
 
   // ========================== Ref ===========================
   const inputRef = React.useRef<InputRef>(null);
+  const composingRef = React.useRef(false);
 
   React.useImperativeHandle(ref, () => inputRef.current!);
 
   // ========================= Input ==========================
   const onInternalChange: React.InputEventHandler<HTMLInputElement> = (e) => {
+    if (composingRef.current || (e.nativeEvent as InputEvent).isComposing) {
+      return;
+    }
+
     onChange(index, (e.target as HTMLInputElement).value);
+  };
+
+  const onInternalCompositionStart: React.CompositionEventHandler<HTMLInputElement> = (e) => {
+    composingRef.current = true;
+    onCompositionStart?.(e);
+  };
+
+  const onInternalCompositionEnd: React.CompositionEventHandler<HTMLInputElement> = (e) => {
+    composingRef.current = false;
+    onCompositionEnd?.(e);
   };
 
   // ========================= Focus ==========================
@@ -80,6 +106,8 @@ const OTPInput = React.forwardRef<InputRef, OTPInputProps>((props, ref) => {
         ref={inputRef}
         value={value}
         onInput={onInternalChange}
+        onCompositionStart={onInternalCompositionStart}
+        onCompositionEnd={onInternalCompositionEnd}
         onFocus={onInternalFocus}
         onKeyDown={onInternalKeyDown}
         onMouseDown={syncSelection}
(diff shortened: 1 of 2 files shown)


afc163 (changes requested):
This implementation still depends on a normal `input` event after `compositionend`. Some QQ/Sogou IME sequences do not provide that reliably, so the committed text can be lost. Please use the explicit commit-and-dedupe model covered in #58216 and add the corresponding event-sequence tests.

wanxiankai:
Thanks for the review. I updated this PR to follow the explicit commit-and-dedupe model from #58216:\n\n- commit the final IME value on `compositionend` so QQ/Sogou-style sequences do not lose the committed text\n- de-dupe the trailing `input` event when a browser emits the same committed value right after `compositionend`\n- keep the `nativeEvent.isComposing` guard for intermediate input events\n- add focused event-sequence tests for intermediate input, no trailing input, trailing duplicate input, multi-character commit, cancelled composition, and composition event forwarding\n\nValidated locally with:\n\n`npm test -- components/input/__tests__/otp.test.tsx --runInBand`\n`pnpm exec eslint components/input/OTP/OTPInput.tsx components/input/__tests__/otp.test.tsx`

afc163:
The new compositionend commit-and-dedupe behavior resolves the previous functional blocker. One changed line is still uncovered, leaving both Codecov patch and project checks failing; please cover that remaining branch/cleanup path so the required 100% target is restored.

wanxiankai:
Thanks for catching this. I added a focused regression test for the remaining cleanup path: after `compositionend`, the dedupe marker is cleared on the next frame, so a later input with the same value is handled normally instead of being suppressed indefinitely.

Validated locally with:

- `npm test -- components/input/__tests__/otp.test.tsx --runInBand --coverage --collectCoverageFrom=components/input/OTP/OTPInput.tsx` (33 tests passed; 100% statements/functions/lines for `OTPInput.tsx`)
- `pnpm exec eslint components/input/OTP/OTPInput.tsx components/input/__tests__/otp.test.tsx`

Pushed in `c0fdf7b`.

afc163 (approved):
This is the OTP IME candidate to keep. It implements explicit composition-end commit and deduplication, forwards consumer composition callbacks, guards native isComposing state, and covers cancellation, multi-character composition, cleanup, and trailing input. Required checks are green.

nrps9909 (approved):
Verified exact head `c0fdf7b84d026e7509c14177f8c8ffae7ccbe98c` independently.

- `npm test -- components/input/__tests__/otp.test.tsx --runInBand`: 33/33 tests and snapshot pass
- ESLint and Biome pass for both changed files; `git diff --check` is clean
- Causal mutation check: temporarily removing the composition/native-`isComposing` and trailing-input deduplication path makes four focused IME tests fail (intermediate composition text, native composing events, duplicate post-composition input, and multi-character commit), and restoring this head returns the suite to green
- The broader Input run passed 18 suites / 369 tests; its remaining demo snapshot failures are unrelated Select/Picker markup drift from the shared local dependency tree, not OTP behavior

The implementation covers composition cancellation, multi-character commits, native `isComposing`, duplicate and differing trailing input events, cleanup after the next frame, and consumer handler forwarding. I did not find a functional blocker in this head.

GitHub currently reports the PR as conflicting with `master`, so this approval is exact-head evidence only; the conflict resolution should be rechecked before merge.

AI assistance disclosure: Codex was used to inspect the implementation and history, run the exact-head checks, perform the causal mutation test, and draft this review. I independently verified the reported results.

PR #58774, opened 2026-07-22
fix: support Yarn PnP subpath imports

drewmt:
### 🤔 This is a ...

- [x] 🐞 Bug fix
- [x] ✅ Test Case

### 🔗 Related Issues

Fixes #56858

### 💡 Background and Solution

Yarn PnP cannot use legacy extension and directory lookup for deep imports such as `antd/es/input/Search`. A simple `exports` wildcard fixes that lookup, but breaks TypeScript because the compiled output mixes standalone files with directories containing `index.d.ts`.

This change adds conditional package exports and generates small forwarding entries for directory-based modules after compilation. The proxies preserve named exports, default exports, and `use client` directives, while Node conditions keep using the CommonJS build and browser/import conditions keep using the ES build.

The existing package checks now cover file-based and directory-based subpaths under TypeScript `bundler` and `node16` resolution, CommonJS entry points, and the generated artifacts.

Validation performed:

- `npm run compile`
- `./node_modules/.bin/antd-tools run dist`
- `npm run test:dekko`
- `npm run tsc`
- scoped ESLint and Biome checks
- the public Yarn 4.12 PnP reproduction on Node 22 using a packed local artifact: 2/2 tests passed

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Yarn PnP resolution for Ant Design subpath imports without breaking TypeScript type resolution. |
| 🇨🇳 Chinese | 修复 Yarn PnP 下 Ant Design 子路径导入解析，同时保持 TypeScript 类型解析兼容性。 |

PR #57968, opened 2026-05-14
fix: remove exports field to fix TypeScript module resolution

afc163:
### 🤔 What is the nature of this change?

- [x] 🐞 Bug fix

### 🔗 Related Issue

close #56858

### 💡 Background and Solution

**Problem**

PR #57930 (commit 597f94c1) added an `exports` field to `package.json` to support Yarn PnP subpath resolution. However, this breaks TypeScript type resolution under `moduleResolution: "bundler"` (Vite default) and `"nodenext"`, causing ~200+ `TS2339` errors in downstream projects.

**CI failure**: https://github.com/ant-design/ant-design-pro/actions/runs/25850042548/job/75954108350?pr=11783

**Root cause**

TypeScript's exports-based resolution does **NOT** perform directory/index lookups. When resolving `antd/es/theme/interface`, it only tries `./es/theme/interface.d.ts` — it will never check `./es/theme/interface/index.d.ts`.

Since antd's `es/` output uses a mixed structure:

- File-based: `es/config-provider/context.d.ts`
- Directory-based: `es/theme/interface/index.d.ts`

No single wildcard pattern in `exports` can correctly resolve types for both cases:

| Pattern | File-based | Directory-based |
|---|---|---|
| `"./es/*": { "types": "./es/*.d.ts" }` | ✅ | ❌ |
| `"./es/*": { "types": "./es/*/index.d.ts" }` | ❌ | ✅ |
| `"./es/*": "./es/*"` | ❌ | ❌ |

`antd-style` deep-imports `AliasToken` from `antd/es/theme/interface`. When that import fails to resolve, `FullToken` becomes an empty interface, causing all token properties to report TS2339.

**Solution**

Remove the `exports` field to restore traditional Node.js module resolution (which supports directory/index lookups). The `main`, `module`, and `typings` fields continue to work correctly for the main entry point.

Yarn PnP users can add `packageExtensions` in `.yarnrc.yml` as a workaround:

```yaml
packageExtensions:
  "antd@*":
    exports:
      "./es/*": "./es/*"
      "./lib/*": "./lib/*"
```

**Test**

Added `tests/dekko/exports.test.ts`, which verifies deep subpath type imports resolve correctly under `moduleResolution: "bundler"` in a temporary project:
- ✅ Passes when `exports` field is absent
- ❌ Fails when the broken `exports` from #57930 is present, reporting `TS2307: Cannot find module 'antd/es/theme/interface'`

### 📝 Changelog

| Language | Description |
| -------- | ----------- |
[...]

PR #58775, opened 2026-07-22
fix: support Yarn PnP subpath imports

### LINKED ISSUE #56858: Missing exports field in package.json causes Yarn PnP resolution failures for subpath imports

drewmt:
### 🤔 This is a ...

- [x] 🐞 Bug fix
- [x] ✅ Test Case

### 🔗 Related Issues

Fixes #56858

This supersedes #58774. That PR was automatically closed because its first revision added a file under the team-maintained `scripts/` path. This version keeps the generator in the existing root build configuration and contains no restricted-path changes.

### 💡 Background and Solution

Yarn PnP cannot use legacy extension and directory lookup for deep imports such as `antd/es/input/Search`. A simple `exports` wildcard fixes that lookup, but breaks TypeScript because the compiled output mixes standalone files with directories containing `index.d.ts`.

This change adds conditional package exports and generates exact export entries for directory-based modules after compilation. Normal ESM imports of explicit `antd/es/*` paths keep using the ES build, CommonJS `require` keeps using the `lib` build, and the generic Node condition remains a CommonJS fallback.

The package checks cover file-based and directory-based subpaths under TypeScript `bundler` and `node16` resolution, CommonJS runtime imports, normal Node ESM resolution, browser ESM resolution, and generated export-map freshness.

Validation performed:

- `npm run compile`
- `./node_modules/.bin/antd-tools run dist`
- `npm run test:dekko`
- `npm run tsc`
- scoped ESLint and Biome checks
- a packed artifact containing the existing 3,972-file package surface

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Yarn PnP resolution for Ant Design subpath imports without breaking TypeScript type resolution. |
| 🇨🇳 Chinese | 修复 Yarn PnP 下 Ant Design 子路径导入解析，同时保持 TypeScript 类型解析兼容性。 |


QDyanbing:
Compared with #57930 and #57968, the generated proxies cover the previous TypeScript resolution regression. However, this PR introduces three other issues:

1. For `./es/*`, `node` is matched before `import`, so Node ESM resolves explicit `antd/es/*` imports to `lib/*`. This also changes the default-export shape. The current `require.resolve` test does not cover runtime imports.

2. The generated proxies add 676 files, increasing the package file count from 3972 to 4648 (+17%). This is the package-layout overhead previously discussed in #57930.

3. `BUG_VERSIONS.json` is included in the published package but missing from `exports`, making `antd/BUG_VERSIONS.json` inaccessible.

afc163 (changes requested):
The generated export map still orders `node` before `import` for `./es/*`. In normal Node ESM resolution both conditions are active, so imports such as `antd/es/input` resolve to `lib/input/index.js`, changing the explicit ES entry and its default-export shape. The current ESM test forces the `browser` condition and therefore misses this path. Please put the ESM condition before `node` (while preserving `require` for CJS) and add a normal Node ESM resolution/runtime test.

drewmt:
Thanks for catching this. I reordered import and require ahead of node for the explicit ./es entries, so normal Node ESM now keeps the ES target while CommonJS keeps lib. I also added a plain Node ESM resolution test without the browser condition; the full Dekko suite, TypeScript, and scoped lint checks pass locally.

nrps9909 (approved):
Reviewed exact head `ae6474c54a725bf9dec81cc99512ddde4c17b99d`.

The three blockers documented on the earlier head are resolved:

- `./es/*` entries now place `import` before the generic `node` condition, while `require` still targets `lib/*`; the committed test also covers ordinary Node ESM without forcing the browser condition.
- The implementation generates exact export-map entries without adding proxy files. I downloaded the Continuous Releases artifact identified by `x-commit-key: ant-design:ant-design:ae6474c...`; it contains the original 3,972-file package surface rather than the previous 4,648-file proxy layout.
- `./BUG_VERSIONS.json` is exported and covered by the package test.

Independent exact-artifact checks:

- Node ESM resolved `antd/es`, `antd/es/input`, and `antd/es/input/Search` to their `/es/...` files.
- CommonJS resolved both `antd/es/input` and `antd/lib/input` to `lib/input/index.js`, and resolved `antd/BUG_VERSIONS.json` plus `antd/package.json`.
- Node with `--conditions=browser` kept the two explicit ES subpaths on `/es/...`.
- TypeScript's resolution API found `antd/es/theme/interface`, `antd/es/button`, and `antd/es/input/Search` under both `bundler` and `node16`, each at the intended `.d.ts` path.
- An independent export-map audit found 169 compiled directory entrypoints, zero missing ES/lib entries, 358 total exports, and the explicit BUG_VERSIONS mapping.
- All current remote checks are green, including build, test-node, lib/es modules, legacy/latest React matrices, dist matrices, lint, coverage, size, and visual regression.

A local source compile with today's shared dependency tree reached an unrelated historical-branch type drift at `components/menu/SubMenu.tsx` (`rc-menu` now narrows `title`), so I did not use that environment mismatch as package-export evidence; the checks above use the exact built artifact associated with this head.

I found no remaining blocker in this change.

AI assistance disclosure: Codex helped prepare the isolated worktree and artifact-resolution matrix; I reviewed the diff, exact artifact identity, command outputs, and conclusion before submitting this review.

PR #58818, opened 2026-07-28
fix: make Table and Sider triggers keyboard accessible

afc163:
### 🤔 这个变动的性质是？

- [ ] 🆕 新特性提交
- [x] 🐞 Bug 修复
- [ ] 📝 站点、文档改进
- [ ] 📽️ 演示代码改进
- [ ] 💄 组件样式/交互改进
- [ ] 🤖 TypeScript 定义更新
- [ ] 📦 包体积优化
- [ ] ⚡️ 性能优化
- [ ] ⭐️ 功能增强
- [ ] 🌐 国际化改进
- [ ] 🛠 重构
- [ ] 🎨 代码风格优化
- [ ] ✅ 测试用例
- [ ] 🔀 分支合并
- [ ] ⏩ 工作流程
- [x] ⌨️ 无障碍改进
- [ ] ❓ 其他改动（是关于什么的改动？）

### 🔗 相关 Issue

Related to #37508.

### 💡 需求背景和解决方案

Table 的筛选图标和 Layout.Sider 的默认收起图标无法通过 Tab 聚焦，键盘用户也不能用 Enter 或空格触发。

[CodeSandbox 复现](https://codesandbox.io/s/39tymf)

1. 点击 “Check focusability”，可以看到两个触发器都不是 `tabIndex=0`。
2. 再用 Tab 浏览页面，焦点会跳过它们。

本次为 antd 自己渲染的触发器补充焦点、键盘操作和 ARIA 状态。自定义交互图标仍由使用方自行处理，避免一次按键触发两次；rc-dropdown 内部焦点管理不在本 PR 范围内。

### 📝 更新日志

| 语言    | 更新描述 |
| ------- | -------- |
| 🇺🇸 英文 | Improve keyboard access for Table filter and Layout.Sider triggers. |
| 🇨🇳 中文 | 改进 Table 筛选和 Layout.Sider 收起触发器的键盘操作。 |

### DIFF
--- a/components/table/hooks/useFilter/FilterDropdown.tsx
+++ b/components/table/hooks/useFilter/FilterDropdown.tsx
@@ -6,7 +6,7 @@ import { clsx } from 'clsx';
 
 import type { FilterState } from '.';
 import { useSyncState } from '../../../_util/hooks';
-import { isFunction, isNumber } from '../../../_util/is';
+import { isFunction, isNumber, isString } from '../../../_util/is';
 import type { AnyObject } from '../../../_util/type';
 import { devUseWarning } from '../../../_util/warning';
 import Button from '../../../button/Button';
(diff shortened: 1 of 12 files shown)


afc163:
Current head still has two Table accessibility/behavior blockers. Wrapping every custom filterIcon with role=button and tabIndex=0 creates nested interactive controls and two tab stops when the custom icon is already a button; the custom-trigger path should remain untouched, as in the Sider change. Also, the keyboard handler calls triggerVisible directly and bypasses onVisibleChange, so controlled filteredKeys synchronization and filterOnClose confirmation differ from the mouse path. Please route keyboard activation through the same public visibility path and update the now-failing snapshots.

components/table/hooks/useFilter/FilterDropdown.tsx
```diff
@@ -549,14 +549,35 @@ const FilterDropdown = <RecordType extends AnyObject = AnyObject>(
       filterIcon = <FilterFilled />;
     }
 
+    const isInteractiveFilterIcon =
+      React.isValidElement<React.HTMLAttributes<HTMLElement> & { href?: string }>(filterIcon) &&
+      ((isString(filterIcon.type) &&
```
zombieJ: 这个写法很怪异，对于封装组件应该是没用的。比如这里使用 antd 的 Button 也无法命中 `button`

afc163: 确实，ReactElement.type 只能可靠识别原生标签，无法判断封装组件最终渲染出的 DOM。已在 431d669888 中移除这段检测，让 filterIcon 保持图标内容语义，由外层 trigger 统一承担键盘交互。

nrps9909:
One related localization/accessibility gap surfaced while tracing this trigger: `TableLocale.filterTitle` already exists (`Filter menu` in en-US and `篩選器` in zh-TW), but `FilterDropdown` does not currently use it. As a result, the wrapper's accessible name comes from `FilterFilled`'s hard-coded English `aria-label="filter"`; a custom unnamed `filterIcon` may leave the trigger without a localized name.

Since this PR already updates the same trigger and its snapshots, would it make sense to add `aria-label={locale.filterTitle}` here and cover it with a localized/custom-icon assertion? I did not open a parallel PR because it would overlap this work. If you prefer to keep the scope separate, I can re-check after this lands and prepare a follow-up.

PR #58834, opened 2026-07-29
fix(Anchor): avoid duplicate onChange with getCurrentAnchor

nikzanda:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

- Follow-up to #43916, which enabled `getCurrentAnchor` during scrolling and documented the mismatched active-link comparison.
- Preserves the click and `onChange` behavior introduced for #30584 by #30601.

### 💡 Background and Solution

When `getCurrentAnchor` transforms the link detected during scrolling, `activeLinkRef` stores the transformed value while `setCurrentActiveLink` compares it with the original link. If the two values differ, every scroll event is treated as an active-link change and triggers a duplicate `onChange` callback even though the effective highlighted anchor has not changed.

This change resolves the transformed link before comparing it with `activeLinkRef`. Repeated scroll events are ignored when the effective active link remains the same.

The existing click behavior is preserved: selecting a different original link still triggers `onChange` with that original link, including when `getCurrentAnchor` maps it to the currently highlighted anchor.

A regression test verifies that repeated scroll events continue to call `getCurrentAnchor` without emitting duplicate `onChange` callbacks. The existing test for the click behavior remains unchanged and passes.

This change does not introduce or modify any public API and has no visual impact.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Anchor triggering duplicate `onChange` callbacks on scroll when using `getCurrentAnchor`. |
| 🇨🇳 Chinese | 修复 Anchor 使用 `getCurrentAnchor` 时滚动会重复触发 `onChange` 的问题。 |


nikzanda:
@afc163  CI note: I checked the full logs for both failing jobs.

The complete Anchor test suite passes in both `test-react-latest (dom, 1/2)` and `test-react-legacy (18, 1/2)`, including the two regression tests added by this PR:

- `should not trigger onChange repeatedly when scrolling with getCurrentAnchor`
- `should apply getCurrentAnchor once when it changes`

The only failure is unrelated to Anchor. In both jobs, `components/tree-select/__tests__/demo-extend.test.ts` captures an asynchronous `rc-virtual-list` warning instead of the expected empty error list:

> Warning: Seems `scrollTo` with `rc-virtual-list` reach the max limitation. Please fire issue for us. Thanks.

This PR changes only `components/anchor/Anchor.tsx` and its test file; there are no TreeSelect changes. The same two matrix jobs also fail on the base commit `5bf6ee53`: [React latest](https://github.com/ant-design/ant-design/actions/runs/30612896888/job/91099491418) and [React 18](https://github.com/ant-design/ant-design/actions/runs/30612896888/job/91099490926).

The failing TreeSelect test also passes locally with `CI=1` (2/2 snapshots). I therefore did not update unrelated TreeSelect code or snapshots in this Anchor-only PR. Could the two failed jobs be rerun?

nrps9909 (changes requested):
Thanks for addressing the duplicate scroll callbacks. One raw-vs-mapped click-identity regression still breaks the behavior described in this PR.

Reproduction:
1. Configure `getCurrentAnchor` so `#raw-a` maps to `#mapped-b`.
2. Click `#raw-a`. Anchor highlights `#mapped-b`, and `onChange` receives `#raw-a`.
3. Clear the callback spy and click the distinct original link `#mapped-b`.

Expected: `onChange("#mapped-b")` is called exactly once because the user selected a different original link.

Actual: it is called 0 times.

`handleScrollTo` captures `previousActiveLink` from `activeLinkRef`, which stores the mapped/highlighted link, and compares it with the raw clicked `link`. Here both strings are `#mapped-b`, so `forceTriggerChange` is false; `setCurrentActiveLink` then sees the same mapped link and returns before `onChange`. The animation same-link guard also makes this mapped-vs-raw comparison.

Please preserve raw click identity separately: capture `rawActiveLinkRef.current` before calling `setCurrentActiveLink` and compare raw-to-raw when deciding whether this is the same click/target. `activeLinkRef` can remain responsible for deduplicating mapped visible state. Please add a regression for raw A -> mapped B, followed by clicking raw B, asserting exactly one `onChange` call for raw B.

Codex-assisted validation at head `ae3ea32103d1c3a1ca506454d0a909d4766ea71a` in a fresh detached checkout: the regression probe failed consistently (expected 1 call, received 0); the focused Anchor unit run passed 92 tests and 41 snapshots; changed-file ESLint, Biome, Prettier, and `git diff --check` passed. Same-head GitHub CI is green. Codex was used to inspect the implementation, design and run the probe, and draft this review.

nikzanda:
@nrps9909 Thanks for the detailed reproduction, you were right.

Fixed in [`63bde83`](https://github.com/ant-design/ant-design/commit/63bde836ecaa064229bea5f0eeacc94a31266c77). `handleScrollTo` now captures `rawActiveLinkRef.current` before calling `setCurrentActiveLink` and uses raw-to-raw comparisons both when deciding whether to force `onChange` and in the animated-scroll same-target guard. `activeLinkRef` remains responsible for deduplicating the mapped highlighted state.

I also added the requested regression test for raw A → mapped B, followed by clicking raw B. It verifies that the first click highlights mapped B and calls `onChange(raw A)`, while the second distinct raw click calls `onChange(raw B)` exactly once.

Could you please take another look?

nrps9909 (approved):
Verified the requested raw-vs-mapped click-identity fix on exact head `63bde836ecaa064229bea5f0eeacc94a31266c77`. `handleScrollTo` now captures `rawActiveLinkRef.current` before updating it and uses raw-to-raw comparisons for both forced `onChange` delivery and the animated-scroll same-target guard. The new raw A -> mapped B -> click raw B regression passes and asserts exactly one callback for raw B.

Fresh detached validation: full Anchor unit scope passed 7 suites, 93 tests, and 41 snapshots; changed-file ESLint, Biome, Prettier, and `git diff --check` passed. Repository TypeScript reported only the known unrelated Table demo token and shared Puppeteer type mismatches; same-head GitHub CI is green.

Codex-assisted review: Codex inspected the updated state paths, reran the original reproducer and related suite, and drafted this follow-up.

yoyo837 (commented):
Thanks for the fix — LGTM. One minor nit: in `Anchor.test.tsx`, the `Demo` component uses an inline props type (`React.FC<{ mappedLink: string }>`). The repo guidelines prefer a named interface; please use `interface DemoProps { mappedLink: string }` and `React.FC<DemoProps>`.

nrps9909 (approved):
Codex-assisted revalidation on a4234a37f20d45b0b082d1217dd211da4967ee94: the commits after the previously approved head only extract the test fixture props into DemoProps and merge the latest master; the Anchor runtime fix and raw-versus-mapped regression coverage are unchanged. I reran the complete Anchor scope (7 suites, 93 tests, 41 snapshots), plus target ESLint, Biome, Prettier, and git diff --check; all passed. The current remote matrix is also green (37 successful checks, no failures or pending jobs). No blocking findings.

yoyo837 (approved):

PR #58860, opened 2026-07-31
feat(Table): support force rendering expanded rows

### LINKED ISSUE #54798: Table expandable `forceRender` prop

Inside the expandable table, some required form inputs are rendered. If I submit the form while some rows are not expanded, validation is not triggered for fields that are not yet rendered. I know that the `defaultExpandAllRows` prop could solve this problem, but since my table may contain many rows, I don’t want to expand them all immediately.  
In general, I think this could be a useful feature, especially considering that `forceRender` exists in other components like Tabs, Modal, and Collapse.

```tsx
expandable={{
   forceRender: true
   // ...
}}
```

nikzanda:

### 🤔 This is a ...

- [x] 🆕 New feature
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

- Closes #54798
- Upstream implementation: react-component/table#1503
- Released in [`@rc-component/table@1.11.0`](https://github.com/react-component/table/releases/tag/%40rc-component%2Ftable%401.11.0)

### 💡 Background and Solution

Expanded row content is normally mounted only after its first expansion. As a result, fields rendered inside `expandedRowRender` are not registered by Form during initial validation if their rows have never been expanded.

This PR upgrades `@rc-component/table` to `1.11.0` and exposes the new `expandable.forceRender` option:

```tsx
<Table
  expandable={{
    forceRender: true,
    expandedRowRender: (record) => <ExpandedContent record={record} />,
  }}
/>
```

When enabled, expanded row content is mounted before the first expansion and kept hidden while collapsed. The default remains `false`, so existing behavior is unchanged.

The option works with regular and virtual tables. In virtual mode, only rows currently mounted by the virtual list are force-rendered; off-screen rows may still be unmounted.

### 📝 Change Log

| Language   | Changelog                                                                                                   |
| ---------- | ----------------------------------------------------------------------------------------------------------- |
| 🇺🇸 English | 🆕 Add Table `expandable.forceRender` to render expanded row content before the first expansion.            |
| 🇨🇳 Chinese | 🆕 新增 Table `expandable.forceRender`，支持在首次展开前渲染展开行内容。                                     |


yoyo837:
Thanks for adding `expandable.forceRender`. Could you please add a corresponding demo to the Table documentation? It would help users understand the intended usage and make the new behavior easier to verify, especially for Form fields rendered inside `expandedRowRender`.

nikzanda:
> Thanks for adding `expandable.forceRender`. Could you please add a corresponding demo to the Table documentation? It would help users understand the intended usage and make the new behavior easier to verify, especially for Form fields rendered inside `expandedRowRender`.

I’ve added a dedicated `forceRender` demo to the Table documentation.
The demo renders required `Form.Item` fields inside `expandedRowRender` and allows users to validate them while all rows are still collapsed. Users can then expand each row to inspect and complete the registered fields.

zombieJ:
Thanks for the contribution and for following up with the demo! The `forceRender` API itself looks good. I’m going to make a few adjustments on my side. In particular, I don’t think we need a dedicated demo for `forceRender`; documenting it in the API table should be sufficient.

To avoid conflicts, no further changes are needed from your side for now—I’ll take care of the remaining adjustments.

zombieJ (approved):

PR #58879, opened 2026-08-03
fix(Checkbox): filter out Checkbox.Group options with nullish values

GuillaumeZ2SS:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

None, to my knowledge

### 💡 Background and Solution

When creating an option with an `undefined` value in a Checkbox.Group, typescript doesn't complain and javascript throws a runtime exception : `TypeError: can't access property "toString", e.value is undefined`.
The issue is in the definition of the React key for the option. The option value is stringified using `toString()`. While this works for most values, some value don't have a `toString()` function, notably `undefined` and `null`.

Solution : Use `String` function instead of call to `toString()`, which has fewer chance of failing.
This enables the usage of `undefined` and `null` as option values.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | fix(Checkbox): filter out Checkbox.Group options with nullish values |
| 🇨🇳 Chinese | 修复 Checkbox.Group 在给出未定义或空选项值时会报错 |


components/checkbox/__tests__/group.test.tsx
```diff
@@ -237,6 +237,27 @@ describe('CheckboxGroup', () => {
...
   });
 
+  it('should support undefined and null as option value', () => {
+    const onChange = jest.fn();
+
+    const { container } = render(
+      <Checkbox.Group
+        options={[
+          { value: undefined, label: 'undefined' },
+          { value: null, label: 'null' },
```
QDyanbing, line 246: 感觉没啥必要，不是合理的option

GuillaumeZ2SS: If we agree on the base issue, I could try to implement an other solution for it :
* Either update typescript type to not allow `undefined` and `null` as values
* Or filter out options having `undefined` or `null` as value
* Any other fix you might think of

components/checkbox/Group.tsx
```diff
@@ -124,7 +124,7 @@ const CheckboxGroup = React.forwardRef(
       ? memoizedOptions.map<React.ReactNode>((option) => (
           <Checkbox
             prefixCls={prefixCls}
-            key={option.value.toString()}
+            key={String(option.value)}
```
li-jia-nan:
```ts
const mergedKey = isNonNullable(option.value) ? option.value.toString() : `checkbox-${index}`;
```

GuillaumeZ2SS: I think this a great way to handle the issue while addressing the point made by coderabbit.
I implemented it in my latest commit.
Thank you for your help !

li-jia-nan:
@GuillaumeZ2SS After giving this some more thought, I believe options with `undefined` or `null` values should be filtered out rather than preserved, as retaining them could introduce unnecessary edge cases and unexpected behavior.

li-jia-nan (approved):

PR #58932, opened 2026-08-10
fix(Select): keep customize input aligned on clear hover

### LINKED ISSUE #58931: AutoComplete with customize input: hover shifts `.ant-select-content` and misaligns the custom input's border
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/devbox/suspicious-feather-dhcyh3?workspaceId=ws_UJWDJFW6u5eT5bSdmMxLcC)

### Steps to reproduce

1. Render an `AutoComplete` with `allowClear` and a customize input as its child (see reproduction):
   <AutoComplete allowClear options={[{ value: 'foo' }]} style={{ width: 300 }}>
     <Input.TextArea autoSize={{ minRows: 3 }} />
   </AutoComplete>
2. Move the mouse over the field.
3. Watch the right edge of the inner input.

Not specific to `Input.TextArea` — a plain `<Input />` child shows the same thing.
The trigger is `allowClear`   customize input.

### What is expected?

The geometry stays stable on hover. The customize input owns the border and padding, so
its border should stay flush with the select box — which is exactly what the customize
style block already declares (`border: 0`, `padding: 0`, `-content { margin: 0 }`).

On hover, `.ant-select-content` receives `margin-inline-end: 18px`, so the customize input
shrinks by 18px while the outer select box keeps its size. The inner border no longer lines
up with the select box — you see two edges of different width — plus a layout jump, and in a
TextArea the text reflows on mouseover.

On hover, `.ant-select-content` receives `margin-inline-end: 18px`, so the customize input
shrinks by 18px while the outer select box keeps its size. The inner border no longer lines
up with the select box — you see two edges of different width — plus a layout jump, and in a
TextArea the text reflows on mouseover.
Root cause: the hover rule in `components/select/style/index.ts` (`hoverShowClearStyle`)

```
  [`&${componentCls}-allow-clear:not(${componentCls}-show-arrow) ${componentCls}-content`]: {
    marginInlineEnd: token.showArrowPaddingInlineEnd,
  }
```

has no `:not(${componentCls}-customize)` guard, and at specificity (0,5,0) it also overrides
the reset in `components/select/style/select-input-customize.ts` (0,3,0):

```
  [`&${componentCls}-customize`]: {
    border: 0, padding: 0,
    [`${componentCls}-content`]: { margin: 0, padding: 0 },
  }
```

`AutoComplete` passes `suffixIcon: null`, so `ant-select-show-arrow` is never applied and the
`:not()` in the hover rule always matches. The same declaration is duplicated under
`@media (hover: none)`, so touch devices get the offset permanently.

| Environment | Info |
| --- | --- |
| antd | 6.6.0 |
| React | 19.2.8 |
| System | Windows 11 |
| Browser | Latest Chrome |

---
Suggested fix — either exclude the customize mode from `hoverShowClearStyle`:
  [`&${componentCls}-allow-clear:not(${componentCls}-show-arrow):not(${componentCls}-customize) ${componentCls}-content`]
or, if the space for the clear button should be reserved in customize mode as well, apply it
as a permanent `padding-inline-end` on `${componentCls}-input` instead of a hover-only margin
on the content. The latter also removes the layout jump on hover.
This looks like the same class of bug as #58669 (outer select still styling the box in
customize mode, same file) — there the fix was a transparent background for filled inputs.
Unrelated but noticed nearby: `components/select/style/select-input.ts` still filters on the
v5 class name in
  [`&-show-search:not(${componentCls}-customize-input):not(${componentCls}-disabled)`]
v6 only ever applies `${prefixCls}-customize` (see `components/auto-complete/AutoComplete.tsx`),
so this `:not()` is dead and the rule applies in customize mode too.
Current userland workaround:

```
  .ant-select.ant-select-customize.ant-select-allow-clear:not(.ant-select-show-arrow) {
    .ant-select-content,
    &:hover .ant-select-content { margin-inline-end: 0; }
    > .ant-select-content > .ant-select-input { padding-inline-end: 24px; }
  }
```

lazerg:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

fix #58931

### 💡 Background and Solution

`hoverShowClearStyle` reserves room for the clear icon by putting `margin-inline-end` on `-content`, but it never excluded customize mode. AutoComplete with a custom child input passes `suffixIcon: null`, so `-show-arrow` is absent and the rule matches; at (0,5,0) it also beats the `margin: 0` reset in `select-input-customize.ts` (0,3,0). Hovering an AutoComplete with `allowClear` and a custom input therefore shrinks the child by 18px while the outer box keeps its width, so the two borders stop lining up and the content jumps. The `@media (hover: none)` copy makes it permanent on touch.

Added `:not(-customize)` to that selector so the custom child keeps owning its own geometry, which is what the customize block already declares. Covered by a hover image test on a new debug demo, asserting the content margin stays `0px` while hovering.

The issue also reports `-customize-input` in `select-input.ts` as a dead v5 class. That one is still live: `@rc-component/select` applies it in `BaseSelect`, so the guard works and I left it alone.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix AutoComplete with custom input and `allowClear` shifting its content on hover. |
| 🇨🇳 Chinese | 🐞 修复 AutoComplete 自定义输入组件配合 `allowClear` 时悬浮导致内容偏移的问题。 |


components/auto-complete/__tests__/image.test.ts
```diff
@@ -1,6 +1,39 @@
...
+
+    return getComputedStyle(content).marginInlineEnd;
+  }, `${customizeClearSelector} .ant-select-content`);
+
+  expect(marginInlineEnd).toBe('0px');
+};
 
 describe('AutoComplete image', () => {
+  describe('customize input clear hover', () => {
+    imageTest(
```
QDyanbing, line 45: The dedicated hover `imageTest` is useful. However, `imageDemoTest("auto-complete")` also auto-discovers `customize-clear-debug.tsx` and produces three additional non-hover snapshots that are unrelated to this regression. Of the six new visual baselines, please keep the three hover baselines and add `customize-clear-debug.tsx` to the `skip` list used by `imageDemoTest` to exclude the three regular demo baselines.

components/auto-complete/__tests__/image.test.ts
```diff
@@ -1,6 +1,39 @@
...
+
+import imageTest, { imageDemoTest } from '../../../tests/shared/imageTest';
+import CustomizeClearDebug from '../demo/customize-clear-debug';
+
+const customizeClearDebugFilename = 'components/auto-complete/demo/customize-clear-debug.tsx';
+const customizeClearSelector = '.ant-select-customize.ant-select-allow-clear';
+
+const expectContentNotShifted = async (testPage: Page) => {
+  const marginInlineEnd = await testPage.evaluate((selector) => {
+    const content = document.querySelector<HTMLElement>(selector);
```
QDyanbing: Both `querySelector` and `hoverSelector` only target the first `<Input />` instance in the demo, so the second `<Input.TextArea />` hover case is not actually verified. Please keep the hover regression test, but use independent selectors in the same `beforeScreenshot` flow to hover and assert both instances; this does not require another set of screenshots.

lazerg: Good catch. `beforeScreenshot` now hovers each instance with its own selector, `:has(input.ant-select-input)` and `:has(textarea.ant-select-input)`, and asserts the content margin for both. It also asserts the hovered root matches `:hover`, so a selector that stops matching fails instead of passing silently. I checked it against the old CSS and the TextArea case alone fails with 18px (15px in compact). Same screenshot set as before.

components/auto-complete/__tests__/image.test.ts
lazerg: Added `customize-clear-debug.tsx` to the `imageDemoTest` skip list. The suite drops from 54 image tests to 51, so the 3 regular demo baselines are gone and only the 3 hover ones stay.

QDyanbing:
LGTM

MadCcc (approved):

QDyanbing (approved):

PR #58956, opened 2026-08-11
fix(table): localize row-selection accessible labels and make selection menu trigger keyboard accessible

### LINKED ISSUE #58954: Table rowSelection accessible names ignore ConfigProvider locale
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/sandbox/n6xlct?file=%2Fsrc%2Fmain.jsx)

### Steps to reproduce

1. Open the reproduction. It renders checkbox and radio Tables inside `<ConfigProvider locale={zhTW}>`.
2. Inspect the row-selection inputs with a browser accessibility tree or screen reader.
3. In the checkbox Table, note the header and row accessible names.
4. Select the row and inspect its accessible name again.
5. With `rowSelection.selections: true`, press Tab and try to reach/open the adjacent selection menu with Enter or Space.

The reproduction also prints the observed `aria-label` values below the Tables.

### What is expected?

Default row-selection controls should use accessible names from the active ConfigProvider locale.

For zh-TW, the header checkbox, row checkbox/radio, and selected-row state should be announced in Traditional Chinese. The custom-selection menu trigger should have its own localized name and be keyboard reachable. Existing `getTitleCheckboxProps` and `getCheckboxProps` overrides should continue to take precedence.

The default accessible names are hard-coded English even under `zhTW`:

- Header checkbox: `Custom selection` (or `Select all`)
- Unselected checkbox/radio: `Select row 1`
- Selected checkbox: `Row 1 selected`

When custom selections are enabled, `Custom selection` is assigned to the header checkbox, while the actual adjacent dropdown trigger is a hover-only `<span>` with no button role, tab stop, `aria-haspopup`, or localized name.

| Environment | Info |
| --- | --- |
| antd | 6.6.0 |
| React | 19.2.8 |
| System | macOS 26.5.2 |
| Browser | Chrome 151.0.7922.76 |

---
### Source

Ant Design 6.6.0 currently hard-codes these names in `useSelection.tsx`:

- [Header checkbox](https://github.com/ant-design/ant-design/blob/6.6.0/components/table/hooks/useSelection.tsx#L496-L516)
- [Radio row](https://github.com/ant-design/ant-design/blob/6.6.0/components/table/hooks/useSelection.tsx#L533-L555)
- [Checkbox row states](https://github.com/ant-design/ant-design/blob/6.6.0/components/table/hooks/useSelection.tsx#L560-L585)
- [Custom-selection trigger](https://github.com/ant-design/ant-design/blob/6.6.0/components/table/hooks/useSelection.tsx#L466-L472)

Consumers can work around individual labels through `getTitleCheckboxProps` and `getCheckboxProps`, but there is no localized default.

Because a complete fix may add public `TableLocale` fields and affect every bundled locale, I am opening this issue before a PR. Possible API directions for maintainer guidance:

1. Reuse string-valued `selectAll` for the header and add dedicated row templates such as `selectRow`, `selectedRow`, plus `selectionMenu`.
2. Add four explicit string-only accessibility fields if reusing the existing `ReactNode` field is undesirable.

The dropdown semantics can be implemented as a second phase if preferred, but it is included here because the current `Custom selection` label is attached to the wrong control.

I can implement the agreed direction with checkbox/radio, selected-state, locale fallback, runtime locale switch, custom override, keyboard, and snapshot coverage.

AI assistance disclosure: Codex assisted with source inspection, duplicate searches, reproduction assembly, and issue drafting. The reported behavior was verified against Ant Design 6.6.0 and current master.

waterWang:
## Summary

The Table row-selection controls (`Checkbox`, `Radio`) have hard-coded English accessible names (`"Select row 1"`, `"Row 1 selected"`, `"Custom selection"`, `"Select all"`) that ignore the locale set via `ConfigProvider`. This causes screen readers to announce in English even when `locale={zhTW}` (or any non-English locale) is configured.

### 1. New `TableLocale` fields (components/table/interface.ts)

Added four optional fields to `TableLocale`:

| Field | Default (en_US) | Purpose |
|-------|----------------|--------|
| `selectRow` | `"Select row ${index + 1}"` | Accessible name for an unselected row checkbox/radio |
| `selectedRow` | `"Row ${index + 1} selected"` | Accessible name for a selected row checkbox |
| `customSelection` | `"Custom selection"` | Accessible name for the header checkbox when custom selections are enabled |
| `selectionMenu` | `"Selection menu"` | Accessible label for the custom-selection dropdown trigger |

### 2. Locale-aware `aria-label` (components/table/hooks/useSelection.tsx)

- **Header checkbox**: Uses `tableLocale.customSelection` (when custom selections enabled) or `tableLocale.selectAll`, falling back to the existing hard-coded English.
- **Radio row**: Uses `tableLocale.selectRow` with `{index + 1}` template substitution.
- **Checkbox row (unselected)**: Uses `tableLocale.selectRow`.
- **Checkbox row (selected)**: Uses `tableLocale.selectedRow`.

Existing `getTitleCheckboxProps` / `getCheckboxProps` overrides still take precedence (they spread after `aria-label`).

### 3. Keyboard-accessible selection menu trigger

Changed the custom-selection dropdown trigger from a bare `<span>` to a `<button type="button">` with `aria-label`, `aria-haspopup="menu"`, and `trigger={['click', 'hover']}` so it is:
- Tab-focusable (native `<button>` behavior)
- Screen-reader discoverable (announces the localized `selectionMenu` name)
- Keyboard-activatable (Enter/Space click opens the dropdown)

### 4. Locale updates

- `en_US.ts` - English defaults
- `zh_CN.ts` - Simplified Chinese translations
- `zh_TW.ts` - Traditional Chinese translations

### 5. CSS (components/table/style/selection.ts)

Reset browser default button styles (border, background, padding) for the selection menu trigger to preserve the existing visual appearance.

Closes #58954

---

AI assistance disclosure: The bug was identified and the fix was implemented with AI assistance. The reported behavior was verified against Ant Design 6.6.0 and current master. Human review is requested.


nrps9909 (commented):
Thanks for picking up #58954. I checked this branch against the reported behavior and ran the focused validation locally. There are a few issues to address before this is ready:

1. **The current branch does not pass the focused checks.** ESLint reports three `no-template-curly-in-string` errors in `useSelection.tsx` for the literal `${index + 1}` tokens. `Table.rowSelection.test.tsx` also has three snapshot failures from the changed header label and trigger DOM. The PR currently adds no behavioral tests for the new locale API or keyboard interaction.

2. **Most bundled locales still fall back to English.** Only `en_US`, `zh_CN`, and `zh_TW` define the four new keys; the other 70 runtime locale packs do not. `AGENTS.md` / `CLAUDE.md` explicitly require all locale files to be updated together. Since this issue is specifically about non-English `ConfigProvider` locales, please either complete the locale rollout or agree on a different API/fallback design with a maintainer first.

3. **The header checkbox is still named for the wrong control.** Even when custom selections are present, that checkbox still executes the select-all action. It should retain select-all semantics; the adjacent menu button is the control that should use `selectionMenu`. I would remove `customSelection` rather than move the same semantic mismatch into a locale key.

4. **`tableLocale.selectAll as string` is not runtime-safe.** `TableLocale.selectAll` publicly accepts `ReactNode`; a type assertion does not turn an element or number into a valid `aria-label`. Please use a runtime string guard or introduce a dedicated string-only accessible-name key.

5. **Runtime locale changes can retain stale labels.** The memoized `transformColumns` dependency list does not include the newly consumed locale fields (nor `selectAll`). With stable table props, switching `ConfigProvider` locale can leave the previous accessible names in place.

6. **The translation placeholder should be data, not an expression.** A token such as `${index}` is easier to preserve and validate than the exact source expression `${index + 1}`. The code can substitute the already computed one-based index and retain an English fallback when a partial/custom locale omits the key.

Suggested focused coverage: checkbox and radio names before/after selection under `zh_TW`; the select-all checkbox versus selection-menu button names; runtime locale rerender; `getTitleCheckboxProps` / `getCheckboxProps` override precedence; and the menu trigger's keyboard/state semantics. The existing row-selection snapshots should also be updated once the final DOM/API shape is settled.

These are based on the current head `249ecf4`; happy to re-check after an update.

PR #58969, opened 2026-08-12
fix(Table): improve row selection accessibility

### LINKED ISSUE #58954: Table rowSelection accessible names ignore ConfigProvider locale

thlovey:
### 🤔 这个变动的性质是？

- [ ] 🆕 新特性提交
- [x] 🐞 Bug 修复
- [ ] 📝 站点、文档改进
- [ ] 📽️ 演示代码改进
- [ ] 💄 组件样式/交互改进
- [ ] 🤖 TypeScript 定义更新
- [ ] 📦 包体积优化
- [ ] ⚡️ 性能优化
- [ ] ⭐️ 功能增强
- [x] 🌐 国际化改进
- [ ] 🛠 重构
- [ ] 🎨 代码风格优化
- [ ] ✅ 测试用例
- [ ] 🔀 分支合并
- [ ] ⏩ 工作流程
- [x] ⌨️ 无障碍改进
- [ ] ❓ 其他改动（是关于什么的改动？）

### 🔗 相关 Issue

close #58954

### 💡 需求背景和解决方案

Table 行选择控件的默认无障碍名称使用硬编码英文，不会跟随 ConfigProvider 的语言设置；自定义选择菜单也无法通过键盘访问。

本次改动为表头、行选择及已选状态补充本地化无障碍名称，并同步所有语言包。自定义选择菜单改为语义化按钮，支持 Tab、Enter、Space、点击和悬停操作，同时保留现有自定义 aria-label 的覆盖能力。

### 📝 更新日志

| 语言    | 更新描述 |
| ------- | -------- |
| 🇺🇸 英文 | Table row selection now uses localized accessible names and supports keyboard access to the custom selection menu |
| 🇨🇳 中文 | Table 行选择控件现支持本地化无障碍名称，并可通过键盘访问自定义选择菜单 |


yoyo837:
只添加中英文，其他的留给native speaker来加。

QDyanbing:
也搞到feature吧

thlovey:
> 也搞到feature吧

hao

nrps9909 (changes requested):
**Blocking: the localized row labels remove the row identity.**

The existing behavior names body controls `Select row 1`, `Select row 2`, etc. This patch changes every unselected checkbox/radio to the same `Select row` label and every selected checkbox to the same `Row selected` label; the new tests explicitly assert those duplicates. In a table with several selection controls, a screen-reader user can no longer distinguish which row each control operates on from its accessible name. This also regresses the behavior described in #58954, which proposed localized row templates rather than dropping the index.

Please preserve the existing 1-based row identity through localizable templates, for example `Select row ${index}` / `Row ${index} selected` (and the corresponding translations), then interpolate `index + 1` at the call site. The checkbox, radio, selected-state, locale-switch, fallback, and custom `getCheckboxProps` tests should verify both localization and distinct row numbers. A placeholder-parity assertion for the new locale fields would also prevent one locale from silently dropping `${index}`.

I verified this directly against head `bd34834`: the base tests expect `Select row 1/2/3`, while the new tests at lines 156-158 and 178-180 expect three identical labels. The selection-menu button work is otherwise directionally sound.

Codex-assisted review: I traced the row-selection rendering and locale data flow and compared the feature-base behavior with this PR head in an isolated checkout. No repository files were changed.

components/table/hooks/useSelection.tsx
```diff
@@ -40,6 +40,30 @@ export const SELECTION_NONE = 'SELECT_NONE' as const;
...
+  const [open, setOpen] = React.useState(false);
+
+  return (
+    <Dropdown
+      menu={menu}
+      getPopupContainer={getPopupContainer}
+      trigger={['click', 'hover']}
+      onOpenChange={setOpen}
+    >
+      <button type="button" aria-label={label} aria-haspopup="menu" aria-expanded={open}>
```
zombieJ: 这里是可以不用 button 直接给 DornOutlined 加个 role 为 button 就行了。不用改 dom 结构

thlovey: ok  我改下

nrps9909 (changes requested):
Rechecked the force-pushed head `32254570453c8d823995161c701c817e4ef0cf58`. The original accessibility blocker remains unchanged after the feature-branch merge: `useSelection.tsx` still assigns the bare localized `selectRow` / `selectedRow` strings to every body control, and the focused tests still explicitly assert duplicate names (`Select row`, `Select row`, ... / `Row selected`, `Row selected`, ...).

Fresh exact-head validation ran the three relevant row-selection tests successfully; they pass because they encode those duplicate labels. The previous 1-based row identity is still lost, so screen-reader users cannot distinguish which row each default control operates on. Please preserve a localizable row-index placeholder and interpolate `index + 1`, with parity and checkbox/radio/selected-state coverage as requested in the earlier review.

AI assistance disclosure: Codex was used to check out the rewritten exact head, trace the current label call sites, rerun the focused tests, and draft this follow-up. I verified the files, commit, and results.

zombieJ (approved):

zombieJ:
> Fresh exact-head validation ran the three relevant row-selection tests successfully; they pass because they encode those duplicate labels. The previous 1-based row identity is still lost, so screen-reader users cannot distinguish which row each default control operates on. Please preserve a localizable row-index placeholder and interpolate `index + 1`, with parity and checkbox/radio/selected-state coverage as requested in the earlier review.

It's safe to merge since the row of the table strucure already has line info with a11y tool.

PR #58981, opened 2026-08-13
fix: preserve Form labelCol offset in vertical layout

### LINKED ISSUE #51630: vertical布局的Form的labelCol.offset无效
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/sandbox/antd-reproduction-template-forked-sdgg8r)

### Steps to reproduce

直接查看效果

### What is expected?

image

未对齐

| Environment | Info |
| --- | --- |
| antd | 5.22.1 |
| React | 18 |
| System | windows11 |
| Browser | chrome |

nikzanda:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

- Fix #51630
- Follow-up to #52257

### 💡 Background and Solution

In a vertical Form, the label wrapper is rendered as a Grid `Col`. Grid offsets set `margin-inline-start`, but the more specific vertical Form label rule applied the `verticalLabelMargin` shorthand, overriding the offset.

This change separates the vertical label margin from the other vertical label styles and applies it through a structural `:where()` selector with the same specificity as a Grid offset class. Since Grid styles are emitted after Form styles, `margin-inline-start` now wins while the remaining vertical label spacing is preserved.

The selectors are limited to the direct label wrapper so nested Form.Item instances are not affected by an ancestor's layout styles. Horizontal and inline layouts keep their existing margin behavior.

The regression test validates the generated CSS cascade rather than only checking Grid class names. It covers Form-level and Form.Item-level `labelCol`, fixed and responsive offsets, responsive `offset: 0`, `span: 24`, and preservation of `verticalLabelMargin`.

### 📝 Change Log

| Language   | Changelog                                                                     |
| ---------- | ----------------------------------------------------------------------------- |
| 🇺🇸 English | Fix Form vertical layout `labelCol.offset` being overridden by label margins. |
| 🇨🇳 Chinese | 修复 Form 垂直布局中标签外边距覆盖 `labelCol.offset` 的问题。                 |


nrps9909 (approved):
Codex-assisted review: I verified the CSS cascade against #51630 and found no blocking issues. The direct-child selectors keep nested Form.Item labels out of an ancestor layout rule, the per-item CSS variable preserves `verticalLabelMargin`, and the later equal-specificity Grid offset rule correctly restores `margin-inline-start` for fixed and responsive offsets.

Local verification on head `7caae18`: Form Jest suite 105 passed / 1 skipped with 7 snapshots, `npm run compileOnly`, target ESLint, target Biome, and `git diff --check`. The Form index suite is intentionally excluded by this repository’s Vitest manifest, so Jest is the applicable focused runner here.

afc163 (approved):

PR #58984, opened 2026-08-13
fix(Dropdown): auto-focus menu when opened to enable keyboard navigation (Fixes #58980)

### LINKED ISSUE #58980: Keyboard focus never enters an open Dropdown menu, so keyboard users cannot operate it (biggest of 12 conformance findings, shared ahead of publication)
### Reproduction link

[**Railing**](https://github.com/amansoomro062/railing) · the open source conformance harness that produced these results. The ~50-line antd adapter and the runner reproduce every finding from a clone.

### Steps to reproduce

1. Render a default Dropdown with a menu of items (no custom props, no hand-written ARIA).
2. Tab to the trigger and open it (click or Enter).
3. Press Down Arrow.

### What is expected?

Per the W3C APG menu button pattern (https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/#keyboardinteraction) and WCAG 2.1.1 Keyboard: when the menu opens, focus or aria-activedescendant moves into it, the first item becomes active, and arrow keys move between items.

Focus stays on the trigger, no aria-activedescendant is set, and Down Arrow has nothing to move between. A keyboard user cannot operate the Dropdown at all.

| Environment | Info |
| --- | --- |
| antd | 6.6.0 |
| React | 18.3.1 |
| System | macOS |
| Browser | Chromium via Playwright 1.62.1, verified by hand in Chrome |

---
These results come from a full accessibility conformance run against `antd@6.6.0`: the bug above is the worst of 12 findings, of which 2 are blockers (both share this one cause). Ant Design passes 52 of 64 checks overall; Tabs pass everything.

Related but distinct: #57766 covers focus handling of link items inside an already-focused menu. This report is about focus never entering the menu on open at all, so the fix surface is different.

The results are from [Railing](https://railing.dev), an open source project that tests React component libraries against the W3C ARIA Authoring Practices Guide and publishes the results. Ant Design is one of seven libraries measured, and you are hearing about the results before anything is public: nothing is published yet and nothing will be for at least fourteen days from this issue. If fourteen days is tight for a release train your size, just say so and we will happily extend it. And if a fix ships inside the window, the fixed score is what gets published.

How the measurement works: [railing.dev/method](https://railing.dev/method) · [scoring](https://railing.dev/scoring) · [disclosure policy](https://railing.dev/disclosure) · [source](https://github.com/amansoomro062/railing)

Everything is reproducible from a clone, and every run has a replayable Playwright trace:

```bash
pnpm --filter @railing-dev/adapter-antd run dev
npx @railing-dev/runner run --target antd --component menu \
  --base-url http://localhost:5185 --repeat 3
```

The full report is below: every check, the specification clause behind it, severities, and links to the complete adapter source showing exactly how components were mounted.

<details>
<summary>Full report</summary>

# Accessibility conformance results for Ant Design

You are hearing this before we publish anything. No score for Ant Design is on our index, and none will be for at least fourteen days, whatever this report says.

Railing runs component libraries against the W3C ARIA Authoring Practices Guide and publishes the results. Every check cites the clause it measures, every score names an exact version, and every result is run repeatedly and discarded if the answer changes.

## The short version

| Component | Score | Failing checks |
| --- | --- | --- |
| accordion | 84% | 3 |
| combobox | 93% | 1 |
| dialog | 91% | 2 |
| menu | 51% | 6 |
| tabs | 100% | none |

## Read this before the findings

We mounted your components like this: Ant Design, mounted with default configuration using only components the library exports and no hand-written ARIA. Ant Design has no single combobox component: AutoComplete is the closest match to the APG editable-combobox pattern and is what is measured here. Its accordion is Collapse, and its menu is Dropdown with a menu prop.

If that choice is the disagreement rather than the findings themselves, say so and we will publish your reasoning beside the score. It is a judgement call, not a measurement, and you are better placed to argue it than we are.

## The 12 findings

12 findings. 5 of them come from 2 underlying causes, grouped below.

### The trigger is not marked up as a menu button

_3 findings, one cause._ Dropdown renders its trigger without any of the three attributes that tell assistive technology a menu is attached and whether it is open. They are absent together, and adding them is a single change to the trigger rather than three separate ones.

#### The menu button does not declare that it opens a menu.

- **Check:** `menu.trigger-has-haspopup`, "The menu button declares that it opens a menu"
- **Component:** menu
- **Severity:** serious, meaning the task is completable but confusing, or state is not communicated
- **Why it matters:** aria-haspopup is what warns a screen reader user that activating this button opens something, rather than performing an action.
- **Expected:** aria-haspopup="menu"
- **We measured:** attribute absent
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/) · [WCAG 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html)
- **Reproduce:** `pnpm railing run --target antd --component menu --only menu.trigger-has-haspopup --base-url http://localhost:5185`

#### The collapsed menu button does not expose aria-expanded="false".

- **Check:** `menu.trigger-collapsed-by-default`, "aria-expanded is present and false before the menu opens"
- **Component:** menu
- **Severity:** moderate, meaning behaviour departs from the specified pattern; users are inconvenienced
- **Why it matters:** Without it the user is not told the menu is currently shut, only that something can be opened.
- **Expected:** aria-expanded="false"
- **We measured:** attribute absent
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/) · [WCAG 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html)
- **Reproduce:** `pnpm railing run --target antd --component menu --only menu.trigger-collapsed-by-default --base-url http://localhost:5185`

#### The menu button does not expose aria-expanded="true" while its menu is open.

- **Check:** `menu.expanded-state-communicated`, "aria-expanded becomes true when the menu opens"
- **Component:** menu
- **Severity:** serious, meaning the task is completable but confusing, or state is not communicated
- **Why it matters:** A sighted user sees the menu appear. Without the state change a screen reader user is never told it did.
- **Expected:** aria-expanded="true"
- **We measured:** attribute absent
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/) · [WCAG 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html)
- **Reproduce:** `pnpm railing run --target antd --component menu --only menu.expanded-state-communicated --base-url http://localhost:5185`

### Keyboard focus never enters the open menu

_2 findings, one cause._ The menu does open on click, so this is not a failure to open. Focus stays on the trigger and no aria-activedescendant is set, which leaves nothing for the arrow keys to move between. The second finding follows from the first: fix where focus goes on open and both should pass.

#### No menu item became active when the menu opened.

- **Check:** `menu.focus-enters-menu-on-open`, "Opening the menu moves to its first item"
- **Component:** menu
- **Severity:** blocker, meaning the task cannot be completed by that input method at all
- **Why it matters:** This is the difference between a menu and a dropdown. If nothing becomes active on open, a keyboard user is left with a visible menu they cannot enter.
- **Expected:** the first item is focused, or referenced by aria-activedescendant
- **We measured:** [data-testid="hr-trigger"], and no aria-activedescendant on the menu
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/#keyboardinteraction) · [WCAG 2.4.3 Focus Order](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html)
- **Reproduce:** `pnpm railing run --target antd --component menu --only menu.focus-enters-menu-on-open --base-url http://localhost:5185`

#### No item is active after pressing Down Arrow.

- **Check:** `menu.arrow-moves-between-items`, "Down Arrow moves to the next item"
- **Component:** menu
- **Severity:** blocker, meaning the task cannot be completed by that input method at all
- **Why it matters:** Without arrow navigation a keyboard user can reach the first item and nothing beyond it.
- **Expected:** an active menu item
- **We measured:** neither DOM focus on an item nor aria-activedescendant
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/#keyboardinteraction) · [WCAG 2.1.1 Keyboard](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html)
- **Reproduce:** `pnpm railing run --target antd --component menu --only menu.arrow-moves-between-items --base-url http://localhost:5185`

### The rest, each on its own

Not related to the above, or to each other.

#### One or more header controls are not contained in a heading.

- **Check:** `accordion.header-is-heading`, "Each header control sits inside a heading"
- **Component:** accordion
- **Severity:** moderate, meaning behaviour departs from the specified pattern; users are inconvenienced
- **Why it matters:** Screen reader users navigate documents by heading. If accordion headers are not headings, an entire page structure disappears from the shortcut that most users rely on to move around it.
- **Expected:** each control inside an h1–h6 or role=heading element
- **We measured:** hr-header-1, hr-header-2, hr-header-3 have no heading ancestor
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#wai-ariaroles,states,andproperties) · [WCAG 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html)
- **Reproduce:** `pnpm railing run --target antd --component accordion --only accordion.header-is-heading --base-url http://localhost:5185`

#### The header control does not reference its panel.

- **Check:** `accordion.trigger-controls-panel`, "The header control references its panel with aria-controls"
- **Component:** accordion
- **Severity:** moderate, meaning behaviour departs from the specified pattern; users are inconvenienced
- **Why it matters:** The relationship is what lets assistive technology move from the control to the content it reveals.
- **Expected:** aria-controls referencing the panel
- **We measured:** attribute absent
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#wai-ariaroles,states,andproperties) · [WCAG 1.3.1 Info and Relationships](https://www.w3.org/WAI/WCAG22/Understanding/info-and-relationships.html)
- **Reproduce:** `pnpm railing run --target antd --component accordion --only accordion.trigger-controls-panel --base-url http://localhost:5185`

#### Space did not expand the focused section.

- **Check:** `accordion.space-expands`, "Space expands the focused section"
- **Component:** accordion
- **Severity:** serious, meaning the task is completable but confusing, or state is not communicated
- **Why it matters:** The APG requires both Enter and Space on the header. Space is routinely missed by implementations that attach a click handler to a non-button element.
- **Expected:** aria-expanded="true" on the first header
- **We measured:** aria-expanded="false"
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/accordion/#keyboardinteraction) · [WCAG 2.1.1 Keyboard](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html)
- **Reproduce:** `pnpm railing run --target antd --component accordion --only accordion.space-expands --base-url http://localhost:5185`

#### Only 1 of 3 options is exposed at a time, and the list does not say how many there are.

- **Check:** `combobox.options-have-option-role`, "Each option exposes role=option"
- **Component:** combobox
- **Severity:** serious, meaning the task is completable but confusing, or state is not communicated
- **Why it matters:** Without the option role, a screen reader cannot count the choices, announce position within them, or let the user jump between them.
- **Expected:** either every option is exposed, or aria-setsize and aria-posinset describe the full set
- **We measured:** 1 option(s) present, aria-setsize=absent, aria-posinset=absent
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/#wai-ariaroles,states,andproperties) · [WCAG 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html)
- **Reproduce:** `pnpm railing run --target antd --component combobox --only combobox.options-have-option-role --base-url http://localhost:5185`

#### Focus left the dialog after 4 Tab presses.

- **Check:** `dialog.tab-never-leaves-dialog`, "Tab never leaves the dialog, even briefly"
- **Component:** dialog
- **Severity:** moderate, meaning behaviour departs from the specified pattern; users are inconvenienced
- **Why it matters:** A tab stop on nothing gives a keyboard user no feedback and no way to tell whether they have fallen out of the dialog. They have not, and can carry on, which is why this is not a blocker, but the APG asks Tab to cycle among the dialog's own elements.
- **Expected:** focus stays on the dialog's own elements throughout the cycle
- **We measured:** <body> (nothing focused)
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/#keyboardinteraction) · [WCAG 2.4.3 Focus Order](https://www.w3.org/WAI/WCAG22/Understanding/focus-order.html)
- **Reproduce:** `pnpm railing run --target antd --component dialog --only dialog.tab-never-leaves-dialog --base-url http://localhost:5185`

#### Background content is still exposed to assistive technology while the modal is open.

- **Check:** `dialog.background-inert`, "Content behind the dialog is hidden from assistive technology"
- **Component:** dialog
- **Severity:** serious, meaning the task is completable but confusing, or state is not communicated
- **Why it matters:** A modal is visually exclusive. If the page behind it is still in the accessibility tree, a screen reader user can read and operate content that a sighted user cannot reach.
- **Expected:** hr-outside-content hidden from the accessibility tree (aria-hidden, inert, or equivalent)
- **We measured:** hr-outside-content still present and not ignored
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/dialog-modal/) · [WCAG 4.1.2 Name, Role, Value](https://www.w3.org/WAI/WCAG22/Understanding/name-role-value.html)
- **Reproduce:** `pnpm railing run --target antd --component dialog --only dialog.background-inert --base-url http://localhost:5185`

#### The menu did not open when Down Arrow was pressed on the focused button.

- **Check:** `menu.opens-on-down-arrow`, "Down Arrow opens the menu"
- **Component:** menu
- **Severity:** serious, meaning the task is completable but confusing, or state is not communicated
- **Why it matters:** APG requires Down Arrow to open the menu and move to its first item, which is how most keyboard users reach it.
- **Expected:** menu visible
- **We measured:** menu absent or hidden after 2000ms
- **Measured against:** [APG pattern](https://www.w3.org/WAI/ARIA/apg/patterns/menu-button/#keyboardinteraction) · [WCAG 2.1.1 Keyboard](https://www.w3.org/WAI/WCAG22/Understanding/keyboard.html)
- **Reproduce:** `pnpm railing run --target antd --component menu --only menu.opens-on-down-arrow --base-url http://localhost:5185`

## How we tested it

- **Versions:** antd@6.6.0
- **Browser:** chromium 151.0.7922.34
- **Configuration:** default. We use only what the library exports and never hand-write ARIA, even where the documentation instructs the developer to. If we did, the score would measure how carefully we copied your docs rather than what ships in the box.
- **Traces:** every run above has a replayable Playwright trace, captured from exactly the run that was scored. Ask and we will send them, or regenerate them with the command below.

Reproduce any of it from a clone of the repository:

```bash
pnpm --filter @railing-dev/adapter-antd run dev
pnpm railing run --target antd --component <component> \
  --base-url http://localhost:5185 --repeat 3
```

## Where we might be wrong

The most likely cause of a wrong result is our adapter, not your library.

It has already happened to you. Your first run scored 27% on combobox and 42% on accordion. Almost all of that was our adapter, not your library: we required the listbox to be CSS-visible when yours is a zero-height virtual-scroll node a screen reader reaches perfectly well, and we were looking for v5 class names against v6. Both were corrected before this report, and the dialog finding was downgraded from blocker to moderate once we accepted that focus reaching <body> is a wasted stop rather than an escape.

So the adapter source is included below. Please tell us if we mounted your component in a way you would not recommend, or if a check misreads the specification. We will correct it and, if it has already been published, correct that too.

## Where to check us

- [How the measurement works](https://railing.dev/method), including why the runner never learns which library it is testing.
- [How scoring works](https://railing.dev/scoring), with every weighting and what is deliberately left out.
- [Our disclosure policy](https://railing.dev/disclosure), covering notice, right of reply, and conflicts of interest.
- [The decision log](https://railing.dev/decisions), which is largely a record of results we got wrong and corrected.
- [The source](https://github.com/amansoomro062/railing). Every adapter, spec and score, and the code that produced this report.

## What happens next

- You have **fourteen days** before anything about this library is published.
- If you reply, your response is published beside the score, in full and unedited.
- **If you ship a fix in that window, we publish the fixed score.** A finding that gets fixed before publication is the best outcome this project has, not a story we lost.
- If fourteen days is not enough for your release process, ask for more and you will get it.

## The adapter we used

This is the whole of the library-specific code. There is nothing else.

The five harness files are in the repo, about fifty lines each:
[accordion.tsx](https://github.com/amansoomro062/railing/blob/main/adapters/antd/src/harnesses/accordion.tsx) ·
[combobox.tsx](https://github.com/amansoomro062/railing/blob/main/adapters/antd/src/harnesses/combobox.tsx) ·
[dialog.tsx](https://github.com/amansoomro062/railing/blob/main/adapters/antd/src/harnesses/dialog.tsx) ·
[menu.tsx](https://github.com/amansoomro062/railing/blob/main/adapters/antd/src/harnesses/menu.tsx) ·
[tabs.tsx](https://github.com/amansoomro062/railing/blob/main/adapters/antd/src/harnesses/tabs.tsx)

</details>

waterWang:
## Problem

When a Dropdown menu is opened via click or Enter, focus stays on the trigger and the Down Arrow key has no effect. A keyboard user cannot operate the Dropdown at all.

## Root cause

The Dropdown component did not pass the `autoFocus` prop to the underlying `@rc-component/dropdown`. Without `autoFocus`, the `useAccessibility` hook in `@rc-component/dropdown` does not move focus into the menu overlay when the dropdown opens, and the Down Arrow key has nothing to move between.

## Fix

Pass `autoFocus={mergedOpen}` to `RcDropdown` when the dropdown is visible, unless the user has explicitly set `autoFocus` to a different value. This leverages the existing `autoFocus` mechanism in `@rc-component/dropdown`'s `useAccessibility` hook, which focuses the menu overlay when the dropdown opens. Once the menu has focus, the `@rc-component/menu`'s built-in keyboard handler takes over ArrowDown/ArrowUp navigation.

## Affected W3C compliance

- WCAG 2.1.1 Keyboard
- W3C APG Menu Button Pattern

Closes #58980

### DIFF
--- a/components/dropdown/dropdown.tsx
+++ b/components/dropdown/dropdown.tsx
@@ -357,6 +357,7 @@ const Dropdown: CompoundedComponent = React.forwardRef<HTMLElement, DropdownProp
       mouseEnterDelay={mouseEnterDelay}
       mouseLeaveDelay={mouseLeaveDelay}
       visible={mergedOpen}
+      autoFocus={props.autoFocus !== undefined ? props.autoFocus : mergedOpen}
       builtinPlacements={builtinPlacements}
       arrow={!!arrow}
       overlayClassName={overlayClassNameCustomized}


nrps9909 (changes requested):
**Blocking: the default hover trigger now steals focus.**

`trigger` is optional here, so the underlying Dropdown keeps its default hover behavior. With `autoFocus` defaulting to `mergedOpen`, every hover-opened Dropdown now enables the rc-dropdown accessibility effect and moves focus into the first menu item after the RAF delay. This can pull focus out of an unrelated input merely because the pointer crosses a Dropdown trigger.

I reproduced this on head `9c90ee0` with a focused text input next to a default Dropdown: after `mouseEnter` and flushing the existing fake timers, `document.activeElement` changed from the input to `.ant-dropdown-menu-item-active`. Removing this added line made the same regression test pass.

Please make the default autofocus conditional on a keyboard-triggered open rather than on visibility alone, while preserving explicit `autoFocus={true|false}`. Regression coverage should include at least:

- hover open preserves the previously focused element;
- Enter/keyboard open moves focus into the menu;
- an explicit `autoFocus={false}` remains authoritative.

Codex-assisted review: I traced the Ant and rc-dropdown focus paths and ran the focused red/green reproduction locally in an isolated checkout. No repository files were changed.

PR #58999, opened 2026-08-15
fix(DatePicker): hide decorative range separator

### LINKED ISSUE #58998: [DatePicker] RangePicker separator exposes hardcoded English accessible names
### Reproduction link

[https://ant.design/components/date-picker#components-date-picker-demo-range-picker](https://ant.design/components/date-picker#components-date-picker-demo-range-picker)

### Steps to reproduce

1. Open the official RangePicker demo.
2. Render `DatePicker.RangePicker` under a non-English `ConfigProvider` locale such as `zhTW`.
3. Inspect `.ant-picker-separator` and its default icon with accessibility tooling.
4. Optionally pass a custom separator with its own accessible name, for example `separator={<span aria-label="自訂範圍分隔符">→</span>}`.

### What is expected?

The default range separator should not expose hardcoded English in a non-English interface. If the default arrow is decorative, it should be hidden from the accessibility tree. If it is intended to convey a semantic relationship, that name should come from the active locale.

A custom separator should preserve its own semantics instead of being overridden by an English name on its parent.

`components/date-picker/generatePicker/generateRangePicker.tsx` always renders an outer `aria-label="to"`. On current master, `dom-accessibility-api` computes the wrapper name as `to` and the default icon name as `swap-right`. With the custom Traditional Chinese separator above, the wrapper still computes to `to` because the parent label overrides its descendant content.

This creates an English accessibility fallback in every non-English locale and prevents a custom separator from supplying its own accessible name.

| Environment | Info |
| --- | --- |
| antd | 6.6.0 / current master `8b4c973` |
| React | 19.2.8 |
| System | macOS |
| Validation | Focused Jest reproduction, 2/2 assertions passed |

---
I checked all current open PR changed files and found no PR touching `generateRangePicker.tsx` or `RangePicker.test.tsx`. I also searched the open issues and PRs for the hardcoded labels, RangePicker separator accessibility, and screen-reader wording, and found no duplicate.

This likely needs maintainer input on whether the default separator should be decorative (`aria-hidden`) or localized. A focused regression test should cover the default arrow and custom separator behavior without adding a locale API unless the semantic wording is intentionally exposed.

AI assistance disclosure: Codex was used to trace the source, audit open PR files and issues, and build the focused reproduction. The reported DOM and accessible-name behavior was verified against current master.

nrps9909:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

Fixes #58998.

### 💡 Background and Solution

`RangePicker` currently puts the hardcoded accessible name `to` on its separator wrapper. The default `SwapRightOutlined` child also exposes `swap-right`, so non-English screen readers receive English-only names for a decorative element. Because a parent `aria-label` takes precedence over descendant content, the wrapper also overrides an accessible name supplied by a custom `separator`.

This change:

- marks only the default separator as decorative with `aria-hidden="true"`;
- leaves a custom separator exposed without a parent label, preserving the custom node's own semantics;
- adds focused assertions for both paths and refreshes the directly affected Jest and Vitest snapshots;
- does not add or change any public API or locale key.

Validation:

- Jest related tests: 228 suites, 3,206 passed, 95 skipped, 3,596 snapshots passed;
- Vitest related tests: 10 files, 159 tests passed;
- `npm run compileOnly`;
- ESLint and Biome on the changed source and test files;
- `git diff --check`;
- full `tsc --noEmit` produces the same seven pre-existing Table demo/Puppeteer errors on an unchanged `master`, with no error from this patch.

AI assistance disclosure: Codex was used to trace the accessibility behavior, draft the focused implementation and tests, refresh affected snapshots, and run the validation commands. The behavior and results above are directly reproducible from this pull request.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | DatePicker RangePicker no longer exposes its decorative default separator to screen readers and preserves accessible names on custom separators. |
| 🇨🇳 Chinese | DatePicker RangePicker 的装饰性默认分隔符不再被屏幕阅读器朗读，并会保留自定义分隔符的无障碍名称。 |

### DIFF
--- a/components/date-picker/generatePicker/generateRangePicker.tsx
+++ b/components/date-picker/generatePicker/generateRangePicker.tsx
@@ -93,6 +93,7 @@ const generateRangePicker = <DateType extends AnyObject = AnyObject>(
     const rootPrefixCls = getPrefixCls();
 
     const mergedSeparator = separator ?? rangePicker?.separator;
+    const hasCustomSeparator = mergedSeparator !== undefined && mergedSeparator !== null;
 
     const [variant, enableVariantCls] = useVariant('rangePicker', customVariant, bordered);
 
@@ -162,8 +163,11 @@ const generateRangePicker = <DateType extends AnyObject = AnyObject>(
       <ContextIsolator space>
         <RCRangePicker<DateType>
           separator={
-            <span aria-label="to" className={`${prefixCls}-separator`}>
-              {mergedSeparator ?? <SwapRightOutlined />}
+            <span
+              aria-hidden={hasCustomSeparator ? undefined : true}
+              className={`${prefixCls}-separator`}
+            >
+              {hasCustomSeparator ? mergedSeparator : <SwapRightOutlined />}
             </span>
           }
           disabled={mergedDisabled}
--- a/components/date-picker/generatePicker/generateRangePicker.tsx
+++ b/components/date-picker/generatePicker/generateRangePicker.tsx
@@ -9,6 +9,7 @@ import { clsx } from 'clsx';
 
 import ContextIsolator from '../../_util/ContextIsolator';
 import { useAllowClear, useZIndex } from '../../_util/hooks';
+import { isNonNullable } from '../../_util/is';
 import { getMergedStatus, getStatusClassNames } from '../../_util/statusUtils';
 import type { AnyObject } from '../../_util/type';
 import { devUseWarning } from '../../_util/warning';
@@ -93,7 +94,7 @@ const generateRangePicker = <DateType extends AnyObject = AnyObject>(
     const rootPrefixCls = getPrefixCls();
 
     const mergedSeparator = separator ?? rangePicker?.separator;
-    const hasCustomSeparator = mergedSeparator !== undefined && mergedSeparator !== null;
+    const hasCustomSeparator = isNonNullable(mergedSeparator);
 
     const [variant, enableVariantCls] = useVariant('rangePicker', customVariant, bordered);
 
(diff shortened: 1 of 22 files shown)


components/date-picker/generatePicker/generateRangePicker.tsx
```diff
@@ -93,6 +93,7 @@ const generateRangePicker = <DateType extends AnyObject = AnyObject>(
     const rootPrefixCls = getPrefixCls();
 
     const mergedSeparator = separator ?? rangePicker?.separator;
+    const hasCustomSeparator = mergedSeparator !== undefined && mergedSeparator !== null;
```
QDyanbing: Could we use the shared isNonNullable helper here to follow the repository convention?

nrps9909: Updated to use `isNonNullable(mergedSeparator)` in 31dfbc1. The focused Jest suite (23 tests), ESLint, Biome, and diff check pass. The RangePicker suite is intentionally excluded from the Vitest manifest.

afc163 (approved):

PR #59072, opened 2026-08-23
fix(Skeleton): hide decorative title placeholder from assistive technology

giaBaoJS:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

No linked issue. The problem is recorded in the repository itself:

- `components/skeleton/__tests__/a11y.test.ts` carried `disabledRules: ['empty-heading']` with a `// waiting for fix` comment.
- `components/statistic/__tests__/a11y.test.ts` skipped `basic.tsx` with a `// wait for skeleton fix` comment.

### 💡 Background and Solution

`Skeleton`'s title placeholder renders an empty `<h3 class="ant-skeleton-title" />` (`components/skeleton/Title.tsx`). It is purely decorative, but assistive technology sees a real heading with no accessible name, so screen readers announce an empty heading and it shows up in the document outline. axe reports it as `empty-heading`, and both linters had to be silenced on that element (`/* eslint-disable jsx-a11y/heading-has-content */` and a `biome-ignore`).

The placeholder is now marked `aria-hidden`, matching how other decorative elements are handled in the library. Nothing about the rendered layout or the public API changes.

Because the placeholder is no longer exposed, the two suppressions above are removed and the checks now run for real:

- `components/skeleton/__tests__/a11y.test.ts`: `empty-heading` is enabled again (the `list.tsx` / `element.tsx` skips are unrelated — those demos fail on `button-name` from a `Switch` without a label).
- `components/statistic/__tests__/a11y.test.ts`: `basic.tsx` is no longer skipped, since its `<Statistic loading />` was the only reason it failed.

A direct regression test was added in `components/skeleton/__tests__/index.test.tsx`. Reverting only `Title.tsx` turns it red (`Expected: "true"  Received: null`) together with both a11y suites (6 failures, all `empty-heading`).

The `eslint-disable` is dropped because `jsx-a11y/heading-has-content` accepts elements hidden from screen readers. The `biome-ignore` is kept, since `lint/a11y/useHeadingContent` does not consider `aria-hidden`, and its reason was updated.

Snapshot updates are limited to the added `aria-hidden="true"` attribute — the diff on `*.snap` files contains 62 added lines and nothing else.

Verification:

- `npm test -- components/skeleton components/statistic components/list components/config-provider` → 47 suites passed, 801 tests passed, 547 snapshots passed.
- `npx vitest run components/skeleton/__tests__/index.test.tsx` → 32 passed.
- `npm run tsc`, `eslint`, `biome lint` → clean.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Skeleton exposing its title placeholder to screen readers as an empty heading. |
| 🇨🇳 Chinese | 🐞 修复 Skeleton 标题占位元素被屏幕阅读器识别为空标题的问题。 |

### DIFF
--- a/components/config-provider/__tests__/__snapshots__/components.test.tsx.snap
+++ b/components/config-provider/__tests__/__snapshots__/components.test.tsx.snap
@@ -23902,6 +23902,7 @@ exports[`ConfigProvider components Skeleton configProvider 1`] = `
     class="config-skeleton-section"
   >
     <h3
+      aria-hidden="true"
       class="config-skeleton-title"
       style="width: 50%;"
     />
@@ -23930,6 +23931,7 @@ exports[`ConfigProvider components Skeleton configProvider componentDisabled 1`]
     class="config-skeleton-section"
   >
     <h3
+      aria-hidden="true"
       class="config-skeleton-title"
       style="width: 50%;"
     />
@@ -23958,6 +23960,7 @@ exports[`ConfigProvider components Skeleton configProvider componentSize large 1
     class="config-skeleton-section"
   >
     <h3
+      aria-hidden="true"
       class="config-skeleton-title"
       style="width: 50%;"
     />
@@ -23986,6 +23989,7 @@ exports[`ConfigProvider components Skeleton configProvider componentSize medium 
     class="config-skeleton-section"
   >
     <h3
+      aria-hidden="true"
       class="config-skeleton-title"
       style="width: 50%;"
     />
@@ -24014,6 +24018,7 @@ exports[`ConfigProvider components Skeleton configProvider componentSize small 1
     class="config-skeleton-section"
   >
     <h3
+      aria-hidden="true"
       class="config-skeleton-title"
       style="width: 50%;"
     />
@@ -24042,6 +24047,7 @@ exports[`ConfigProvider components Skeleton normal 1`] = `
     class="ant-skeleton-section"
   >
     <h3
+      aria-hidden="true"
       class="ant-skeleton-title"
       style="width: 50%;"
     />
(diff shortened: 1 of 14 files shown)


nrps9909 (approved):
Reviewed exact head `c4eb744a2e7fb7aa86b1af3b00d2938c25fb2a27`.

The placeholder `<h3>` is always decorative and has no content API, so hiding it from the accessibility tree removes the false empty heading without suppressing user content. I verified:

- The complete local Skeleton + Statistic Jest scope passes: 12 suites, 117 tests, 88 snapshots.
- The applicable Skeleton Vitest file passes 32/32; the accessibility demo files are intentionally Jest-only in this configuration.
- An independent rerender probe passes under both Jest and Vitest: while loading, the physical placeholder remains with `aria-hidden="true"` and no heading role is exposed; after `loading={false}`, the placeholder is removed and a real child `<h2>` is exposed by role and name.
- Focused ESLint, Biome, and `git diff --check` pass.
- All completed remote exact-head checks are green, including the full 801-test submitted matrix, legacy/latest React, coverage, and visual regression.

The restored Skeleton and Statistic accessibility coverage directly exercises the prior `empty-heading` suppression. No blocking behavior or semantic regression found.

PR #59073, opened 2026-08-23
fix(Steps): mark the progress icon as a progressbar role

giaBaoJS:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

No linked issue. The problem is recorded in the repository itself: `components/steps/__tests__/a11y.test.ts` skipped `title-placement.tsx` and `progress.tsx` with a `// we can set aria attribute to fix it` comment.

### 💡 Background and Solution

When a `Steps` item is in the `process` state and `percent` is set, `ProgressIcon` renders:

```html
<svg class="ant-steps-item-progress-icon-svg" aria-valuemax="100" aria-valuemin="0" aria-valuenow="60">
  <title>Progress</title>
  ...
</svg>
```

The `aria-valuemin` / `aria-valuemax` / `aria-valuenow` attributes are only permitted on a role that supports a value (`progressbar`, `slider`, `spinbutton`, …). The element carries no `role`, so an `<svg>` maps to a role that does not allow them, and the attributes are dropped by assistive technology instead of announcing the step's progress. axe reports it as `aria-allowed-attr`.

Adding `role="progressbar"` makes the attributes valid and matches what the markup already intended — the `aria-value*` triplet and the `<title>Progress</title>` were both already there. The `<title>` supplies the accessible name, so `aria-progressbar-name` is satisfied too. Nothing about the rendered output or the public API changes.

With the element valid, `title-placement.tsx` and `progress.tsx` no longer need to be skipped in the accessibility suite. `inline.tsx` is still skipped, but for an unrelated reason — its demo avatar images have no `alt` text — and the comment has been corrected to say so.

A direct regression test was added in `components/steps/__tests__/index.test.tsx`. Reverting only `ProgressIcon.tsx` turns it red (`Expected: "progressbar"  Received: null`) along with the two re-enabled demo suites (2 failures, both `aria-allowed-attr`).

Snapshot updates are limited to the added `role="progressbar"` attribute — the diff on `*.snap` files contains 20 added lines and nothing else.

Verification:

- `npm test -- components/steps` → 7 suites passed, 85 tests passed, 64 snapshots passed.
- `npx vitest run components/steps/__tests__/index.test.tsx` → 9 passed.
- `npm run tsc`, `eslint`, `biome lint` → clean.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Steps progress icon using `aria-valuenow` without a `progressbar` role, so screen readers ignored the step progress. |
| 🇨🇳 Chinese | 🐞 修复 Steps 进度图标使用 `aria-valuenow` 但缺少 `progressbar` role，导致屏幕阅读器无法播报步骤进度的问题。 |


nrps9909 (changes requested):
Reviewed exact head `52ece294d820cffaf0dfa14efc314f1885da55bf`.

The role addition fixes the `aria-allowed-attr` violation, and the submitted verification is healthy: the complete Steps Jest scope passes (7 suites, 85 tests, 64 snapshots), the applicable Vitest file passes 9/9, focused ESLint/Biome/diff checks pass, and all completed remote checks are green.

There is one accessibility blocker before this should merge. Adding `role="progressbar"` makes the existing hardcoded SVG `<title>Progress</title>` the widget's active accessible name. An independent Jest/Vitest probe rendered Chinese item titles under `ConfigProvider locale={zhTW}`; `getByRole('progressbar')` still had the accessible name `Progress`, and no `進度`-named progressbar existed. There is also no Steps locale field or public accessible-name override for this internal SVG.

This therefore replaces an ignored value attribute with a correctly exposed but English-only widget name in every non-English locale. Please source the progressbar name from localized or caller-supplied content (with a supported override path), and add a non-English/override regression that asserts the actual accessible name—not only the role and value. The implementation choice may need maintainer input because Steps currently has no locale contract.

giaBaoJS:
The `Progress` string is real, but it is not something this PR introduced, and I removed it anyway.

On the origin: `<title>Progress</title>` sits at `components/steps/ProgressIcon.tsx:35` and landed in 8db39e34a7 (#53789, 2025-05-13). An SVG `<title>` child is the accessible name of the `<svg>` element under SVG-AAM regardless of role, so the name was already `Progress` before `role="progressbar"` existed. I checked this rather than assuming it: rendering Steps under `ConfigProvider locale={zhTW}` and calling `computeAccessibleName` on the SVG returns `Progress` both with the role attribute and with the role attribute stripped from the DOM node. So the effect of the role change was to move an already exposed English name from a graphic onto a widget, not to create one. The `<title>` was also rendering as a native English tooltip on hover in every locale, which is the same defect on the visual side.

On the fix: I did not add a Steps locale field. The repo already has two other progressbars, `components/progress/progress.tsx:344` and `components/spin/Indicator/Progress.tsx:73`, and neither exposes an accessible name; `components/progress/__tests__/a11y.test.ts:5` disables `aria-progressbar-name` for exactly that reason. Introducing a Steps locale contract here would be a public API decision that also leaves Progress inconsistent, and you are right that it needs maintainer input. Instead I deleted the hardcoded string, so Steps now matches the two existing progressbars: role plus `aria-valuemin`/`aria-valuemax`/`aria-valuenow`, no untranslatable name. `aria-progressbar-name` is a best-practice rule, not the WCAG failure this PR is fixing, and axe reports no violation for the Steps demos after the change.

On the regression test: `components/steps/__tests__/index.test.tsx` now renders Steps inside `ConfigProvider locale={zhTW}` with Chinese item titles and asserts the actual accessible name, not just the role and value. With `<title>Progress</title>` restored the assertion fails with `Expected element not to have accessible name: Received: Progress`. Full Steps Jest scope after the change: 7 suites, 86 tests, 64 snapshots, all passing.

If a maintainer would rather have a localized name than none, the natural shape is a `Steps` locale entry applied to Progress and Spin as well, and I am happy to do that in a separate PR.

nrps9909 (approved):
Re-reviewed exact head `1c01b4cb5dd36ad277605a5ca6262cf43b943a0a`. The previous blocker is resolved: the hardcoded English SVG title is gone, so a zh-TW Steps instance no longer exposes `Progress` as the progressbar name, while the explicit `progressbar` role keeps `aria-valuemin`, `aria-valuemax`, and `aria-valuenow` valid. The new regression asserts both the role/value contract and the absence of the untranslatable accessible name. A future localized naming contract can be designed consistently across Steps, Progress, and Spin without blocking this scoped standards fix.

Independent validation on this exact head:

- complete Steps Jest scope: 7 suites, 87 tests, 64 snapshots
- applicable Vitest scope: 11/11 tests
- focused ESLint, Biome, and diff validation passed
- all current repository CI, coverage, build, module, and visual checks are green

Approved.

afc163 (approved):

PR #59074, opened 2026-08-23
fix: keep transfer pagination on a valid page

biubiukam:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

> No related issue.

### 💡 Background and Solution

> When the filtered Transfer list becomes empty, the pagination effect can clamp the current page to `0`. If the filtered data is restored afterwards, the current page remains invalid and the list renders no items.
>
> Minimal reproduction (pseudo code):
>
> ```ts
> const items = [{ key: 'a' }, { key: 'b' }];
> renderTransfer({ items, pagination: { pageSize: 1 } });
> goToPage(2);
> updateFilteredItems([]);    // current becomes 0
> updateFilteredItems(items); // current is still 0
>
> // Before: visible items = []
> // After: current is clamped to 1, visible items = ['a']
> ```
>
> The page state is now clamped to the valid range `1..maxPageCount`, so restoring filtered data renders the first available page.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Transfer pagination staying on page 0 after filtered data is restored. |
| 🇨🇳 Chinese | 修复 Transfer 过滤数据恢复后分页停留在第 0 页的问题。 |

### DIFF
--- a/components/transfer/ListBody.tsx
+++ b/components/transfer/ListBody.tsx
@@ -71,7 +71,7 @@ const TransferListBody: React.ForwardRefRenderFunction<
   React.useEffect(() => {
     if (mergedPagination) {
       const maxPageCount = Math.ceil(filteredRenderItems.length / pageSize!);
-      setCurrent(Math.min(current, maxPageCount));
+      setCurrent(Math.max(1, Math.min(current, maxPageCount)));
     }
   }, [filteredRenderItems, mergedPagination, pageSize]);
 
--- a/components/transfer/__tests__/list-body.test.tsx
+++ b/components/transfer/__tests__/list-body.test.tsx
@@ -1,0 +1,58 @@
+import React from 'react';
+import { fireEvent, render, waitFor } from '@testing-library/react';
+
+import type { KeyWiseTransferItem } from '..';
+import ListBody from '../ListBody';
+import type { TransferListBodyProps } from '../ListBody';
+import type { RenderedItem } from '../Section';
+
+const items: RenderedItem<KeyWiseTransferItem>[] = [
+  {
+    item: { key: 'a', title: 'a' },
+    renderedEl: <span>a</span>,
+    renderedText: 'a',
+  },
+  {
+    item: { key: 'b', title: 'b' },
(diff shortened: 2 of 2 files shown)


nrps9909 (approved):

nrps9909 (commented):
Follow-up evidence for my approval of exact head `fa70bae5fbbdb002b77218aed8e500ce279f7d88` (the approval review was submitted without its intended body due to a CLI stdin mistake).

I verified the pagination clamp itself and the surrounding Transfer behavior:

- The submitted focused regression test passes in both Jest and Vitest (1/1 in each).
- An independent rerender probe passes in both runners (1/1 in each): start on page 3 with `pageSize=1`, change to `pageSize=2` and observe page 2/item C, then shrink to one item and observe page 1/item A.
- The broader non-environment-sensitive Transfer Jest scope passes: 11 suites, 116 tests, 22 snapshots.
- The full Transfer Vitest scope passes: 7 files, 30 tests.
- ESLint, Biome, and `git diff --check` pass for the changed files.
- All completed remote checks are green, including the full React/test matrix and coverage.

One local caveat: the Jest `demo-extend` snapshot suite differs only in generated `rc-menu` IDs (`rc-menu-uuid-test-id-*` vs `rc-menu-uuid-*`) under this checkout's installed dependency state; the PR's remote exact-head snapshot run is green, and the difference is unrelated to the Transfer pagination change.

The new `Math.max(1, Math.min(current, maxPageCount))` lower bound directly fixes the empty-data page-0 case while retaining the existing upper clamp. The CodeRabbit type-assertion/snapshot suggestions are non-blocking for this functional fix.

yoyo837 (approved):

PR #59092, opened 2026-08-25
fix(Select): keep single mode height independent of global lineHeight

### LINKED ISSUE #59091: 全局token 设置 lineHeight=2 时， Select 单选与多选  会出现高度不一样
### Reproduction link

[https://github.com/ant-design/ant-design/issues](https://github.com/ant-design/ant-design/issues)

### Steps to reproduce

全局token 设置 lineHeight=2 后， Select 单选会比多选 高4个像素， 
全局token lineHeight不设置，使用默认值， Select 单选会比多选是一样的高的。
全局token 设置 lineHeight=1 后， Select 单选会比少选 高6个像素，

### What is expected?

Select 组件不应该与全局token  lineHeight 有关联

全局token lineHeight 变化时，会影响 Select 单选

| Environment | Info |
| --- | --- |
| antd | 6.5.0 |
| React | 19.2.7 |
| System | win11 |
| Browser | chrome 140 |

lazerg:
### 🤔 This is a ...

- [x] 🐞 Bug fix
- [x] 📽️ Demo improvement

### 🔗 Related Issues

close #59091

### 💡 Background and Solution

Single mode renders its content with `var(--ant-select-line-height)`, but the vertical padding is `(height - font-height) / 2 - border-size`, and `font-height` is a px value derived when the theme is built. Overriding `lineHeight` moves one value and not the other, so the control grows past `controlHeight`. On master, single vs multiple measures 38 / 32 at `lineHeight: 2` and 24 / 32 at `lineHeight: 1`.

`--ant-select-font-height` now resolves to `calc(font-size * line-height)`, so the padding follows the configured line height instead of the one baked in at theme build time. The content keeps using the token, so overriding `lineHeight` still changes the selected text. The `font-height` override in `&-lg` is gone because it derives now, which also covers `lineHeightLG: 2`, where large stretched to 48px against a 40px control.

The padding is clamped at 0, since the recomputed value turns negative once `fontSize * lineHeight` passes the control height. In that case the line box no longer fits, so the control grows exactly as it does on master. `size="small"` at `lineHeight: 2` is one example: a 28px line box cannot sit inside a 24px control.

The default theme renders unchanged.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Select height changing when the `lineHeight` token is customized. |
| 🇨🇳 Chinese | 🐞 修复 Select 在自定义 `lineHeight` token 时高度变化的问题。 |

### DIFF
--- a/components/select/__tests__/image.test.ts
+++ b/components/select/__tests__/image.test.ts
@@ -8,8 +8,8 @@ import imageTest, { imageDemoTest } from '../../../tests/shared/imageTest';
 const clearSuffixDebugFilename = 'components/select/demo/clear-suffix-debug.tsx';
 const lineHeightDebugFilename = 'components/select/demo/line-height-debug.tsx';
 const interactiveSuffixSelector = '.ant-select-show-arrow';
+const selectSelector = '.ant-select';
 const singleSelector = '.ant-select-single';
-const multipleSelector = '.ant-select-multiple';
 
 const expectSuffixNotHitTarget = async (testPage: Page) => {
   const suffixHit = await testPage.evaluate((selector) => {
@@ -28,25 +28,39 @@ const expectSuffixNotHitTarget = async (testPage: Page) => {
   expect(suffixHit).toBe(false);
 };
 
-const expectHeightKeepsControlHeight = async (testPage: Page) => {
-  const heights = await testPage.evaluate(
-    (selectors) =>
-      selectors.map((selector) => {
-        const select = document.querySelector<HTMLElement>(selector);
+const expectLineHeightKeepsControlHeight = async (testPage: Page) => {
+  const selects = await testPage.evaluate(
+    (selector, singleClsSelector) =>
+      Array.from(document.querySelectorAll<HTMLElement>(selector)).map((select) => {
+        const content = select.querySelector<HTMLElement>(`${selector}-content`);
 
-        if (!select) {
-          throw new Error(`Missing select: ${selector}`);
+        if (!content) {
+          throw new Error(`Missing content: ${selector}`);
         }
 
-        const controlHeight = getComputedStyle(select).getPropertyValue('--ant-select-height');
+        const style = getComputedStyle(select);
+        const readVar = (name: string) => Number.parseFloat(style.getPropertyValue(name));
 
-        return [select.getBoundingClientRect().height, Number.parseFloat(controlHeight)];
+        return {
+          single: select.matches(singleClsSelector),
+          height: select.getBoundingClientRect().height,
+          controlHeight: readVar('--ant-select-height'),
+          contentLineHeight: Number.parseFloat(getComputedStyle(content).lineHeight),
+          tokenLineHeight: readVar('--ant-select-font-size') * readVar('--ant-select-line-height'),
+        };
       }),
-    [singleSelector, multipleSelector],
+    selectSelector,
+    singleSelector,
   );
 
-  heights.forEach(([height, controlHeight]) => {
+  expect(selects).not.toHaveLength(0);
+
+  selects.forEach(({ single, height, controlHeight, contentLineHeight, tokenLineHeight }) => {
     expect(height).toBe(controlHeight);
+
+    if (single) {
+      expect(contentLineHeight).toBeCloseTo(tokenLineHeight, 3);
+    }
   });
 };
 
@@ -79,7 +93,7 @@ describe('Select image', () => {
       React.createElement(LineHeightDebug),
       'select-line-height-debug',
       lineHeightDebugFilename,
-      { beforeScreenshot: expectHeightKeepsControlHeight },
+      { beforeScreenshot: expectLineHeightKeepsControlHeight },
     );
   });
 
--- a/components/select/__tests__/image.test.ts
+++ b/components/select/__tests__/image.test.ts
@@ -2,9 +2,11 @@ import React from 'react';
 import type { Page } from 'puppeteer';
 
 import ClearSuffixDebug from '../demo/clear-suffix-debug';
+import LineHeightDebug from '../demo/line-height-debug';
 import imageTest, { imageDemoTest } from '../../../tests/shared/imageTest';
 
 const clearSuffixDebugFilename = 'components/select/demo/clear-suffix-debug.tsx';
+const lineHeightDebugFilename = 'components/select/demo/line-height-debug.tsx';
 const interactiveSuffixSelector = '.ant-select-show-arrow';
 
 const expectSuffixNotHitTarget = async (testPage: Page) => {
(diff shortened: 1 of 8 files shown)


QDyanbing:
The issue itself is valid, but I’m not sure pinning the single-select content’s `line-height` to `font-height` is the right direction.

Looking through the history, this regression was introduced during the v6 Select semantic-structure refactor: the content uses the current `lineHeight`, while the vertical padding is still calculated from the previously derived `fontHeight`. When `lineHeight` is overridden independently, those two values become inconsistent.

This change keeps the control height stable, but it also makes both the global `lineHeight` and `components.Select.lineHeight` ineffective for the selected text. For example, with either `lineHeight: 1` or `lineHeight: 2`, the computed line height remains the default `22px`.

The `lineHeight: 1` used in multiple mode is not directly comparable here—it normalizes an internal flex wrapper, while the actual selected item has its own explicit height and line height.

Could we keep using the configured `lineHeight` and instead recompute the font height used by the padding from the actual `fontSize * lineHeight`? That should preserve `controlHeight` without ignoring the configured token. It would also be helpful for the test to verify the computed line height for both global and component-level overrides, in addition to checking the root height.

lazerg:
You're right on both counts. I checked the pinned line height: the computed value stayed 22px under `lineHeight: 1`, `lineHeight: 2`, and `components.Select.lineHeight: 2`, so the token really was inert. The multiple mode comparison was wrong too, since the item there carries its own height and line height.

Reworked in 3611416. `--ant-select-font-height` now resolves to `calc(font-size * line-height)`, so the padding follows the configured value while the content keeps `var(--ant-select-line-height)`. I also dropped the `font-height` override in `&-lg`, because it derives now. That covers `lineHeightLG: 2` as well, which stretched large to 48px against a 40px control.

The padding is clamped at 0. The recomputed value turns negative once `fontSize * lineHeight` passes the control height. `size="small"` at `lineHeight: 2` asks for a 28px line box inside a 24px control, so it still grows to 30px, the same as master. Nothing holds both values there without ignoring the token.

The test now checks the computed line height against `font-size * line-height` for global and component level overrides, next to the root height. It fails on master (38px against 32px) and on the previous approach (22px against 28px).

components/select/__tests__/image.test.ts
QDyanbing, line 1: Is this change necessary? imageDemoTest('select') already automatically picks up line-height-debug.tsx and runs the visual regression tests. Skipping the demo and registering it manually only adds assertions against internal CSS variables, duplicating the existing visual coverage and coupling the test to implementation details. This file can remain unchanged.

lazerg: Agreed, reverted in ef1e71f16. The file is back to master.

I checked the numbers before reverting. The manual registration used the identifier `select-line-height-debug`, and that is the same string `imageDemoTest` builds at `tests/shared/imageTest.tsx:374` from `${component}-${basename}`. Both paths write the same three PNGs. After the revert the suite still runs 113 tests, and the three files hash identically to the ones the manual test produced, so no baseline moves.

I also confirmed the screenshot alone catches this bug. With the fix in `select-input.ts` reverted, the demo snapshot grows in every theme: default 80 -> 86 px, dark 80 -> 86 px, compact 76 -> 80 px. The visual diff sees that, so the CSS variable assertions added nothing.

select / auto-complete / cascader / tree-select: 26 suites, 436 passed. tsc, eslint and biome are clean.

components/select/demo/line-height-debug.tsx
```diff
@@ -0,0 +1,20 @@
...
+  <Flex gap="small" align="start">
+    <ConfigProvider theme={{ token: { lineHeight: 2 } }}>
+      <Select style={selectStyle} defaultValue="light" options={options} />
+      <Select style={selectStyle} mode="multiple" defaultValue={['light']} options={options} />
+    </ConfigProvider>
+    <ConfigProvider theme={{ components: { Select: { lineHeight: 2 } } }}>
+      <Select style={selectStyle} defaultValue="light" options={options} />
+    </ConfigProvider>
+  </Flex>
+);
```
QDyanbing, line 36: Could we also add small and large cases to the demo? The large case should use lineHeightLG: 2.

QDyanbing:
The current implementation still does not seem to guarantee a stable control height. With size="small" and lineHeight: 2, the single Select is 30px while the multiple Select remains 24px, and larger lineHeight values can still expand the default size. max(..., 0px) only clamps the padding, not the line box itself. However, I’m not fully certain about the intended behavior here, so please don’t adjust it based solely on my suggestion—it would be better to get input from someone more familiar with this implementation. 

@zombieJ @li-jia-nan cc

components/select/demo/line-height-debug.tsx
lazerg: Added both in f9eb07a7e. `small` now sits beside the default one in the `lineHeight: 2` group, and large gets its own `lineHeightLG: 2` group, single and multiple in each.

lazerg:
The numbers hold up. I measured this branch in Chrome, default theme:

- `size="small"` with `lineHeight: 2`: single 30px, multiple 24px
- `size="middle"` with `lineHeight: 3`: single 44px, multiple 32px

On master the same two single cases measure 30px and 52px. So the small case is not a regression from this PR, but this PR does not fix it either.

The line box does not fit. A small control is 24px tall with 2px of border, which leaves 22px of content box, while `14px * 2` asks for a 28px line box. The padding is already at 0, so there is nothing left to give back. Multiple mode keeps 24px only because `-selection-item` sets its own line height from `multi-item-height` at select-input-multiple.ts:94 and never reads the `lineHeight` token. That is the real split between the two modes: single honors the token, multiple ignores it.

There is a shape that keeps both stable and keeps the token live. Clamp the font height var to the space that exists, then drive the root line height from it:

```ts
[varName('font-height')]:
  `min(calc(${varRef('font-size')} * ${varRef('line-height')}), calc(${varRef('height')} - ${varRef('border-size')} * 2))`,
// ...
lineHeight: varRef('font-height'),
```

I measured 36 combinations (single and multiple, small/middle/large, `lineHeight` 1 / default / 2 / 3, plus `lineHeightSM: 2` and `lineHeightLG: 2`) across the default, dark and compact themes. Root height matched `--ant-select-height` in 108 of 108 cases. Default theme output does not move: content line height stays 22px for small and middle, 24px for large. The token still applies wherever it fits, so `lineHeight: 2` at middle gives a 28px line box and `lineHeightLG: 2` at large gives 32px. It only clamps when the requested line box is taller than the control itself.

I have not pushed it, since you asked for @zombieJ and @li-jia-nan to weigh in first. I can add it as a commit if you three agree it is the behavior you want.

nrps9909 (approved):
Reviewed exact head `f9eb07a7ea60d9d4b6854310f2c9c8e986713260`.

The dynamic `font-height` calculation keeps the content line box and vertical-padding formula on the same configured token value, while `max(..., 0px)` safely handles values that cannot physically fit inside the control. Independent real-Chrome measurements confirmed normal Selects stay at 32px for global `lineHeight: 2` and `lineHeight: 1`, and large Selects stay at 40px for `lineHeightLG: 2`; single and multiple modes agree in each supported fitting case. The disclosed small/28px-line-box boundary clamps padding to zero rather than producing invalid negative spacing.

Independent validation on this exact head:

- complete Select Jest scope: 6 suites, 156 tests passed, 5 skipped, 104 snapshots
- focused ESLint, Biome, and diff validation passed
- all substantive remote build, lint, coverage, module, React 18/19, Vitest, and visual snapshot jobs are green
- the remaining visual-diff gate is the expected manual approval for three newly added debug-demo images, not a detected changed-image regression

Approved.

zombieJ:
Thanks for the contribution. Superseded by #59298, which fixes the underlying theme token derivation so fontHeight follows the final fontSize and lineHeight values.

PR #59106, opened 2026-08-26
fix(Result): hide decorative exception illustrations

nrps9909:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

No linked issue. I checked current open issues and pull requests for this Result exception-illustration accessibility defect and found no duplicate.

### 💡 Background and Solution

The built-in 403, 404, and 500 Result SVGs expose hardcoded English `<title>` elements. Result already renders a visible title and subtitle, so screen readers can receive duplicated content and non-English interfaces can leak English.

This change:

- marks all three built-in exception illustrations as decorative with `aria-hidden="true"`
- keeps them out of legacy keyboard focus with `focusable="false"`
- adds a regression covering 403, 404, and 500

Custom `icon` content is unchanged.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Result hides the built-in 403, 404, and 500 decorative illustrations from assistive technology to avoid duplicated English announcements. |
| 🇨🇳 Chinese | Result 内置的 403、404 和 500 装饰插图不再被辅助技术读取，避免重复播报英文内容。 |

### Verification

- Exact-base regression: failed because the built-in SVG had no `aria-hidden`
- Jest: 17/17 Result tests passed
- Vitest: 17/17 Result tests passed
- ESLint, Biome, Prettier, and `git diff --check`: passed
- Full `tsc --noEmit`: after excluding an unrelated nested review worktree, the only remaining error is the existing `@types/puppeteer@5` versus `puppeteer-core@25` Page mismatch in `tests/shared/imageTest.tsx`

AI assistance disclosure: Codex was used to trace the accessibility behavior, check for duplicate work, implement the focused change, and run verification.

### DIFF
--- a/components/result/noFound.tsx
+++ b/components/result/noFound.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
-const NoFound: React.FC = () => (
-  <svg width="252" height="294" aria-hidden="true" focusable="false">
+const NoFound: React.FC<React.SVGProps<SVGSVGElement>> = (props) => (
+  <svg width="252" height="294" {...props}>
     <title>No Found</title>
     <g fill="none" fillRule="evenodd">
       <circle cx="126.75" cy="128.1" r="126" fill="#E4EBF7" />
--- a/components/result/noFound.tsx
+++ b/components/result/noFound.tsx
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 const NoFound: React.FC = () => (
-  <svg width="252" height="294">
+  <svg width="252" height="294" aria-hidden="true" focusable="false">
     <title>No Found</title>
     <g fill="none" fillRule="evenodd">
       <circle cx="126.75" cy="128.1" r="126" fill="#E4EBF7" />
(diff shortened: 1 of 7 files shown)


nrps9909:
The first full CI run exposed six expected Result demo snapshot mismatches for the new SVG attributes. Commit 4923b8e updates only the 403/404/500 normal and extended demo snapshots. The two affected demo suites now pass locally: 22 tests and 30 snapshots. A fresh CI run is in progress.

components/result/noFound.tsx
```diff
@@ -1,7 +1,7 @@
 import * as React from 'react';
 
 const NoFound: React.FC = () => (
-  <svg width="252" height="294">
+  <svg width="252" height="294" aria-hidden="true" focusable="false">
```
yoyo837: These SVG components are also exposed as Result.PRESENTED_IMAGE_403/404/500. Hard-coding aria-hidden here makes standalone usage inaccessible and cannot be overridden. Could we apply it only when Result renders the built-in illustration, or allow SVG props?

nrps9909: Fixed in signed commit 7ddd647. Result now applies aria-hidden and focusable=false only when it renders a built-in exception illustration. The public PRESENTED_IMAGE_403/404/500 components accept standard SVG props and remain semantic in standalone use; a new regression verifies an explicit role and accessible name are preserved. The Result scope passes 54 Jest tests with 32 snapshots and 21 Vitest tests. Focused ESLint, Biome, formatting, and diff checks pass; TypeScript reaches only the same pre-existing Puppeteer Page mismatch reproduced on the exact base.

PR #59110, opened 2026-08-27
feat: support aria-describedby on handles

yoyo837:

### 🤔 This is a ...

- [x] 🆕 New feature see https://github.com/react-component/slider/pull/1088
- [ ] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

### 💡 Background and Solution

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English |    feat: support aria-describedby on handles       |
| 🇨🇳 Chinese |   feat: 支持为滑块手柄设置 aria-describedby      |

### DIFF
--- a/components/slider/index.tsx
+++ b/components/slider/index.tsx
@@ -98,6 +98,7 @@ export interface SliderBaseProps {
   tabIndex?: SliderProps['tabIndex'];
   ariaLabelForHandle?: SliderProps['ariaLabelForHandle'];
   ariaLabelledByForHandle?: SliderProps['ariaLabelledByForHandle'];
+  ariaDescribedByForHandle?: SliderProps['ariaDescribedByForHandle'];
   ariaRequired?: SliderProps['ariaRequired'];
   ariaValueTextFormatterForHandle?: SliderProps['ariaValueTextFormatterForHandle'];
 }
--- a/package.json
+++ b/package.json
@@ -149,7 +149,7 @@
     "@rc-component/switch": "~1.0.3",
     "@rc-component/table": "~1.11.1",
     "@rc-component/tabs": "~1.13.0",
-    "@rc-component/tooltip": "~1.5.0",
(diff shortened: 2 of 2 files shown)


yoyo837:
~Waiting  https://github.com/ant-design/ant-design/pull/59111~

nrps9909 (changes requested):
The new prop is exposed in the public type but is not preserved by the default Ant Design Slider rendering path. I verified this on exact head `c53a696` with `@rc-component/slider@1.2.0`:

```tsx
const { container } = render(
  <Slider defaultValue={20} ariaDescribedByForHandle="description" />,
);
expect(container.querySelector('[role="slider"]')).toHaveAttribute(
  "aria-describedby",
  "description",
);
```

The assertion fails because the handle has no `aria-describedby`. The same probe rendered directly with `RcSlider` passes. Ant Design always supplies its internal `handleRender`; the resulting handle is wrapped in `SliderTooltip`, and rc-tooltip clones the child with `aria-describedby: undefined` while the tooltip is closed, overriding the consumer-provided description. When the tooltip opens, its generated id would also replace rather than compose with the supplied description.

Please preserve the caller-provided description in the default tooltip path (and compose it with the tooltip id while open), with single-handle and per-handle range regressions. The existing Slider scope passes 77 tests / 49 snapshots, but none exercises this new contract.

yoyo837:
~Wating https://github.com/react-component/tooltip/pull/542 release patch.~

QDyanbing (approved):

PR #59112, opened 2026-08-27
fix(Typography): avoid nested edit button semantics

nrps9909:
### 🤔 This is a ...

- [x] 🐞 Bug fix
- [x] ✅ Test Case

### 🔗 Related Issues

No existing issue or open PR was found for this exact Typography behavior.

### 💡 Background and Solution

Editable Typography already renders its edit action as a native `<button>` with a localized `aria-label`, but the default `EditOutlined` child was also forced to `role="button"`. Accessibility queries therefore exposed one visual edit action as two nested buttons.

This removes the redundant role from the default icon so it returns to its normal image semantics while the parent remains the only interactive control. A regression test renders the default editable Typography and requires exactly one accessible button.

Before the fix, the new assertion received two button roles. After the fix:

- complete Typography Jest scope: 10 suites, 163 tests
- 47 snapshots
- focused ESLint and Biome lint
- snapshot formatting and `git diff --check`

I also audited every currently open Ant Design PR changed-file list and found no overlap with the two source/test files.

AI assistance disclosure: Codex was used to trace the rendered roles, write and run the failing regression, audit open-PR file overlap, and draft this description. I verified the failure, fix, exact head, and validation results locally.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Typography's default edit icon exposing a duplicate nested button role. |
| 🇨🇳 Chinese | 修复 Typography 默认编辑图标暴露重复嵌套按钮语义的问题。 |

### DIFF
--- a/components/typography/Base/index.tsx
+++ b/components/typography/Base/index.tsx
@@ -462,7 +462,7 @@ const Base = React.forwardRef<HTMLElement, BlockProps>((props, ref) => {
           aria-label={ariaLabel}
           tabIndex={tabIndex}
         >
-          {icon || <EditOutlined role="button" />}
+          {icon || <EditOutlined />}
         </button>
       </Tooltip>
     ) : null;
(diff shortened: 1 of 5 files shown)


QDyanbing (approved):
LGTM

Issue #58404, opened 2026-06-17
[upgrade] migrate class components to FC


li-jia-nan:
### 这个功能解决了什么问题？

- Ant Design 及部分运行时依赖的构建产物中仍包含一些 React class component。

- 基于当前本地构建产物和依赖产物扫描，示例包括：

  - [x] `Alert.ErrorBoundary`（忽略）
  - [ ] `@rc-component/tree`: `Tree` https://github.com/react-component/tree/pull/1021 @li-jia-nan 
  - [ ] `@rc-component/form`: `Field` https://github.com/react-component/field-form/pull/798 @li-jia-nan 
  - [ ] `@rc-component/menu`: `LegacyMenuItem` https://github.com/react-component/menu/pull/871 @ZQDesigned 
  - [ ] `@rc-component/motion`: `CSSMotionList` https://github.com/react-component/motion/pull/82 @QDyanbing 
  - [ ] `@rc-component/util`: `PortalWrapper` https://github.com/react-component/util/pull/780 @QDyanbing 
  - [ ] `@ant-design/react-slick`: `Slider`、`InnerSlider`、`PrevArrow`、`NextArrow`、`Dots`、`Track`
    - https://github.com/ant-design/react-slick/pull/124 @QDyanbing

- 对于依赖 class 特性的组件，比如 `Alert.ErrorBoundary` ，我们暂时保持现状，除非将来有其他替代方案
- 对于不依赖 class 特性的组件，如果能逐步迁移为函数组件，可以让 Ant Design 及相关依赖的实现更加统一，也更符合当前 React 生态的主流写法

### 期望的 API 是什么样的？

- 不期望引入任何公开 API 变化

QDyanbing:
PortalWrapper 处理中

QDyanbing:
我简单分析了一下 `PortalWrapper` 的使用情况，它看起来像是早期 portal/body scroll lock 逻辑遗留下来的历史产物。

从当前本地源码扫描来看，没有发现 Ant Design 或 rc-component 相关包在运行时使用它。现存引用主要是：

- `rc-util` / `@rc-component/util` 自己的文档和测试
- dialog/image 相关类型定义里对 `GetContainer` 类型的引用

另外 `PortalWrapper` 本身也没有从 root entry 导出，当前主要是 `GetContainer` 类型还放在这个文件里。

[...]

PR #59199, opened 2026-09-03
script: add class-component regression guard to lint

LeonxLJX:
Closing this in favour of QDyanbing's earlier review comment: the
`scripts/check-class-component.ts` lint script duplicates a check
they'd rather keep under code review (per the AGENTS.md direction),
and the regression concern from #58404 can be addressed with a
focused unit test in `components/*/__tests__/` instead — narrower
scope, no new global lint step to maintain.

If/when there's appetite to revisit a CI-level guard, I'm happy
to reopen against an AGENTS.md-aligned design. Sorry for the
context churn in the meantime.


QDyanbing:
我个人感觉这项检查的意义比较有限，也不太希望为此新增脚本或 CI 任务。类组件回退更像是代码规范问题，而不是需要在每次 lint 中强制检查的正确性问题。新增一个全量扫描脚本还会带来额外的执行和维护成本。

我倾向于继续通过 code review 控制；如果主要是约束 AI 生成的代码，在 AGENTS.md 中明确禁止新增类组件就足够了。在实际出现频繁回退之前，我认为没有必要增加永久性的自动检查。

---

My personal feeling is that this check provides relatively limited value, and I would prefer not to add another script or CI task for it. Reintroducing class components seems more like a coding-convention concern than a correctness issue that needs to be enforced during every lint run. A full-source scanning script also introduces additional execution and maintenance costs.

I would prefer to keep this under code review. If the main concern is AI-generated code, adding a clear restriction against new class components to AGENTS.md should be sufficient. Unless this becomes a recurring regression in practice, I don’t think we need a permanent automated check.

PR #59221, opened 2026-09-05
fix(Tabs): apply styles.popup.root to the more dropdown

giaBaoJS:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

No linked issue. The problem is visible in the repository itself: `components/tabs/demo/_semantic.tsx` sets `styles={{ popup: { root: { background: '#fff' } } }}`, and that value never reaches the DOM.

### 💡 Background and Solution

`Tabs` declares the popup semantic node as a nested object on both sides (`components/tabs/index.tsx`):

```ts
classNames?: { ...; popup?: { root?: string } };
styles?:     { ...; popup?: { root?: React.CSSProperties } };
```

`@rc-component/tabs` keeps `popup` flat (`Partial<Record<SemanticName, string>>` / `Partial<Record<SemanticName, CSSProperties>>`), so the nested value has to be unwrapped on the way down. `classNames` is unwrapped:

```tsx
classNames={{
  ...mergedClassNames,
  popup: clsx(popupClassName, hashId, cssVarCls, rootCls, mergedClassNames.popup?.root),
}}
styles={mergedStyles}
```

`styles` is not. `mergedStyles.popup` therefore arrives at rc-tabs as `{ root: CSSProperties }`, is forwarded as `popupStyle` -> `overlayStyle` on the more dropdown, and React drops it because `root` is not a CSS property. The result is that `classNames.popup.root` works and `styles.popup.root` silently does nothing on the same element.

`styles` is now unwrapped the same way `classNames` already is. This is how every sibling component with the same nested popup semantic already behaves, for example `components/select/index.tsx` (`popupStyle={{ ...mergedStyles.popup.root, ... }}`), `components/cascader/index.tsx` and `components/auto-complete/AutoComplete.tsx`.

Nothing else changes: the layout, the public API, and the `classNames` path are untouched, and no snapshot moves because the more dropdown is not opened in any snapshot test.

`components/tabs/__tests__/semantic.test.tsx` gains a case that overflows the tab bar, opens the more dropdown and asserts both halves on the popup element. Reverting only `components/tabs/index.tsx` turns it red on the style assertion while the class assertion above it still passes:

```
● Tabs.Semantic › support popup classNames and styles

  expect(element).toHaveStyle()

  - Expected

  - color: rgb(255, 0, 0);
  + color: var(--ant-color-text);
```

Verification:

- `npx jest --config .jest.js --no-cache components/tabs components/config-provider` -> 27 suites passed, 662 tests passed, 455 snapshots passed, 0 snapshots written.
- `npx vitest run components/tabs/__tests__/semantic.test.tsx` -> 4 passed.
- `npm run tsc`, `eslint`, `biome lint`, `biome format` -> clean.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | 🐞 Fix Tabs ignoring `styles.popup.root` on the more dropdown. |
| 🇨🇳 Chinese | 🐞 修复 Tabs 折叠下拉菜单未应用 `styles.popup.root` 的问题。 |

### DIFF
--- a/components/tabs/__tests__/semantic.test.tsx
+++ b/components/tabs/__tests__/semantic.test.tsx
@@ -2,12 +2,12 @@ import React from 'react';
 
 import Tabs from '..';
 import type { TabsProps } from '..';
-import { fireEvent, render, triggerResize, waitFakeTimer } from '../../../tests/utils';
-import ConfigProvider from '../../config-provider';
 import {
   expectSemanticRootStylePriority,
   semanticRootStylePriority,
 } from '../../../tests/shared/semanticStylePriority';
+import { render } from '../../../tests/utils';
+import ConfigProvider from '../../config-provider';
 
 describe('Tabs.Semantic', () => {
   it('support classnames and styles', () => {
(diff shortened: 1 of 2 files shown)


nrps9909 (approved):
Reviewed exact head `f91167d0c96b96af3926a20fbd3ba1ad12b61d03` against base `58b0ee1554835b3ea2d0541f4986306f3d70d0ce`. The projection of `mergedStyles.popup.root` matches rc-tabs' flat popup style contract; I found no blocker in this scoped change.

Validation:

- The new overflow-popup regression fails against the base implementation and passes on the PR head. The four semantic tests also pass under Vitest.
- Two independent DOM probes pass: provider-only popup fallback, local popup precedence, merged popup class names, preservation of root/header styles, callback evaluation, and clearing obsolete inline styles after rerenders.
- The complete Tabs + ConfigProvider run reports 663 passed, 4 skipped, and one failed test; 454 snapshots pass and one differs. The remaining `tabs/demo/nest.tsx` extend-context snapshot gains Select `aria-posinset`/`aria-setsize` attributes. The exact base produces the identical 16-line diff with the same installed dependencies. I did not update that snapshot or count this as a clean full-suite pass.
- ESLint, Biome and diff checks pass for the changed files. All 42 applicable remote checks pass; two are skipped.

Separate existing behavior observed during probing: when both ConfigProvider and Tabs supply `styles.popup.root`, the shared `mergeStyles` helper replaces the nested root object, so unrelated provider properties are lost. That helper is unchanged by this PR. Provider-only fallback and the local styles work with this fix; property-level merging of nested style slots would be a separate follow-up.

Review and independent validation performed with OpenAI Codex assistance.

components/tabs/__tests__/semantic.test.tsx
```diff
@@ -102,6 +102,67 @@ describe('Tabs.Semantic', () => {
     expect(root).toHaveClass('custom-card-root');
     expect(root).toHaveStyle({ backgroundColor: 'rgb(255, 0, 0)' });
   });
+  it('support popup classNames and styles', async () => {
+    // The more dropdown only mounts when the tabs overflow, so fake the layout
+    // Longest first: `ant-tabs-nav` is a prefix of the other two
+    const sizes: Record<string, number> = {
```
QDyanbing: Could we open the dropdown directly by configuring `more`, then verify the popup's class and style? This test mainly checks whether the styles are applied, so it should not need to simulate element dimensions, tab overflow, or resize and hover events. Could we simplify the test setup?

giaBaoJS: Updated. The test now passes `more={{ visible: true }}` and opens the popup directly.

In rc-tabs `OperationNode`, `more` is spread onto the Dropdown after its own `visible: tabs.length ? open : false`, so `more.visible` wins and the popup mounts without any overflowing tabs. That removed the `offsetWidth`/`offsetHeight`/`offsetLeft` prototype patching, the resize trigger, the mouseEnter event and the fake timers. The test is synchronous now and asserts the same class and style on `.ant-tabs-dropdown`.

Coverage is unchanged: with `components/tabs/index.tsx` reverted to the base version, only the style assertion fails (`color: var(--ant-color-text)` instead of `rgb(255, 0, 0)`) while the class assertion still passes. Green under both jest and vitest.

components/tabs/__tests__/semantic.test.tsx
QDyanbing, line 33: Can the new case be added directly to this case?

giaBaoJS: Done, folded into `support classnames and styles`. `popup: { root: ... }` now sits in the existing `customClassnames` and `customStyles` objects, the render gets `more={{ visible: true }}` so the dropdown mounts, and the two assertions moved in beside the others. The separate case is gone.

One thing worth a look: the popup is portaled, so it is read off `document.body` rather than `container` like the rest.

Counterfactual still holds. With `components/tabs/index.tsx` reverted, the case fails on `color: var(--ant-color-text)` instead of `rgb(0, 255, 255)`. `components/tabs` goes from 88 to 87 tests, no snapshot change.

QDyanbing:
LGTM。这里参照 `classNames.popup.root` 的处理方式，将 `styles.popup.root` 转为 rc-tabs 接收的扁平结构。再看看其他 reviewer 是否还有需要确认的地方。

---

LGTM. This handles `styles.popup.root` the same way as `classNames.popup.root`, adapting it to the flat structure expected by rc-tabs. Let’s see if the other reviewers have anything else to check.

zombieJ (approved):
LGTM

PR #59224, opened 2026-09-05
fix(locale): de_DE DatePicker fieldDateFormat DD.MM.YYYY

zkkk9555:
### 🤔 This is a ...
- [ ] 🆕 New feature
- [ ] 🐞 Bug fix
- [x] 🌐 Internationalization

Closes #59149.

### 改动点
- components/date-picker/locale/de_DE.ts :: lang 加 fieldDateFormat='DD.MM.YYYY' + fieldDateTimeFormat='DD.MM.YYYY HH:mm:ss'（sq_AL同形；rc-picker de_DE无此四键，唯一德语值）。
- components/locale/__tests__/__snapshots__/index.test.tsx.snap 更新（de快照title US→德语，改动生效证据）。

### 验证命令与结果（短ASCII路径 C:/a/ad 实跑）
- npx tsc --noEmit → exit 0（全仓）
- npx eslint components/date-picker/locale/de_DE.ts → exit 0
- npx jest --config .jest.js components/locale/__tests__/index.test.tsx --no-cache → 77/77（-u 1 updated）；复跑77/77无flake

PR #59231, opened 2026-09-06
fix(Select): keep single selector height stable under custom global lineHeight

### LINKED ISSUE #59091: 全局token 设置 lineHeight=2 时， Select 单选与多选  会出现高度不一样

zkkk9555:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

- Fix #59091

### 💡 Background and Solution

With a custom global `lineHeight` token, the single Select box drifts away from `controlHeight` while the multiple Select stays put (e.g. `lineHeight=1` → single 24px vs multiple 32px; `lineHeight=2` → single 38px vs multiple 32px; default token is unaffected at 32/32).

Root cause: the single `.ant-select-content` line box inherits the global `lineHeight` token through the root container, but the vertical padding is computed from the fixed `font-height` (`(height - font-height) / 2 - border`), so the leftover stretches (or shrinks) the border-box past `controlHeight`. Multiple locks its content to `lineHeight: 1` and never drifts.

This PR pins the single content line box to `var(--ant-select-font-height)` so the total stays at `controlHeight` under any global `lineHeight`. Text stays vertically centered via the existing `alignSelf: center`.

Verified with a headless Chromium (puppeteer + esbuild bundle from source) measuring `.ant-select-single` / `.ant-select-multiple` offset heights under `ConfigProvider theme.token.lineHeight` in {default, 1, 2}:
- before: default 32/32, LH=1 24/32, LH=2 38/32
- after: 32/32 at all three levels; placeholder-only and showSearch single states also 32/32

### 📝 Change Log

| Language   | Changelog                                              |
| ---------- | ------------------------------------------------------ |
| 🇺🇸 English | Fix Select single selector height drifting with the global `lineHeight` token. |
| 🇨🇳 Chinese | 修复 Select 单选框高度随全局 `lineHeight` token 漂移的问题。 |

### DIFF
--- a/components/select/__tests__/__snapshots__/index.test.tsx.snap
+++ b/components/select/__tests__/__snapshots__/index.test.tsx.snap
@@ -146,3 +146,99 @@ exports[`Select rtl render component should be rendered correctly in RTL directi
   </div>
 </div>
 `;
+
+exports[`Select single selector height vs global lineHeight locks single content line box to font-height when global lineHeight is 1: single-selector-lineHeight-1 1`] = `
+<div
+  class="ant-select ant-select-outlined css-var-test-id ant-select-css-var ant-select-single ant-select-show-arrow"
+>
+  <div
+    class="ant-select-content ant-select-content-has-value"
+    title="One"
+  >
+    One
+    <input
+      aria-autocomplete="list"
+      aria-expanded="false"
+      aria-haspopup="listbox"
+      autocomplete="new-password"
+      class="ant-select-input"
+      id="test-id"
+      readonly=""
+      role="combobox"
+      type="text"
+      value=""
+    />
+  </div>
+  <div
+    class="ant-select-suffix"
+  >
+    <span
+      aria-label="down"
+      class="anticon anticon-down"
+      role="img"
+    >
+      <svg
+        aria-hidden="true"
+        data-icon="down"
+        fill="currentColor"
+        focusable="false"
+        height="1em"
+        viewBox="64 64 896 896"
+        width="1em"
+      >
+        <path
+          d="M884 256h-75c-5.1 0-9.9 2.5-12.9 6.6L512 654.2 227.9 262.6c-3-4.1-7.8-6.6-12.9-6.6h-75c-6.5 0-10.3 7.4-6.5 12.7l352.6 486.1c12.8 17.6 39 17.6 51.7 0l352.6-486.1c3.9-5.3.1-12.7-6.4-12.7z"
+        />
+      </svg>
+    </span>
+  </div>
+</div>
+`;
+
+exports[`Select single selector height vs global lineHeight locks single content line box to font-height when global lineHeight is 2: single-selector-lineHeight-2 1`] = `
+<div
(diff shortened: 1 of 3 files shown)


zkkk9555:
Follow-up pushed (commit `ca3b4d6`, rebased on latest `master`):

- Removed the unrelated `de_DE` locale commits from this PR — the diff now contains only the Select fix + its regression test (+30/-0, single commit). The German `fieldDateFormat` change stays in #59224.
- Reworked the regression test per the review: split into `it.each([1, 2])` so each lineHeight value gets its own case with an isolated render (`unmount()` after each), instead of one test rendering both and matching across shared styles.

Local verification on the rebased branch: `jest components/select` 55/55 (3 snapshots), `tsc --noEmit` clean, `eslint` on touched files clean; the new cases fail without the style change and pass with it.

zkkk9555:
Follow-up pushed (commit `f5cc2b7`) addressing the remaining inline comment (per-`lineHeight` verification + snapshot):

- Added a per-value DOM snapshot (`single-selector-lineHeight-1/2`) inside the already-split `it.each([1, 2])` cases, so structural regressions in this scenario are caught alongside the style-text assertion. Snapshots use the test-env fixed ids (`css-var-test-id`/`test-id`), no dynamic hashes.
- The core style assertion stays per-case isolated (own render + `unmount()`), so one value's styles can't mask the other's regression — the original concern is already covered; the snapshot is an extra lock.

Local verification: `jest components/select` 55/55 (5 snapshots, re-run stable), `eslint` on touched files clean. The fix itself is unchanged (single commit before this one).

QDyanbing:
感谢贡献！这个问题已有更早提交的 PR #59092，请避免重复提交。提交 PR 前，请先搜索相关 issue 和 PR；如果已有其他贡献者在处理，请先在原讨论中沟通，确认对方不再继续跟进后再接手，以免重复工作。

---

Thanks for contributing! An earlier PR (#59092) already addresses this issue, so please avoid submitting a duplicate. Before opening a PR, please search for related issues and PRs. If another contributor is already working on it, please coordinate in the existing discussion and confirm that they are no longer pursuing the fix before taking over, to avoid duplicating effort.

zkkk9555:
Closing as duplicate: this issue is already addressed by earlier PR #59092. Thanks for pointing it out — I should have searched for existing PRs before opening this. Sorry for the noise.

PR #59084, opened 2026-08-24
fix(Splitter): respect size limits when container resizes

QDyanbing:
### 🤔 这个变动的性质是？

- [ ] 🆕 新特性提交
- [x] 🐞 Bug 修复
- [ ] 📝 站点、文档改进
- [ ] 📽️ 演示代码改进
- [ ] 💄 组件样式/交互改进
- [ ] 🤖 TypeScript 定义更新
- [ ] 📦 包体积优化
- [ ] ⚡️ 性能优化
- [ ] ⭐️ 功能增强
- [ ] 🌐 国际化改进
- [ ] 🛠 重构
- [ ] 🎨 代码风格优化
- [ ] ✅ 测试用例
- [ ] 🔀 分支合并
- [ ] ⏩ 工作流程
- [ ] ⌨️ 无障碍改进
- [ ] ❓ 其他改动（是关于什么的改动？）

### 🔗 相关 Issue

- close #59083

### 💡 需求背景和解决方案

Splitter 面板拖动后会保存像素尺寸。当窗口或父容器尺寸变化时，这些尺寸会按比例重新计算，但未重新应用 Panel 的 `min` 和 `max` 约束，可能导致面板尺寸越界。

本次改动在按容器尺寸缩放后重新应用尺寸约束，并将产生的尺寸差额分配给仍有调整空间的其他面板，确保 Panel 遵守限制的同时继续填满容器。折叠面板保持为 0，不涉及 API 变更。

已补充数值及百分比限制、折叠状态、不可满足约束和拖动后缩放容器的测试。

### 📝 更新日志

| 语言    | 更新描述 |
| ------- | -------- |
| 🇺🇸 英文 | 🐞 Fix Splitter Panel `min` and `max` constraints not being respected when its container resizes. |
| 🇨🇳 中文 | 🐞 修复 Splitter 容器尺寸变化时 Panel 的 `min` 和 `max` 约束失效的问题。 |

PR #59232, opened 2026-09-06
fix(Splitter): respect min widths when container resizes

### LINKED ISSUE #59083: Splitter min widths are not respected when component is resized
### Reproduction link

[Edit on CodeSandbox](https://codesandbox.io/p/sandbox/fervent-snow-wr29hm)

### Steps to reproduce

Resize the first panel to the minimum width and then reduce the page width

### What is expected?

Min widths of content panels should be respected if the component resizes, for example if the window itself is resized

Component is reduced to lower than min width

| Environment | Info |
| --- | --- |
| antd | 6.6.1 |
| React | 18.3.1 |
| System | Mac Os 15.6.1 |
| Browser | Chromium 146.0.7680.178 |

zkkk9555:
### 🤔 This is a ...

- [x] 🐞 Bug fix

### 🔗 Related Issues

- Fix #59083

### 💡 Background and Solution

Dragging a panel stores absolute px sizes, but when the container later shrinks (e.g. window resize), `useSizes` re-bases those stale px values on the new container size without re-clamping against `min`/`max`: sizes `[200, 800]` captured at 1000px become `[100, 400]` at 500px even with `min: 200`, and `autoPtgSizes` only enforces limits for `auto` (undefined) panels.

This PR re-clamps explicit sizes in the container-resize path:

- new `limitPtgSizes` in `sizeUtil.ts` pins sub-`min` explicit sizes back to `min` (collapsed `0` and `auto` entries untouched; `max` side deliberately unenforced so pre-existing collapse-restore snapshots are preserved);
- the all-defined branch of `autoPtgSizes` now shrinks only the above-`min` part proportionally instead of scaling everything (including `min` itself) down, with the previous proportional behavior kept for the impossible case;
- `postPxSizes` (`useSizes`) and `getPxSizes` (`useResize`) absorb float dust into the last panel so totals and collapse snapshots stay exact.

Verified: new `size.test.tsx` case (explicit `[200, 800]` + `min: 200` at a 500px container) fails before (`100`) and passes after (`>= 200`); full `components/splitter` suite 111/111 with 44 snapshots; `tsc --noEmit` and `eslint` on touched files pass.

### 📝 Change Log

| Language   | Changelog                                              |
| ---------- | ------------------------------------------------------ |
| 🇺🇸 English | Fix Splitter ignoring panel `min` widths when the container resizes. |
| 🇨🇳 Chinese | 修复 Splitter 在容器尺寸变化时忽略面板 `min` 宽度的问题。 |

### DIFF
--- a/components/splitter/__tests__/size.test.tsx
+++ b/components/splitter/__tests__/size.test.tsx
@@ -164,4 +164,26 @@ describe('useSizes', () => {
 
     expect(postPxSizes).toEqual([200, 200]);
   });
+
+  // Review follow-up of https://github.com/ant-design/ant-design/pull/59232:
+  // when mins exactly fill the container, every panel must pin at its `min`
+  // instead of being scaled below it.
+  it('should pin every panel at min when mins exactly fill the container', () => {
+    const items = [
+      {
+        size: 900,
+        min: 500,
+      },
+      {
+        size: 300,
+        min: 500,
+      },
+    ];
+
+    const { result } = renderHook(() => useSizes(items, 1000));
+    const [, postPxSizes] = result.current;
+
+    expect(postPxSizes[0]).toBeGreaterThanOrEqual(500);
+    expect(postPxSizes[1]).toBeGreaterThanOrEqual(500);
+  });
 });
--- a/components/splitter/__tests__/size.test.tsx
+++ b/components/splitter/__tests__/size.test.tsx
@@ -186,4 +186,31 @@ describe('useSizes', () => {
     expect(postPxSizes[0]).toBeGreaterThanOrEqual(500);
     expect(postPxSizes[1]).toBeGreaterThanOrEqual(500);
   });
+
+  // Review follow-up of https://github.com/ant-design/ant-design/pull/59232:
+  // when no panel can absorb float dust within `[min, max]`, compensation
+  // must be skipped instead of breaking a panel bound (the old fallback
+  // wrote the dust into the last panel: 149.9999999999999 < min 150).
+  it('should skip dust compensation when no panel can absorb it within bounds', () => {
+    const items = [
+      {
+        size: 290,
+        min: 150,
+        max: 290,
+      },
+      {
+        size: 150,
+        min: 150,
+        max: 150,
+      },
+    ];
+
+    const { result } = renderHook(() => useSizes(items, 1000));
+    const [, postPxSizes] = result.current;
+
+    // Panel 1 is pinned at min=max=150 and cannot absorb the +1.1e-13 dust
+    // without breaking its bound; panel 0 is already at its max=290, so the
+    // dust stays unabsorbed and panel 1 keeps exactly 150.
+    expect(postPxSizes[1]).toBe(150);
+  });
 });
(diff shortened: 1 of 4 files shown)


zkkk9555:
Follow-up pushed (commit `3b6cd95`, rebased on latest `master`):

- Removed the unrelated `de_DE` locale commits — the diff now contains only the 4 Splitter files in a single commit. The German `fieldDateFormat` change stays in #59224.
- Fixed the float-dust compensation per the review: it now only absorbs tiny round-off (gated by an `EPSILON * container * panels` tolerance), and prefers a panel that stays within its `max` after absorbing, so a real `max` leftover is never written into a panel.
- Added the requested regression test: two panels with `max: 200` in a 1000px container stay `[200, 200]` (verified red on the old compensation: it produced `[200, 800]`).

Local verification on the rebased branch: `components/splitter` 9 suites 112/112 (44 snapshots), `tsc --noEmit` clean, `eslint` on touched files clean.

zkkk9555:
Follow-up pushed (commit `f28478e`):

- Verified the finding first: with sizes `[900, 300]` and mins `[500, 500]` (mins exactly fill the container), the old `floorTotal >= 1` guard took proportional scaling and produced `[643, 257]` — the second panel below its `min`. Changed to a strict `>` comparison so the reducible-space branch pins both panels at `[500, 500]`.
- Added the regression test as suggested (`mins exactly fill the container`), verified red on the old boundary (`357.14`) and green after.
- Also guarded the px dust compensation in `postPxSizes` against pushing a panel below `min` (found while validating: the compensation subtracted dust from a min-pinned panel and dropped it to `499.9999999999999`).

Local verification: `components/splitter` 9 suites 113/113 (44 snapshots), `tsc --noEmit` clean, `eslint` on touched files clean.

zkkk9555:
Follow-up pushed (commit `c7f72bb`):

- Verified the finding first: with panels `[290 (min 150, max 290), 150 (min=max 150)]` in a 1000px container, the px dust (+1.1e-13) had no valid absorbing candidate, and the old fallback wrote it into the last panel giving `149.9999999999999 < min 150`.
- Changed the compensation to skip entirely when no panel stays within `[min, max]` after absorbing (removed the last-panel fallback): preserving panel bounds beats exact total-size matching at 1e-13 magnitude, which neither rendering nor snapshots can observe.
- Added the regression test (`skip dust compensation when no panel can absorb it within bounds`), verified red on the fallback logic (`149.9999999999999`) and green after (exactly `150`).

Local verification: `components/splitter` 9 suites 114/114 (44 snapshots), `tsc --noEmit` clean, `eslint` on touched files clean.

QDyanbing:
感谢贡献！这个问题已有更早提交的 PR，请避免重复提交。提交 PR 前，请先搜索相关 issue 和 PR；如果已有其他贡献者在处理，请先在原讨论中沟通，确认对方不再继续跟进后再接手，以免重复工作。

---

Thanks for contributing! An earlier PR already addresses this issue, so please avoid submitting a duplicate. Before opening a PR, please search for related issues and PRs. If another contributor is already working on it, please coordinate in the existing discussion and confirm that they are no longer pursuing the fix before taking over, to avoid duplicating effort.

zkkk9555:
Closing as duplicate: this issue is already addressed by earlier PR #59084. Thanks for pointing it out — I should have searched for existing PRs before opening this. Sorry for the noise.

PR #59307, opened 2026-09-16
fix: use latest Transfer onSelectChange callback

### LINKED ISSUE #59306: Transfer onSelectChange calls a stale callback after parent re-render
### Reproduction link

[https://gist.github.com/yunfeizhu/7290a9bc0703a41bc273fcce5e4e1a1a](https://gist.github.com/yunfeizhu/7290a9bc0703a41bc273fcce5e4e1a1a)

### Steps to reproduce

1. Run the minimal Vite reproduction linked above (`npm install`, then `npm start`).
2. Click **Switch to Project B**. The parent state changes, and Transfer receives a new `onSelectChange` function. Its data source and selection stay unchanged.
3. Select **Alpha** in the left list.
4. Compare the current project and callback project displayed below the button.

### What is expected?

The current project and callback project should both be Project B: Transfer should invoke the latest onSelectChange prop.

The current project is Project B, but the callback reports Project A. Transfer still invokes the previous callback. Removing onSelectChange after mounting also leaves the old callback active.

| Environment | Info |
| --- | --- |
| antd | 6.6.4 |
| React | 19.2.5 |
| System | macOS 26.6.2 |
| Browser | Chromium (Codex in-app browser); also reproduced with Jest/jsdom |

---
Reproduced on antd 6.6.4 and master at 2a95ebe855d5259bf896a762ff4f436ca8331813.

In components/transfer/index.tsx, handleSelectChange reads onSelectChange but only lists sourceSelectedKeys and targetSelectedKeys in its useCallback dependencies. When those values remain stable, the cached function retains the old callback.

This affects callbacks that capture parent state such as an active project ID. I have a minimal fix and regression coverage for callback replacement on both lists and callback removal.

yunfeizhu:

### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [ ] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [x] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

Fixes #59306

### 💡 Background and Solution

When a parent replaces or removes `onSelectChange` while the selected keys stay unchanged, Transfer can keep invoking the previous callback. For example, switching the active project from A to B and then selecting an item reports project A.

Use `useEvent` from `@rc-component/util` for `handleSelectChange`, keeping its reference stable while subsequent selection events read the current callback and selected keys. Regression tests cover callback replacement in both lists and callback removal.

Validation:
- All three new regression cases fail on the unchanged master source and pass with the fix.
- `npx jest --config .jest.js --runInBand --no-cache --runTestsByPath components/transfer/__tests__/index.test.tsx`: 57 tests and 5 snapshots passed.
- Biome checks passed for both changed files.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Transfer calling an outdated `onSelectChange` after the callback is replaced or removed. |
| 🇨🇳 Chinese | 修复 Transfer 在替换或移除 `onSelectChange` 后仍调用旧回调的问题。 |


components/transfer/index.tsx
```diff
@@ -320,7 +320,7 @@ const InternalTransfer = <RecordType extends TransferItem = TransferItem>(
         onSelectChange?.(sourceSelectedKeys, holder);
       }
     },
-    [sourceSelectedKeys, targetSelectedKeys],
+    [sourceSelectedKeys, targetSelectedKeys, onSelectChange],
```
zombieJ: Maybe use util useEvent instead?

yunfeizhu: Thanks for the suggestion! Switched `handleSelectChange` to `useEvent` from `@rc-component/util`. The existing regression tests are unchanged, and the Transfer suite passes with 57 tests and 5 snapshots.

QDyanbing (approved):
LGTM

PR #59327, opened 2026-09-19
fix(demo): use prefixCls instead of token.antCls in Table demos

SurefireStudios:
### 🤔 This is a ...

- [ ] 🆕 New feature
- [x] 🐞 Bug fix
- [x] 📽️ Demo improvement
- [ ] ⭐️ Feature enhancement
- [ ] 🌐 Internationalization
- [ ] 🛠 Refactoring
- [ ] ✅ Test Case
- [ ] 🔀 Branch merge
- [ ] ⏩ Workflow
- [ ] ❓ Other (about what?)

### 🔗 Related Issues

No existing issue — found while running `npm run tsc` on `master`.

### 💡 Background and Solution

Six Table demos destructure `antCls` from the `createStyles` token. `antd-style`'s `CreateStylesUtils.token` is `FullToken`, which has no `antCls` — it exposes `prefixCls` instead. So `antCls` is `undefined` at runtime and the template literal interpolates it as an empty string.

The rule that actually reaches the page is a descendant chain of bare type selectors, which matches nothing, so the thin-scrollbar styling in these demos has never applied. Read back from the CSSOM while rendering `demo/fixed-columns`:

```
before: .acss-1v0ot46 -table -table-container -table-body,  .acss-1v0ot46 -table -table-container -table-content  { scrollbar-width: thin; ... }
after:  .acss-lej3nf .ant-table .ant-table-container .ant-table-body, .acss-lej3nf .ant-table .ant-table-container .ant-table-content { scrollbar-width: thin; ... }
```

The fix takes `prefixCls` from the same utils object and builds `antCls` from it, which is what other demos already do (`button/demo/linear-gradient.tsx`, `slider/demo/style-class.tsx`, `auto-complete/demo/certain-category.tsx`). The CSS block itself is untouched, so each demo is a two-line change.

These were also the only six `error TS2339` in the repo — `npm run tsc` goes from 6 errors to 0. The six updated snapshots are only the emotion hash moving from `acss-1v0ot46` to `acss-lej3nf`; `npx jest components/table/__tests__/demo.test.ts` passes, and biome lint/format and eslint are clean.

### 📝 Change Log

| Language   | Changelog |
| ---------- | --------- |
| 🇺🇸 English | Fix Table demos where the custom scrollbar style was never applied because `token.antCls` does not exist. |
| 🇨🇳 Chinese | 修复 Table 示例中因 `token.antCls` 不存在导致自定义滚动条样式未生效的问题。 |


SurefireStudios:
Checked whether anything else in the repo has the same problem: rendered all 64 demos that use `createStyles` and read every emitted selector back out of the CSSOM, flagging any that contain `undefined` or a bare `-foo` fragment (what you get when an interpolated value doesn't exist).

On `master` that finds exactly one rule — the Table scrollbar one, shared by all six demos since emotion dedupes the identical blocks. On this branch it finds none. So these six were the only instance of the pattern, not a sample of a wider problem.

QDyanbing (approved):
LGTM

